PR #56, opened 2021-06-29
initial EMR set-up code

Ricardo-C-Oliveira:
This is the initial work to get ds-grid to work on EMR.

Next steps:

- improve how ds-grid gets installed on emr,
- improve how aws creds are used, named profiles should not be used instead we should rely on roles when running on EMR


mooneyme:
@Ricardo-C-Oliveira  what is the status of these requested changes?

lixiliu:
PR #130 (merged into this branch) reduced the number of packages installed for the env (`dsgrid-emr`) for launching EMR, refer back to this PR if more packages are needed in the future.

lixiliu (approved):
Approved per approval for #130

PR #91, opened 2021-11-24
Add data sync capabilities to CLI

mooneyme:
This uses `dsgrid-test-data mm/sync` branch, see this related PR:  https://github.com/dsgrid/dsgrid-test-data/pull/7

...

Completes JIRA Story DSGRID-
Closes GitHub Issue #

## Checklist

- [x] All tests pass
- [x] Consider transferring TODOs to JIRA or GitHub

### DIFF
--- a/dsgrid/cli/registry.py
+++ b/dsgrid/cli/registry.py
@@ -609,9 +609,39 @@ def update_dataset(registry_manager, dataset_config_file, log_message, update_ty
 
 @click.command()
 @click.pass_obj
-def sync(registry_manager):
-    """Sync the official dsgrid registry to the local system."""
+@click.option(
+    "--project_id",
+    "-P",
+    type=str,
+    help="Sync latest dataset(s) version based on Project ID",
+)
+@click.option(
+    "--dataset_id",
+    "-D",
+    type=str,
+    help="Sync latest dataset version based on Dataset ID",
+)
+@click.option(
+    "--version",
+    "-V",
+    type=str,
+    help="Sync a specific dataset version. Must be called with the -D option.",
+)
+def data_sync(registry_manager, project_id, dataset_id):
+    """Sync the official dsgrid registry data to the local system."""
     # aws.sync(REMOTE_REGISTRY, registry_manager.path)
+    # no_prompts = registry_manager.no_prompts
+    print(f"Project_ID = {project_id}")
+    print(f"Dataset_ID = {dataset_id}")
+    registry_manager.data_sync(project_id, dataset_id)
+
+    # no_prompts = ctx.parent.params["no_prompts"]
+    # if "--help" in sys.argv:
+    #     ctx.obj = None
+    # else:
+    #     ctx.obj = RegistryManager.load(
+    #         path, remote_path, offline_mode=offline, dry_run_mode=dry_run, no_prompts=no_prompts
+    #     )
 
 
 dimensions.add_command(list_dimensions)
@@ -640,4 +670,4 @@ registry.add_command(dimensions)
 registry.add_command(dimension_mappings)
 registry.add_command(projects)
 registry.add_command(datasets)
-registry.add_command(sync)
+registry.add_command(data_sync)
--- a/dsgrid/cli/registry.py
+++ b/dsgrid/cli/registry.py
@@ -608,6 +608,7 @@ def update_dataset(registry_manager, dataset_config_file, log_message, update_ty
 
 
 @click.command()
+@click.pass_context
 @click.pass_obj
 @click.option(
     "--project_id",
@@ -627,13 +628,10 @@ def update_dataset(registry_manager, dataset_config_file, log_message, update_ty
     type=str,
     help="Sync a specific dataset version. Must be called with the -D option.",
 )
-def data_sync(registry_manager, project_id, dataset_id):
+def data_sync(ctx, registry_manager, project_id, dataset_id):
     """Sync the official dsgrid registry data to the local system."""
-    # aws.sync(REMOTE_REGISTRY, registry_manager.path)
-    # no_prompts = registry_manager.no_prompts
-    print(f"Project_ID = {project_id}")
-    print(f"Dataset_ID = {dataset_id}")
-    registry_manager.data_sync(project_id, dataset_id)
+    no_prompts = ctx.parent.params["no_prompts"]
+    registry_manager.data_sync(project_id, dataset_id, no_prompts)
 
     # no_prompts = ctx.parent.params["no_prompts"]
     # if "--help" in sys.argv:
--- a/dsgrid/cli/registry.py
+++ b/dsgrid/cli/registry.py
@@ -622,24 +622,10 @@ def update_dataset(registry_manager, dataset_config_file, log_message, update_ty
     type=str,
     help="Sync latest dataset version based on Dataset ID",
 )
-@click.option(
-    "--version",
-    "-V",
-    type=str,
-    help="Sync a specific dataset version. Must be called with the -D option.",
-)
 def data_sync(ctx, registry_manager, project_id, dataset_id):
     """Sync the official dsgrid registry data to the local system."""
     no_prompts = ctx.parent.params["no_prompts"]
     registry_manager.data_sync(project_id, dataset_id, no_prompts)
-
-    # no_prompts = ctx.parent.params["no_prompts"]
-    # if "--help" in sys.argv:
-    #     ctx.obj = None
-    # else:
-    #     ctx.obj = RegistryManager.load(
-    #         path, remote_path, offline_mode=offline, dry_run_mode=dry_run, no_prompts=no_prompts
-    #     )
 
 
 dimensions.add_command(list_dimensions)
--- a/dsgrid/cli/registry.py
+++ b/dsgrid/cli/registry.py
@@ -611,13 +611,13 @@ def update_dataset(registry_manager, dataset_config_file, log_message, update_ty
 @click.pass_context
 @click.pass_obj
 @click.option(
-    "--project_id",
+    "--project-id",
     "-P",
     type=str,
     help="Sync latest dataset(s) version based on Project ID",
 )
 @click.option(
-    "--dataset_id",
+    "--dataset-id",
     "-D",
     type=str,
     help="Sync latest dataset version based on Dataset ID",
--- a/dsgrid/cloud/s3_storage_interface.py
+++ b/dsgrid/cloud/s3_storage_interface.py
@@ -27,9 +27,9 @@ class S3StorageInterface(CloudStorageInterface):
         self._local_filesystem = LocalFilesystem()
         self._s3_filesystem = S3Filesystem(remote_path, profile)
 
-    def _sync(self, src, dst, exclude=None, cp=False):
+    def _sync(self, src, dst, exclude=None, is_file=False):
         start = time.time()
-        if cp:
+        if is_file:
             sync_command = f"aws s3 cp {src} {dst} --profile {self._s3_filesystem._profile}"
         else:
             sync_command = f"aws s3 sync {src} {dst} --profile {self._s3_filesystem._profile}"
@@ -121,7 +121,7 @@ class S3StorageInterface(CloudStorageInterface):
                 )
             filepath.unlink()
 
-    def sync_pull(self, remote_path, local_path, exclude=None, delete_local=False, cp=False):
+    def sync_pull(self, remote_path, local_path, exclude=None, delete_local=False, is_file=False):
         if delete_local:
             local_contents = self._local_filesystem.rglob(local_path)
             s3_contents = {
@@ -133,7 +133,7 @@ class S3StorageInterface(CloudStorageInterface):
                 if relcontent not in s3_contents:
                     self._local_filesystem.rm(content)
                     logger.info("delete: %s because it is not in %s", relcontent, remote_path)
-        self._sync(remote_path, local_path, exclude, cp)
+        self._sync(remote_path, local_path, exclude, is_file)
 
     def sync_push(self, remote_path, local_path, exclude=None):
         self._sync(local_path, remote_path, exclude)
--- a/dsgrid/cloud/s3_storage_interface.py
+++ b/dsgrid/cloud/s3_storage_interface.py
@@ -27,9 +27,12 @@ class S3StorageInterface(CloudStorageInterface):
         self._local_filesystem = LocalFilesystem()
         self._s3_filesystem = S3Filesystem(remote_path, profile)
 
-    def _sync(self, src, dst, exclude=None):
+    def _sync(self, src, dst, exclude=None, cp=False):
         start = time.time()
-        sync_command = f"aws s3 sync {src} {dst} --profile {self._s3_filesystem._profile}"
+        if cp:
+            sync_command = f"aws s3 cp {src} {dst} --profile {self._s3_filesystem._profile}"
+        else:
+            sync_command = f"aws s3 sync {src} {dst} --profile {self._s3_filesystem._profile}"
         if exclude:
             for x in exclude:
                 sync_command = sync_command + f" --exclude {x}"
@@ -118,7 +121,7 @@ class S3StorageInterface(CloudStorageInterface):
                 )
             filepath.unlink()
 
-    def sync_pull(self, remote_path, local_path, exclude=None, delete_local=False):
+    def sync_pull(self, remote_path, local_path, exclude=None, delete_local=False, cp=False):
         if delete_local:
             local_contents = self._local_filesystem.rglob(local_path)
             s3_contents = {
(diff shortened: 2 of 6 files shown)


dsgrid/registry/registry_manager.py
```diff
@@ -221,6 +221,86 @@ def load(
...
+                                f"No registered dataset ID = '{dataset_id}' registered to project ID = '{project_id}'"
+                            )
+                    else:
+                        datasets = []
+                        # TODO: self.dataset_manager.has_id() gives us all datasets in the registry. We want all datasets associated with a project. This is an error at the moment.
+                        for dataset in self.dataset_manager.list_ids():
+                            version = self.dataset_manager.get_current_version(dataset)
+                            datasets.append((dataset, str(version)))
+                else:
+                    raise ValueError(f"No registered project ID = '{project_id}'")
```
mooneyme, line 260: @lixiliu  see TODO comments here. Looking for the right way to filter datasets that belong to a specific project_id.

mooneyme: @daniel-thom thoughts here?

dsgrid/registry/registry_manager.py
```diff
@@ -221,6 +221,86 @@ def load(
...
+        dry_run_mode = self._params.dry_run
+        no_prompts = True  # TODO: how do we pass in no_prompts from registry cli parent?
+
+        if not offline_mode:
+            if project_id:
+                if self.project_manager.has_id(project_id):
+                    if dataset_id:
+                        # TODO: self.dataset_manager.has_id() gives us all datasets in the registry. We want all datasets associated with a project. This is an error at the moment.
+                        if self.dataset_manager.has_id(dataset_id):
+                            version = self.dataset_manager.get_current_version(dataset_id)
```
daniel-thom, line 247: This function has a bit of complexity (this is 5 levels of indentation). It could be simplified by separating concepts into different functions.
- One function could find the dataset_id/version pairs for all datasets in a project. Note the the existing method `ProjectConfig.list_registered_dataset_ids`. A similar method could get the versions along with it. (I think you might have a bug here because a project may not be using the latest version of a dataset. The dataset version needs to come from the project config.).
- One function could perform the real work. Consider this signature:
```
def sync_dataset(self, dataset_id, version):
```

dsgrid/registry/registry_manager.py
```diff
@@ -221,6 +221,86 @@ def load(
...
+            Sync by dataset_id filter
+        """
+        # TODO: docstring
+        fs_interface = self._params.fs_interface
+        cloud_interface = self._params.cloud_interface
+        offline_mode = self._params.offline
+        dry_run_mode = self._params.dry_run
+        no_prompts = True  # TODO: how do we pass in no_prompts from registry cli parent?
+
+        if not offline_mode:
```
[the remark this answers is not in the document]
mooneyme: I set to true temporarily. Wasn't quite sure how to pass in the no_prompts from the parent registry CLI.

dsgrid/cli/registry.py
```diff
@@ -609,9 +609,39 @@ def update_dataset(registry_manager, dataset_config_file, log_message, update_ty
...
+@click.option(
+    "--version",
+    "-V",
+    type=str,
+    help="Sync a specific dataset version. Must be called with the -D option.",
+)
+def data_sync(registry_manager, project_id, dataset_id):
+    """Sync the official dsgrid registry data to the local system."""
     # aws.sync(REMOTE_REGISTRY, registry_manager.path)
+    # no_prompts = registry_manager.no_prompts
```
daniel-thom, line 633: If you want to get `no_prompts` from the parent command you have to set `@click.pass_context` on the command (probably above `@click.pass_obj on line 611`). Then
```
no_prompts = ctx.parent.params["no_prompts"]
```
Did you already try this?

mooneyme: I hadn't gotten there yet but this is helpful! thank you!!

tests_aws/test_registry/test_data_sync.py
```diff
@@ -0,0 +1,165 @@
...
+s3_cloudinterface = S3StorageInterface(
+    remote_path=TEST_REMOTE_REGISTRY,
+    local_path=local_registry,
+    user="Test",
+    uuid=str(uuid.uuid4()),
+    profile=AWS_PROFILE_NAME,
+)
+
+
+def test_data_sync_project_id(s3_cloudinterface, local_registry_data_sync, remote_registry):
```
mooneyme, line 24: This test fails currently with the following error message.

```
_____________________________________________________________ ERROR at setup of test_data_sync_project_id _____________________________________________________________
file /Users/mmooney/Documents/github/github.com/dsgrid/dsgrid/tests_aws/test_registry/test_data_sync.py, line 24
  def test_data_sync_project_id(s3_cloudinterface, local_registry_data_sync, remote_registry):
E       fixture 's3_cloudinterface' not found
>       available fixtures: anyio_backend, anyio_backend_name, anyio_backend_options, cache, capfd, capfdbinary, caplog, capsys, capsysbinary, cov, doctest_namespace, monkeypatch, no_cover, pytestconfig, record_property, record_testsuite_property, record_xml_attribute, recwarn, tmp_path, tmp_path_factory, tmpdir, tmpdir_factory
>       use 'pytest --fixtures [testpath]' for help on them.
```

@daniel-thom any thoughts here? All other tests with the same `s3_cloudinterface` logic work fine and do not result in an error.

dsgrid/registry/registry_manager.py
mooneyme: @daniel-thom I added most of the changes we discussed yesterday. The only one I didn't implement is the one where you mentioned that the project_id and dataset_id components of this function could be broken out into different functions. I don't fully understand how to do this and also not totally sure its necessary given that we have made those other changes. Thoughts?

dsgrid/cli/registry.py
```diff
@@ -609,9 +609,39 @@ def update_dataset(registry_manager, dataset_config_file, log_message, update_ty
 
 @click.command()
 @click.pass_obj
-def sync(registry_manager):
-    """Sync the official dsgrid registry to the local system."""
+@click.option(
+    "--project_id",
```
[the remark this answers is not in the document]
daniel-thom: Can you make this change?

dsgrid/cloud/s3_storage_interface.py
```diff
@@ -64,15 +67,19 @@ def check_valid_lock_file(self, path):
...
         return True
 
-    def get_lock_files(self):
-        contents = self._s3_filesystem.path(".locks").glob(pattern="*")
+    def get_lock_files(self, relative_path=None):
+        contents = list(self._s3_filesystem.path(".locks").glob(pattern="*"))
+        if relative_path:
+            for path in contents:
+                if not path.relative_to(relative_path):
+                    contents.remove(path)
```
daniel-thom, line 82: This modifies the list that is being iterated over. Python will raise an exception. You have to collect the indices in this loop and then delete them in reverse order in a second loop.

dsgrid/registry/registry_manager.py
```diff
@@ -221,6 +222,97 @@ def load(
...
+            if not self.project_manager.has_id(project_id):
+                raise ValueError(f"No registered project ID = '{project_id}'")
+            project_version = self.project_manager.get_current_version(project_id)
+            config_file = self.project_manager.get_config_file(project_id, project_version)
+            config = ProjectConfig.load(
+                config_file, self.dimension_manager, self.dimension_mapping_manager
+            )
+            if dataset_id:
+                if dataset_id not in config.list_registered_dataset_ids():
+                    raise ValueError(
```
daniel-thom, line 250: It's probably better to raise `DSGValueNotRegistered` so that we can handle all dsgrid exceptions in a common way.

dsgrid/registry/registry_manager.py
```diff
@@ -221,6 +222,97 @@ def load(
...
+        cloud_interface = self._params.cloud_interface
+        offline_mode = self._params.offline
+        dry_run_mode = self._params.dry_run
+
+        if offline_mode:
+            raise ValueError("dsgrid data-sync only works in online mode.")
+        if dry_run_mode == True:
+            sync = False
+        else:
+            sync = True
```
daniel-thom, line 278: This is redundant. You can delete. Or `sync = not dry_run_mode`

dsgrid/registry/registry_manager.py
```diff
@@ -221,6 +222,97 @@ def load(
...
+            logger.log(msg)
+            if not no_prompts:
+                msg = msg + "\n... Do you want to continue syncing the registry contents? [Y] >>> "
+                val = input(msg)
+                if val == "" or val.lower() == "y":
+                    sync = True
+                else:
+                    logger.info("Skipping remote registry sync.")
+
+        if sync:
```
daniel-thom, line 298:
```
if dry_run_mode:
    logger.info(...)
else:
    sync_pull(...)
```

dsgrid/registry/registry_manager.py
```diff
@@ -221,6 +222,97 @@ def load(
...
+        if offline_mode:
+            raise ValueError("dsgrid data-sync only works in online mode.")
+        if dry_run_mode == True:
+            sync = False
+        else:
+            sync = True
+
+        lock_files = list(
+            cloud_interface.get_lock_files(
+                relative_path=f"s3:/{cloud_interface._s3_filesystem._bucket}/configs/datasets/{dataset_id}"
```
daniel-thom, line 282: Do you need to include `s3` here? Isn't that part already encapsulated within the s3 cloud interface? Also, you're referring to `cloud_interface._s3_filesystem._bucket.`, which doesn't seem correct because it has two private variable accesses.

dsgrid/registry/registry_manager.py
```diff
@@ -221,6 +222,97 @@ def load(
...
+            )
+        )
+        if lock_files:
```
daniel-thom, line 285: Can there be more than one lock file?

dsgrid/registry/registry_manager.py
```diff
@@ -221,6 +222,97 @@ def load(
...
+            cloud_interface.sync_pull(
+                remote_path=self._params.remote_path + f"/data/{dataset_id}/{version}",
+                local_path=str(self._params.base_path) + f"/data/{dataset_id}/{version}",
+                exclude=SYNC_EXCLUDE_LIST,
+                delete_local=True,
+            )
+            cloud_interface.sync_pull(
+                remote_path=self._params.remote_path + f"/data/{dataset_id}/registry.toml",
+                local_path=str(self._params.base_path) + f"/data/{dataset_id}/registry.toml",
+                exclude=SYNC_EXCLUDE_LIST,
```
daniel-thom, line 309: This isn't needed, right?

dsgrid/registry/registry_manager.py
daniel-thom: I'm OK with it.

dsgrid/registry/registry_manager.py
mooneyme: the first sync_pull uses the aws sync command to sync directories and files. The second one is for a single file which uses the aws cp command. Both are needed.

dsgrid/cli/registry.py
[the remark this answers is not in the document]
mooneyme: I don't follow. Don't I already have that?

dsgrid/registry/registry_manager.py
daniel-thom: I was referring to this specific line:
```
exclude=SYNC_EXCLUDE_LIST,
```

dsgrid/cli/registry.py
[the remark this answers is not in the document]
daniel-thom: You have underscores mixed in both cases. The convention is to use dashes.

dsgrid/cloud/s3_storage_interface.py
mooneyme: Like this?:

```
    def get_lock_files(self, relative_path=None):
        contents = list(self._s3_filesystem.path(".locks").glob(pattern="*"))
        if relative_path:
            relative_contents = []
            for path in contents:
                if path.relative_to(relative_path):
                    relative_contents.append(path)
            return relative_contents
        return contents
```

dsgrid/registry/registry_manager.py
mooneyme: sync doesn't exist before this statement

dsgrid/registry/registry_manager.py
mooneyme: No likely not because the lock is at the dataset config level. Do you want me to make this singular in this case?

dsgrid/registry/registry_manager.py
daniel-thom: I was trying to say that you don't really need the sync variable. You can use dry_run_mode as shown below. If you really want to have the new variable then you can shrink the 4 lines into `sync = not dry_run_mode`.

mooneyme: Hmm. I don't quite follow what to do here with this comment or the one below. Do you want to suggest changes with a commit?

daniel-thom: It's not important. I'll resolve it.

dsgrid/registry/registry_manager.py
daniel-thom: The code is written to use multiple lock files but I didn't see how that was possible. You could either leave it alone or put include an `assert len(lock_files) == 1` in order to inform readers.

dsgrid/cloud/s3_storage_interface.py
```diff
@@ -27,9 +27,12 @@ def __init__(self, local_path, remote_path, uuid, user, profile):
         self._local_filesystem = LocalFilesystem()
         self._s3_filesystem = S3Filesystem(remote_path, profile)
 
-    def _sync(self, src, dst, exclude=None):
+    def _sync(self, src, dst, exclude=None, is_file=False):
         start = time.time()
-        sync_command = f"aws s3 sync {src} {dst} --profile {self._s3_filesystem._profile}"
+        if is_file:
+            sync_command = f"aws s3 cp {src} {dst} --profile {self._s3_filesystem._profile}"
```
daniel-thom, line 33: Here and below with `_bucket` you are referencing private attributes of `_s3_filesystem`. Can you make those public properties to avoid this?
```
@property
def profile(self):
    return self._profile
```

dsgrid/cloud/s3_storage_interface.py
```diff
@@ -64,15 +67,28 @@ def check_valid_lock_file(self, path):
...
+    def get_lock_files(self, relative_path=None):
+        if relative_path:
+            contents = self._s3_filesystem.path(
+                f"{self._s3_filesystem._bucket}/{relative_path}"
+            ).glob(pattern="**/*.locks")
+        else:
+            contents = self._s3_filesystem.path(self._s3_filesystem._bucket).glob(
+                pattern="**/*.locks"
+            )
+        # if relative_path:
```
daniel-thom, line 86: Should this be removed? Is the PR still in draft?

dsgrid/registry/registry_manager.py
```diff
@@ -221,6 +223,97 @@ def load(
...
+            raise ValueError("Must provide a dataset_id or project_id for dsgrid data-sync.")
+
+        if project_id:
+            if not self.project_manager.has_id(project_id):
+                raise DSGValueNotRegistered(f"No registered project ID = '{project_id}'")
+            project_version = self.project_manager.get_current_version(project_id)
+            config_file = self.project_manager.get_config_file(project_id, project_version)
+            config = ProjectConfig.load(
+                config_file, self.dimension_manager, self.dimension_mapping_manager
+            )
```
daniel-thom, line 248: I just noticed that you could replace these five lines (including the exception) with
```
config = self.project_manager.get_by_id(project_id)
```
It handles the versions for you.

dsgrid/registry/registry_manager.py
```diff
@@ -221,6 +223,97 @@ def load(
...
+        )
+        assert len(lock_files) == 1
+        if lock_files:
+            msg = f"There are {len(lock_files)} lock files in the registry:"
+            for lock_file in lock_files:
+                msg = msg + "\n\t" + f"- {lock_file}"
+            logger.log(msg)
+            if not no_prompts:
+                msg = msg + "\n... Do you want to continue syncing the registry contents? [Y] >>> "
+                val = input(msg)
```
daniel-thom, line 294: There is a bug here. The user shouldn't get this prompt if they passed dry_run_mode. You could change line 287 to `if lock_files and not dry_run_mode`.

dsgrid/registry/registry_manager.py
```diff
@@ -221,6 +223,97 @@ def load(
...
+            sync = False
+        else:
+            sync = True
+
+        lock_files = list(
+            cloud_interface.get_lock_files(
+                relative_path=f"{cloud_interface._s3_filesystem._bucket}/configs/datasets/{dataset_id}"
+            )
+        )
+        assert len(lock_files) == 1
```
daniel-thom, line 286: This should be inside the next if statement, right? Normally, there won't be any lock files.

dsgrid/registry/registry_manager.py
mooneyme: nice!!!

mooneyme:
@daniel-thom this PR needs to be approved with https://github.com/dsgrid/dsgrid-test-data/pull/7

daniel-thom (approved):

PR #93, opened 2021-11-29
Allow null data IDs in load_data_lookup

daniel-thom:


mooneyme (approved):

dsgrid/config/project_config.py
```diff
@@ -241,6 +241,11 @@ class ProjectConfigModel(DSGBaseModel):
         optional=True,
         notes=("`[dimension_mappings]` are optional at the project level.",),
     )
+    dimension_associations: List = Field(
```
daniel-thom, line 244: @mooneyme Note that I added this before merging to main.

PR #94, opened 2021-12-11
Initialize a new SparkSession for each test

daniel-thom:
The current code uses the same SparkSession for all tests. This is incorrect behavior.


lixiliu:
Looks good.

lixiliu (approved):

PR #95, opened 2021-12-24
Fix time check for dataset with timeseries of variable ranges

lixiliu:
Not done, still need to test and validate most of the code changes.

Completes JIRA Story DSGRID-
Closes GitHub Issue #

## Checklist

- [ ] All tests pass


dsgrid/registry/dataset_registry_manager.py
```diff
@@ -129,17 +131,78 @@ def _check_lookup_data_consistency(self, config: DatasetConfig, load_data_lookup
...
+                "id",
+                "left",
+            ).select(
+                "id",
+                "geography",
+                "timestamp",
+                F.lit(data_tz).alias("from_timezone"),
+                F.col("to_id").alias("to_timezone"),
+            )
+            udf_find_time_delta = F.udf(lambda fm, to: find_time_delta(fm, to), T.DateType())
```
daniel-thom, line 169: I don't think I'm following the type you are creating here. The find_time_delta function returns a Python timedelta object and I don't see how that gets stored in a DateType. Do you want to convert the timedelta to an integer expressed as number of seconds?

Also, note that UDFs in Python are very slow. There may not be a better way.

lixiliu: You are right, I did not look super hard, maybe `CalendarIntervalType` is more appropriate? https://sparkbyexamples.com/pyspark/pyspark-sql-types-datatype-with-examples/

daniel-thom: I'm not sure that that is actually a Spark type. https://spark.apache.org/docs/latest/sql-ref-datatypes.html

`timedelta.total_seconds()` stored as an integer is probably good enough.

dsgrid/config/dimensions.py
```diff
@@ -339,7 +339,22 @@ class TimeDimensionModel(TimeDimensionBaseModel):
...
+        description="""
+        Time zone of the ranges specified:
+            UTC, 
+            HawaiiAleutianStandard, 
+            AlaskaStandard, AlaskaPrevailing,
+            PacificStandard, PacificPrevailing, 
+            MountainStandard, MountainPrevailing, 
+            CentralStandard, CentralPrevailing, 
+            EasternStandard, EasternPrevailing,
+            LOCAL
```
mooneyme, line 355: Is ComStock/ResStock data in LOCAL prevailing? Or LOCAL standard?

mooneyme, line 355: This should probably go into an enumeration instead of being duplicated here in the description.

mooneyme: Looks like you don't need to add any of these. The docs take care of it:
image

mooneyme (changes requested):
@lixiliu  can you please add a test for these new changes?

dsgrid/dimension/time.py
```diff
@@ -94,22 +94,36 @@ class MeasurementType(DSGEnum):
...
+    America/Denver (MPT)
+    America/Chicago (CPT)
+    America/New_York (EPT)
+    """
 
     UTC = EnumValue(
         value="UTC",
         description="Coordinated Universal Time",
-        tz=datetime.timezone(datetime.timedelta()),
+        tz=pytz.timezone("UTC"),  # datetime.timezone(datetime.timedelta()),
```
daniel-thom, line 116: Remove the old code?

dsgrid/dimension/time.py
```diff
@@ -259,3 +273,16 @@ def make_time_range(start, end, frequency, leap_day_adjustment):
             "366 days not allowed for frequency, use 365 days to specify annual frequency."
         )
     return DatetimeRange(start, end, frequency, leap_day_adjustment)
+
+
+def find_time_delta(timestamp, from_tz: str, to_tz: str):
+    """
+    find datetime.timedelta (in seconds) between two time zones
+    output: integer (sec)
```
daniel-thom, line 281: numpy docstring format
```
Returns
--------
int
    Difference between the two time zones in seconds.
```

dsgrid/config/dimensions.py
lixiliu: LOCAL standard

PR #99, opened 2022-01-19
Support "one_table" data_schema_type

lixiliu:
Completes JIRA Story DSGRID-
Closes GitHub Issue #

This PR to follow[ PR #100](https://github.com/dsgrid/dsgrid/pull/100)
and is in preparation for registering growth rate type of data.

Companion dsgrid-test-data PR: https://github.com/dsgrid/dsgrid-test-data/pull/10

## Description
Code to validate with dataset registration for AEO reference Scenario Commercial `End Uses` on dsgrid-project-StandardScenarios/ll-historical branch [(PR #5](https://github.com/dsgrid/dsgrid-project-StandardScenarios/pull/5)). 

## Checklist

- [x] All tests pass


dsgrid/registry/dataset_registry_manager.py
```diff
@@ -127,37 +143,123 @@ def _check_lookup_data_consistency(self, config: DatasetConfig, load_data_lookup
...
+    def _check_one_table_data_consistency(self, config: DatasetConfig, load_data_one_table):
+        """
+        Check all data dimensions, including Metric records.
+        For Time, only check that time_col_name = timestamp
+        """
+        dimension_types = []
+        metric_cols = []
+
+        for col in load_data_one_table.columns:
+            if col.lower() == "time":
```
daniel-thom, line 155: It may be better to raise an exception if `timestamp` is not a column.

dsgrid/registry/dataset_registry_manager.py
```diff
@@ -127,37 +143,123 @@ def _check_lookup_data_consistency(self, config: DatasetConfig, load_data_lookup
...
+                raise ValueError(
+                    f"col_name={col} for the time dimension is not supported, used col_name='timestamp' instead."
+                )
+            else:
+                try:
+                    dimension_types.append(DimensionType(col))
+                except ValueError:
+                    if col != "timestamp":
+                        metric_cols.append(col)
```
daniel-thom, line 164: Do we know what columns are required ahead of time? If so, it would be better to handle them explicitly. It seems like this implicit handling could be problematic. Should an exception be raised if a column is neither a valid dimension nor a valid metric?

dsgrid/registry/dataset_registry_manager.py
```diff
@@ -127,37 +143,123 @@ def _check_lookup_data_consistency(self, config: DatasetConfig, load_data_lookup
...
+
+        dimension_types.append(DimensionType.METRIC)  # add metric type
+
+        # check dim outside of metric and time
+        expected_dimensions = [d for d in DimensionType if d != DimensionType.TIME]
```
daniel-thom, line 169: Make this a set right away.
```
expected_dimensions = {d for d in DimensionType if d != DimensionType.TIME}
```

dsgrid/registry/dataset_registry_manager.py
```diff
@@ -127,37 +143,123 @@ def _check_lookup_data_consistency(self, config: DatasetConfig, load_data_lookup
...
             )
 
-        timestamps_by_id = (
-            load_data.select("timestamp", "id")
-            .groupby("id")
-            .agg(F.countDistinct("timestamp").alias("distinct_timestamps"))
-        )
-        for row in timestamps_by_id.collect():
-            if row.distinct_timestamps != len(expected_timestamps):
+        dimension_types = [
```
daniel-thom, line 181: This is already `expected_dimensions`.

dsgrid/registry/dataset_registry_manager.py
```diff
@@ -127,37 +143,123 @@ def _check_lookup_data_consistency(self, config: DatasetConfig, load_data_lookup
...
+            # TODO: need to support validation of multiple time ranges: DSGRID-173
+
+            expected_timestamps = time_dim.list_time_range(time_range)
+            actual_timestamps = [
+                x.timestamp
+                for x in load_data.select("timestamp").distinct().sort("timestamp").collect()
+            ]
+            try:
+                actual_timestamps = [x.astimezone().astimezone(tz) for x in actual_timestamps]
+            except AttributeError:
```
daniel-thom, line 223: Why will this be raised? Can that condition be checked explicitly?

dsgrid/registry/dataset_registry_manager.py
```diff
@@ -127,37 +143,123 @@ def _check_lookup_data_consistency(self, config: DatasetConfig, load_data_lookup
...
+                if tz == None:
+                    actual_timestamps = [
+                        pd.Timestamp(str(x)).to_pydatetime() for x in actual_timestamps
+                    ]  # tz=tz does not work here when tz=None
+                else:
+                    actual_timestamps = [
```
daniel-thom, line 229: If I'm reading this correctly, in case 1 we'll have python datetime objects and in case 2 we'll have pandas.Timestamp. Will that be problematic later?

dsgrid/registry/dataset_registry_manager.py
```diff
@@ -127,37 +143,123 @@ def _check_lookup_data_consistency(self, config: DatasetConfig, load_data_lookup
...
+
+            if expected_timestamps != actual_timestamps:
+                mismatch = sorted(
+                    set(expected_timestamps).symmetric_difference(set(actual_timestamps))
+                )
                 raise DSGInvalidDataset(
-                    f"load_data ID={row.id} does not have {len(expected_timestamps)} timestamps: actual={row.distinct_timestamps}"
+                    f"load_data timestamps do not match expected times. mismatch={mismatch}"
                 )
+            else:
```
daniel-thom, line 240: Nit: remove this else

dsgrid/registry/dataset_registry_manager.py
lixiliu: Both are date time objects (i.e. converted from pandas using `.topydatetime()`), another way to get date time objects would be directly using `datetime.datetime(year=year, month=1, day=1, tzinfo=tz)` but that's much lengthier. Since we do use pd.Timestamp in the time class (https://github.com/dsgrid/dsgrid/blob/8cc304b031edb57ca3139b78ebdf52d1b3145e62/dsgrid/config/dimension_config.py#L101), I thought it would be okay here.

dsgrid/registry/dataset_registry_manager.py
lixiliu: Yes, in the case of annual time, x is an integer, not a date time obj, so it will fail at x.astimezone() with AttributeError.

dsgrid/registry/dataset_registry_manager.py
daniel-thom: Sorry, didn't read closely enough.

dsgrid/config/dataset_config.py
```diff
@@ -24,12 +24,13 @@
     "load_data_lookup.csv",
     "load_data_lookup.json",
 )
+ALLOWED_LOAD_DATA_ONE_TABLE_FILENAMES = ("load_data_one_table.parquet", "load_data_one_table.csv")
```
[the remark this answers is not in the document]
lixiliu: I thought about that, but there are tests specific to load_data in the two-table format so I thought one_table having its own name would make it less confusing.

dsgrid/registry/dataset_registry_manager.py
lixiliu: but growth rates would not have a timestamp col.

dsgrid/registry/dataset_registry_manager.py
daniel-thom: It would be much safer to handle it explicitly. Relying on exceptions can result in very strange behaviors if something happens that you don't expect - like if AttributeError is raised for a different reason.

dsgrid/registry/dataset_registry_manager.py
daniel-thom: Ok. I think my main point is that it seems odd to check only for the invalid column "time". Other invalid names are not checked. Can we check explicitly for all required columns?

dsgrid/config/dataset_config.py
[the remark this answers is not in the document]
elainethale: I'm generally in agreement with Dan here, but appreciate your practical point, Lixi. 

The dataset config does specify the schema. How hard would it be to rework the tests to use that field to determine whether we are expecting one file (load_data.parquet) or two (load_data.parquet and load_data_lookup.parquet)?

dsgrid/registry/dataset_registry_manager.py
elainethale: And can we reuse a lot of the validations already created for the two-table format?

dsgrid/config/annual_time_dimension_config.py
```diff
@@ -0,0 +1,30 @@
+from datetime import timedelta
+
+from .dimensions import AnnualTimeDimensionModel
+from .time_dimension_base_config import TimeDimensionBaseConfig
+
+
+class AnnualTimeDimensionConfig(TimeDimensionBaseConfig):
+    """Provides an interface to an AnnualTimeDimensionModel."""
```
elainethale, line 8: I thought we already had an annual time dimension??? Did it not get merged in earlier?

dsgrid/config/date_time_dimension_config.py
```diff
@@ -0,0 +1,102 @@
...
+from dsgrid.exceptions import DSGInvalidDataset
+from .dimensions import DateTimeDimensionModel
+from .time_dimension_base_config import TimeDimensionBaseConfig
+
+
+logger = logging.getLogger(__name__)
+
+
+class DateTimeDimensionConfig(TimeDimensionBaseConfig):
+    """Provides an interface to a DateTimeDimensionModel."""
```
elainethale, line 18: Why is this config needed? How does it relate to annual time? What would happen if we needed to add another type of time representation?

Does it really need its own file or should it also be in dimension_config.py?

dsgrid/config/dimension_config.py
```diff
@@ -140,7 +130,10 @@ def get_tzinfo(self):
 
 
 class AnnualTimeDimensionConfig(TimeDimensionBaseConfig):
-    """Provides an interface to an AnnualTimeDimensionModel."""
+    """
+    Provides an interface to an AnnualTimeDimensionModel.
+    AnnualTimeDimension is primarily for constant annual growth factors
```
elainethale, line 135: There are going to be plenty of data at the annual level, not just growth factors.

dsgrid/config/dimension_config.py
```diff
@@ -1,30 +1,19 @@
...
+from .dimensions import DimensionModel
 from dsgrid.data_models import serialize_model, ExtendedJSONEncoder
-from dsgrid.dimension.time import (
-    DatetimeRange,
-    AnnualTimeRange,
-    TimeZone,
-    TimeDimensionType,
-    make_time_range,
-)
-from dsgrid.exceptions import DSGInvalidOperation
```
elainethale, line 27: I guess we were importing a bunch of things we didn't need before?

dsgrid/config/annual_time_dimension_config.py
elainethale: Is this a duplicate of https://github.com/dsgrid/dsgrid/pull/99/files#diff-37f10c1c7dfd83c009a1527d84670f7f08d89e98d8660205c5439e3e99cbef5cL142? (Why are you adding a class with the same name as another class (even if it is in a different file)?)

dsgrid/config/dimensions.py
```diff
@@ -400,34 +416,22 @@ class AnnualTimeDimensionModel(TimeDimensionBaseModel):
...
         description="Format of the timestamps in the load data",
     )
-    str_format: str = Field(
-        title="str_format",
-        description="Timestamp string format",
-        notes=(
-            "The string format is used to parse the timestamps provided in the time ranges."
-            "Cheatsheet reference: `<https://strftime.org/>`_.",
-        ),
-    )
```
elainethale, line 416: Are we sure this is no longer needed? Generally all this time stuff seems like it should be unrelated to supporting a one-table format?

elainethale (changes requested):
About 1/3 of the way through the files I am concerned that this PR is supposed to be about supporting a one-table schema but there are a whole bunch of time-related changes. Can we untangle this so we can talk through the two issues separately? (In two different PRs?)

dsgrid/config/dimensions.py
daniel-thom: @elainethale You'll want to read https://github.com/dsgrid/dsgrid/pull/100 first. This PR includes those changes. In order to support TEMPO I need to refactor handling of time. Lixi is going to build on top of that.

My refactor involved separating some functionality into different files. You're welcome to voice disagreement in that PR. I hit a point where there was too much different logic in the same file.

elainethale: Thanks, Dan. That was the information I was missing.

dsgrid/registry/dataset_registry_manager.py
```diff
@@ -70,15 +73,41 @@ def _run_checks(self, config: DatasetConfig):
...
+                check_load_data_one_table_filename(path), cache=True
+            )
+            load_data_one_table = config.add_trivial_dimensions(load_data_one_table)
+            with Timer(timer_stats_collector, "check_one_table_data_consistency"):
+                self._check_one_table_data_consistency(config, load_data_one_table)
+            with Timer(timer_stats_collector, "check_dataset_time_consistency"):
+                self._check_dataset_time_consistency(config, load_data_one_table)
+        # else:
+        #     raise ValueError(f'data_schema_type={data_schema_type} not supported.')
+#### ####
```
elainethale, line 110: This checking code should probably be relegated to the DatasetConfig which could more naturally handle different schemas and dataset types.

elainethale (commented):
Have you thought about how to make some simple tests for this new schema?

dsgrid/config/dataset_schema_handler_one_table.py
```diff
@@ -0,0 +1,82 @@
...
+        time_dim = config.get_dimension(DimensionType.TIME)
+        time_columns = set(time_dim.get_timestamp_load_data_columns())
+
+        for col in load_data.columns:
+            if col in time_columns:
+                continue
+            try:
+                dimension_types.append(DimensionType(col))
+            except ValueError:
+                metric_cols.append(col)
```
daniel-thom, line 42: Can you check for load value columns explicitly? Also, are they guaranteed to be of type metric?

dsgrid/config/dataset_schema_handler_one_table.py
```diff
@@ -0,0 +1,82 @@
...
+
+        dimension_types.append(DimensionType.METRIC)  # add metric type
+
+        # check dim outside of metric and time
+        expected_dimensions = set([d for d in DimensionType if d != DimensionType.TIME])
```
daniel-thom, line 47:
```
{d for d in DimensionType if d != DimensionType.TIME}
```

dsgrid/config/dataset_schema_handler_one_table.py
```diff
@@ -0,0 +1,82 @@
...
+        if len(dimension_types) != len(expected_dimensions):
+            raise DSGInvalidDataset(
+                f"load_data does not have the correct number of dimensions specified between trivial and non-trivial dimensions."
+            )
+
+        missing_dimensions = expected_dimensions.difference(dimension_types)
+        if len(missing_dimensions) != 0:
```
daniel-thom, line 54: More pythonic:
```
if missing_dimensions:
```

dsgrid/config/dataset_schema_handler_standard.py
```diff
@@ -0,0 +1,176 @@
...
+                continue
+            try:
+                dimension_types.append(DimensionType(col))
+            except ValueError:
+                raise DSGInvalidDimension(f"load_data_lookup column={col} is not a dimension type")
+
+        if not found_id:
+            raise DSGInvalidDataset("load_data_lookup does not include an 'id' column")
+
+        # TODO: some of this logic will change based on the data table schema type
```
daniel-thom, line 49: Can this TODO be removed?

dsgrid/config/dataset_schema_handler_one_table.py
lixiliu: No, we just assume all that have ValueError are of metric type. They are collected and then compared to the metric dimension records. https://github.com/dsgrid/dsgrid/blob/e5f4080c6432f42d2782bb345451a168bbea5e5b/dsgrid/config/dataset_schema_handler_one_table.py#L66-L75

.gitignore
```diff
@@ -150,3 +150,4 @@ dmypy.json
 .pyre/
 
 tests/data/registry/*
+emr/credentials/
```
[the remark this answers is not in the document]
lixiliu: part of Ricardo's AWS PR.

dsgrid/config/dataset_schema_handler_one_table.py
elainethale: Oh, and I guess you are assuming that the MetricType dimension is the one that is pivoted out. I thought we might have/were planning to make that configurable? (E.g., list in the dataset config which dimension type is pivoted out to form the value columns.)

elainethale: As discussed this afternoon, the schema should specify which dimension type is pivoted, and somewhere the dataset (speaking loosely) should be able to specify what column names it is expecting by combining the dataset config with the dimension records.

dsgrid/config/date_time_dimension_config.py
elainethale: This was resolved in Dan's earlier PR

elainethale (changes requested):
Nearly there, just need to explicitly handle load_data column name expectations and allow contributors to specify which dimension type is pivoted out to form the value columns

dsgrid/config/dataset_config.py
```diff
@@ -198,6 +213,15 @@ class DatasetConfigModel(DSGBaseModel):
         ),
         # TODO: it would be nice to extend the description here with a CLI example of how to list the project's data source IDs.
     )
+    data_schema_type: DataSchemaType = Field(
+        title="data_schema_type",
+        description="Discriminator for data schema",
+        options=DataSchemaType.format_for_docs(),
+    )
+    data_schema: Union[StandardDataSchemaModel, OneTableDataSchemaModel] = Field(
```
daniel-thom, line 221: I have experienced weird problems with Pydantic and Union types. Since you are using `data_schema_type` as a discriminator and creating the object yourself, it would be fine to leave this defined as
```
data_schema: Any
```
If everything is working then you don't need to change it.

dsgrid/config/dataset_schema_handler_standard.py
```diff
@@ -0,0 +1,182 @@
...
+        distinct_counts = timestamps_by_id.select("distinct_timestamps").distinct()
+        if distinct_counts.count() == 1:
+            val = distinct_counts.collect()[0].distinct_timestamps
+            if val != expected_count:
+                raise DSGInvalidDataset(
+                    f"load_data arrays do not have {len(expected_timestamps)} timestamps: actual={row.distinct_timestamps}"
+                )
+            return
+
+        for row in timestamps_by_id.collect():
```
daniel-thom, line 116: I changed the order of this code block and the previous one in a PR based on your comment. Looks like you didn't pick it up.

dsgrid/config/dimensions.py
```diff
@@ -389,31 +395,81 @@ class DateTimeDimensionModel(TimeDimensionBaseModel):
...
+                f'time_type={values["time_type"].value} does not match class_name={values["class_name"]}. \n'
+                " * For class=Time, use time_type=datetime. \n"
+                " * For class=AnnualTime, use time_type=annual."
+            )
+        return values
 
-    @validator("frequency")
-    def check_frequency(cls, value):
-        if value == timedelta(days=366):
+    @root_validator(pre=False)
```
daniel-thom, line 415: Does this need to be a root_validator?

dsgrid/config/dimensions.py
```diff
@@ -389,31 +395,81 @@ class DateTimeDimensionModel(TimeDimensionBaseModel):
...
-        description="""
-        Whether annual time includes leap day. Accepted: 
-            True, False
-        """,
+        description="Whether annual time includes leap day.",
     )
 
+    @root_validator(pre=False)  # TODO: modify as model works with more time_type schema
+    def check_time_type(cls, values):
+        if (
```
daniel-thom, line 454: Assuming this is the same logic as above, can you put it in a function and then call it from both places?

dsgrid/config/dataset_config.py
```diff
@@ -96,19 +96,30 @@ class DataSchemaType(DSGEnum):
...
 
 class DSGDatasetParquetType(DSGEnum):
     """Dataset parquet types."""
 
     LOAD_DATA = EnumValue(
         value="load_data",
-        description="Load data file with id, timestamp, and load value columns",
+        description="""
+        In STANDARD data_schema_type, load_data is a file with id, timestamp, and load value columns. 
+        In ONE_TABLE data_schema_type, load_data is a file with timestamp, load value, and metadata columns.
```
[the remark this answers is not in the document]
lixiliu: How about this:
In ONE_TABLE data_schema_type, load_data is a file with multiple data dimension, timestamp, and load value columns.

dsgrid/config/dataset_config.py
```diff
@@ -96,19 +96,30 @@ class DataSchemaType(DSGEnum):
...
+        """,
     )
     LOAD_DATA_LOOKUP = EnumValue(
         value="load_data_lookup",
-        description="Load data file with dimension metadata and and ID which maps to load_data"
-        "file",
+        description="""
+        load_data is a file with dimension metadata and and ID which maps to load_data file
```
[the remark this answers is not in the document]
lixiliu: How about sth like this:
load_data is a file with multiple data dimension columns and an ID column that maps to load_data file

dsgrid/config/dataset_config.py
```diff
@@ -251,15 +275,6 @@ class DatasetConfigModel(DSGBaseModel):
...
     )
-    data_schema_type: DataSchemaType = Field(
-        title="data_schema_type",
-        description="Discriminator for data schema",
-        options=DataSchemaType.format_for_docs(),
-    )
-    data_schema: StandardDataSchemaModel = Field(
-        title="data_schema",
-        description="Schema (table layouts) used for writing out the dataset",
-    )  # Once we have another schema type this will become Union[StandardDataSchemaModel, OtherSchemaModel]
```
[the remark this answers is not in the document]
lixiliu: For some validations, order matter, but the changes did not trigger any of the pytests.

dsgrid/config/dataset_config.py
lixiliu: The union is working for me

dsgrid/config/dataset_schema_handler_one_table.py
lixiliu: Got it, I'll use `load_data_column_dimension`, which the standard schema is already using

dsgrid/config/dimensions.py
lixiliu: Somehow yes, I tried it without root and it would not raise an error even though the if statement would be TRUE

dsgrid/config/dataset_config.py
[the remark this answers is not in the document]
daniel-thom: To be clear the order is very important here. The validator for data_schema requires that data_schema_type be listed first.

daniel-thom (approved):

dsgrid/config/dataset_config.py
```diff
@@ -96,19 +96,30 @@ class DataSchemaType(DSGEnum):
...
+        In STANDARD data_schema_type, load_data is a file with ID, timestamp, and load value columns. 
+        In ONE_TABLE data_schema_type, load_data is a file with multiple data dimension, timestamp, and load value columns.
+        """,
     )
     LOAD_DATA_LOOKUP = EnumValue(
         value="load_data_lookup",
-        description="Load data file with dimension metadata and and ID which maps to load_data"
-        "file",
+        description="""
+        load_data is a file with multiple data dimension columns and an ID column that maps to load_data file.
```
elainethale, line 121: load_data should be load_data_lookup

dsgrid/config/dataset_config.py
```diff
@@ -96,19 +96,30 @@ class DataSchemaType(DSGEnum):
...
 
 class DSGDatasetParquetType(DSGEnum):
     """Dataset parquet types."""
 
     LOAD_DATA = EnumValue(
         value="load_data",
-        description="Load data file with id, timestamp, and load value columns",
+        description="""
+        In STANDARD data_schema_type, load_data is a file with ID, timestamp, and load value columns. 
+        In ONE_TABLE data_schema_type, load_data is a file with multiple data dimension, timestamp, and load value columns.
```
elainethale, line 115: "multiple data dimension and load value columns"? "timestamp" is just another data dimension after all ...

elainethale: Shoot, and for all of these, instead of "load value" it should say "metric value"

elainethale: This is also making me question our rampant use of "load_data", but I think that's okay--even though we want to register multiple metrics, we are focused on load-relevant data.

dsgrid/config/dataset_config.py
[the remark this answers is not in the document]
elainethale: And will pydantic ignore the field order in the .toml to make it conform to this ordering?

dsgrid/config/dataset_schema_handler_one_table.py
```diff
@@ -0,0 +1,103 @@
...
+            elif col in expected_pivot_columns:
+                dimension_types.add(pivot_dim)
+                pivot_cols.append(col)
+            else:
+                try:
+                    dimension_types.add(DimensionType(col))
+                except ValueError:
+                    raise DSGInvalidDimension(
+                        f"load_data column={col} is not expected or of a known dimension type."
+                    )
```
elainethale, line 60: Good edits!

dsgrid/config/dataset_schema_handler_one_table.py
```diff
@@ -0,0 +1,103 @@
...
+                f"load_data contains duplicated column name(s)={pivot_cols_dup}."
+            )
+
+        expected_dimensions = {d for d in DimensionType}
+        if len(dimension_types) != len(expected_dimensions):
+            raise DSGInvalidDataset(
+                f"load_data does not have the correct number of dimensions specified between trivial and non-trivial dimensions."
+            )
+
+        missing_dimensions = expected_dimensions.difference(dimension_types)
```
elainethale, line 75: You're never going to get here because you're going to raise an exception on Line 72.

elainethale: Since I don't think it's possible to end up with len(dimension_types) > len(expected dimensions), you could just delete lines 70-73

dsgrid/config/dataset_schema_handler_standard.py
```diff
@@ -0,0 +1,187 @@
...
+            DimensionType.TIME,
+            config.model.data_schema.load_data_column_dimension,
+        )
+        expected_dimensions = [d for d in DimensionType if d not in load_data_dimensions]
+        if len(dimension_types) != len(expected_dimensions):
+            raise DSGInvalidDataset(
+                f"load_data_lookup does not have the correct number of dimensions specified between trivial and non-trivial dimensions."
+            )
+
+        missing_dimensions = set(expected_dimensions).difference(dimension_types)
```
elainethale, line 71: This has the same no-op problem as in one_table_schema, but in this case I think you maybe could have len(dimension_types) > len(expected_dimensions)

dsgrid/config/dataset_schema_handler_standard.py
```diff
@@ -0,0 +1,187 @@
...
+                    f"load_data_lookup records do not match dimension records for {name}"
+                )
+
+    def _check_dataset_time_consistency(self, config: DatasetConfig, load_data_df):
+        time_dim = config.get_dimension(DimensionType.TIME)
+        time_dim.check_dataset_time_consistency(load_data_df)
+        self._check_dataset_time_consistency_by_id(time_dim, load_data_df)
+
+    def _check_dataset_time_consistency_by_id(self, time_dim, load_data_df):
+
```
elainethale, line 98: nit: remove this blank line?

elainethale (approved):
Looks good. A few ticky-tack things to clean up and then will be ready to merge.

dsgrid/config/dataset_config.py
daniel-thom: As of now the standard data schema requires a timestamp column. This was a simplification we made since we expect most datasets to follow this convention. If we will support a dataset with a different layout then we would need to make a new schema type - that could be completely generic or specific.

We could change the standard schema now at a cost of a few development days.

dsgrid/config/dataset_config.py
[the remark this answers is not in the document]
daniel-thom: Pydantic receives a dict and assumes no ordering.

daniel-thom (approved):
Nice refactor!

PR #100, opened 2022-01-19
Support TEMPO time representation

daniel-thom:
This is not fully complete because I don't have a complete set of TEMPO data and configs. Lixi needs to build off this branch, however. I'm OK with leaving this open until we get TEMPO data.

Update: It's now critical to get this merged into main in order to move forward with TEMPO. I disabled the code relating to the TEMPO dataset that is not well-tested. Everything here is well-tested. I'll make a new PR to enable the code whenever we have test data.

### DIFF
--- a/dsgrid/config/annual_time_dimension_config.py
+++ b/dsgrid/config/annual_time_dimension_config.py
@@ -1,0 +1,30 @@
+from datetime import timedelta
+
+from .dimensions import AnnualTimeDimensionModel
+from .time_dimension_base_config import TimeDimensionBaseConfig
+
+
+class AnnualTimeDimensionConfig(TimeDimensionBaseConfig):
+    """Provides an interface to an AnnualTimeDimensionModel."""
+
+    @staticmethod
+    def model_class():
+        return AnnualTimeDimensionModel
+
+    def check_dataset_time_consistency(self, load_data_df):
+        assert False, "not supported yet"
+
+    def convert_dataframe(self, df):
+        return df
+
+    def get_frequency(self):
+        return timedelta(days=365)
+
+    def get_time_ranges(self):
+        assert False, "not supported yet"
+
+    def get_timestamp_load_data_columns(self):
+        return ["year"]
+
+    def get_tzinfo(self):
+        return None
--- a/dsgrid/config/dataset_config.py
+++ b/dsgrid/config/dataset_config.py
@@ -1,3 +1,4 @@
+import abc
 from pathlib import Path
 from typing import List, Optional, Dict, Union
 import os
--- a/dsgrid/config/dataset_config.py
+++ b/dsgrid/config/dataset_config.py
@@ -1,4 +1,3 @@
-import abc
 from pathlib import Path
 from typing import List, Optional, Dict, Union
 import os
@@ -414,7 +413,7 @@ class DatasetConfig(ConfigBase):
         for key, dim_config in self.dimensions.items():
             if key.type == dimension_type:
                 return dim_config
-        assert False, key
+        assert False, dimension_type
 
     def add_trivial_dimensions(self, load_data_lookup):
         """Add trivial 1-element dimensions to load_data_lookup."""
--- a/dsgrid/config/date_time_dimension_config.py
+++ b/dsgrid/config/date_time_dimension_config.py
@@ -65,10 +65,9 @@ class DateTimeDimensionConfig(TimeDimensionBaseConfig):
         val = distinct_counts.collect()[0].distinct_timestamps
         if val != expected_count:
             raise DSGInvalidDataset(
-                f"load_data arrays do not have {len(expected_timestamps)} "  \
+                f"load_data arrays do not have {len(expected_timestamps)} "
                 "timestamps: actual={row.distinct_timestamps}"
             )
-
 
     def convert_dataframe(self, df):
         return df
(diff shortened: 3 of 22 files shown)


dsgrid/config/annual_time_dimension_config.py
```diff
@@ -0,0 +1,30 @@
+from datetime import timedelta
```
elainethale, line 1: Making one module per time config seems like it will swamp everything else in this sub-package. Should we

- Put all of the time configs in one time_dimension_configs.py file OR
- Make a configs/time subpackage

?

dsgrid/config/dataset_config.py
```diff
@@ -1,3 +1,4 @@
+import abc
```
elainethale, line 1: Why is this new import not showing up in another diff line where it is used?

dsgrid/config/date_time_dimension_config.py
```diff
@@ -0,0 +1,102 @@
...
+from dsgrid.exceptions import DSGInvalidDataset
+from .dimensions import DateTimeDimensionModel
+from .time_dimension_base_config import TimeDimensionBaseConfig
+
+
+logger = logging.getLogger(__name__)
+
+
+class DateTimeDimensionConfig(TimeDimensionBaseConfig):
+    """Provides an interface to a DateTimeDimensionModel."""
```
elainethale, line 18: I understand that this name is more consistent with other time packages, but are we going to have to re-register all of the time dimensions again?

Is it time to implement more backward compatibility with respect to software versions?

dsgrid/config/time_dimension_base_config.py
```diff
@@ -0,0 +1,77 @@
...
+        Raises
+        ------
+        DSGInvalidDataset
+            Raised if the dataset is inconsistent with the time dimension.
+
+        """
+
+    @abc.abstractmethod
+    def convert_dataframe(self, df):
+        """Convert a dataframe to standard timestamps.
```
elainethale, line 26: What are "standard" timestamps?

dsgrid/dimension/standard.py
```diff
@@ -122,19 +122,19 @@ class EnergyEfficiency(MetricDimensionBaseModel):
...
 
 
-class DayType(TimeDimensionModel):
+class DayType(DateTimeDimensionModel):
     """Day Type attributes"""
 
 
-class Season(TimeDimensionModel):
+class Season(DateTimeDimensionModel):
     """Season attributes"""
```
elainethale, line 138: I am confused as to why Timezone, DayType, and Season were deriving from TimeDimensionModel (and are now deriving from DateTimeDimensionModel)

dsgrid/registry/project_registry_manager.py
```diff
@@ -232,6 +239,37 @@ def _submit_dataset(
             project_config.config_id,
         )
 
+    def check_dataset_base_to_base_mappings(self, project_config, dataset_config, dataset):
```
elainethale, line 242: Should this be check_dataset_base_to_project_base_mappings?

dsgrid/config/dataset_config.py
daniel-thom: This needs to be deleted.

dsgrid/config/time_dimension_base_config.py
daniel-thom: That is a weak description. I'll change it to something like python datetime timestamps.

dsgrid/dimension/standard.py
daniel-thom: I performed a global swap of TimeDimensionModel to DateTimeDimensionModel. DayType and Season are actually duplicate of enums defined elsewhere. Timezone is unused. These don't appear to make sense with our current implementation. If we want to allow users to specify season as a time dimension then we'll need to make a new data model. We should probably delete these until someone needs one.

dsgrid/registry/project_registry_manager.py
daniel-thom: Yes

dsgrid/config/annual_time_dimension_config.py
daniel-thom: @elainethale I think this item is resolved per our conversation, but let me know if you disagree.

dsgrid/config/date_time_dimension_config.py
```diff
@@ -0,0 +1,102 @@
...
+            return
+
+        for row in timestamps_by_id.collect():
+            if row.distinct_timestamps != len(expected_timestamps):
+                logger.error(
+                    "load_data ID=%s does not have %s timestamps: actual=%s",
+                    row.id,
+                    len(expected_timestamps),
+                    row.distinct_timestamps,
+                )
```
lixiliu, line 66: This is specific to the STANDARD dataset schema and needs to be handled by the handler.

dsgrid/config/date_time_dimension_config.py
```diff
@@ -0,0 +1,102 @@
...
+
+        raise DSGInvalidDataset(
+            f"One or more arrays do not have {len(expected_timestamps)} timestamps"
+        )
```
lixiliu, line 70: missing an indentation?

dsgrid/config/date_time_dimension_config.py
daniel-thom: Correct. You will have to move this code into your new handler.

dsgrid/config/date_time_dimension_config.py
daniel-thom: No, it's actually correct. The good case returns on line 57. The for loop just above logs all errors. By the time it gets here, it must raise an exception. I guess I should reconsider the structure.

elainethale (approved):
Looks like my previous comments have been addressed. I'm fine moving forward if that's what makes sense at the moment.

Issue #98, opened 2022-01-18
Dim mappings do not get checked against dim records


lixiliu:
Currently dimension mappings do not get checked against the records specified in their configs. Mappings could get registered successfully when dimension records do not match those in the mappings.

elainethale:
#101 fixed this issue

PR #101, opened 2022-01-22
Check validity of dimension mapping records

daniel-thom:
Fixes GitHub Issue #98


dsgrid/registry/dimension_mapping_registry_manager.py
```diff
@@ -103,6 +107,27 @@ def check_unique_records(self, config: DimensionMappingsConfig, warn_only=False)
                     f"There are {len(duplicates)} dimension mappings with duplicate content (data files)."
                 )
 
+    def _check_records_against_dimension_records(self, config):
+        for mapping in config.model.mappings:
+            actual_from_records = {x.from_id for x in mapping.records}
+            actual_to_records = {x.to_id for x in mapping.records}
+            from_dimension = self._dimension_mgr.get_by_id(mapping.from_dimension.dimension_id)
+            allowed_from_records = from_dimension.get_unique_ids()
+            diff = actual_from_records.difference(allowed_from_records)
```
lixiliu, line 116: Do we only care the excess difference from `actual` to `allowed`? What about when `actual` is smaller than `allowed`, (i.e. A.symmetric_difference(B))?

Same goes for the difference for `to_dimension` below.

elainethale: I think actual can be smaller than allowed, but I wonder if we want the mapping (in its config) to specify that it represents a subset if that's the case?

elainethale (commented):
Generally looks good. One small point for discussion about whether we should be checking/noting if mappings do not map all dimension elements and/or providing a config flag where people can acknowledge that they are mapping a subset.

dsgrid/registry/dimension_mapping_registry_manager.py
daniel-thom: My takeaway from today's meeting was that a dimension mapping record file must include ALL records from the dimension (Lixi's solution). Implemented.

dsgrid/registry/dimension_mapping_registry_manager.py
```diff
@@ -103,6 +107,27 @@ def check_unique_records(self, config: DimensionMappingsConfig, warn_only=False)
...
+        for mapping in config.model.mappings:
+            actual_from_records = {x.from_id for x in mapping.records}
+            actual_to_records = {x.to_id for x in mapping.records}
+            from_dimension = self._dimension_mgr.get_by_id(mapping.from_dimension.dimension_id)
+            allowed_from_records = from_dimension.get_unique_ids()
+            diff = actual_from_records.symmetric_difference(allowed_from_records)
+            if diff:
+                dim_id = from_dimension.model.dimension_id
+                raise DSGInvalidDimensionMapping(
+                    f"Dimension mapping has invalid 'from' records: dimension_id={dim_id} {diff}"
```
elainethale, line 120: "Dimension mapping has invalid and/or missing 'from' records: dimension_id={dim_id} {diff}"

Am I reading this correctly that diff will list both the invalid elements and the missing elements?

dsgrid/registry/dimension_mapping_registry_manager.py
```diff
@@ -103,6 +107,27 @@ def check_unique_records(self, config: DimensionMappingsConfig, warn_only=False)
...
+                )
+            to_dimension = self._dimension_mgr.get_by_id(mapping.to_dimension.dimension_id)
+            allowed_to_records = to_dimension.get_unique_ids()
+            diff = actual_to_records.symmetric_difference(allowed_to_records)
```
elainethale, line 124: Sorry, is this only used in cases in which all 'to' dimension records are required? 

I think we landed on the notion that all mappings should list all from records, but may subset relative to the to-dimension. This looks like a pretty low-level function--should it have a kwarg to specify whether this check should be symmetric or should instead only look for invalid record ids?

elainethale: Presumably the association table aware-subsetting of the to-dimension will happen elsewhere (and in a different PR?)

elainethale (commented):
I think the to-dimension logic may still be incorrect ...

dsgrid/registry/dimension_mapping_registry_manager.py
daniel-thom: `diff` will list the elements that are present in only one of the sets. The message needs to be corrected.

dsgrid/registry/dimension_mapping_registry_manager.py
daniel-thom: I fixed the error messages. The purpose of this PR is to fix a very specific issue. While it's possible that this function could be used for some other purpose, that is beyond the scope of what I want to do here.

lixiliu: If `to_id` can be a subset of the registered dim it's assigned to, then it might lead to the following case:
Suppose we have this mapping:
```
from_id -> to_id
A->i
B->i
C->j
D->k
```

If Project 1 wants {I, j, k, l} for the to_dimension and Project 2 wants {i, j, k, l, m, n}. What should the `to_id` be referred to in the mapping? 
Should each project registers the same mapping but with different to_id? Or should the mapping have its own `to_id` dimension registered and have an association to project?

lixiliu (approved):

dsgrid/registry/dimension_mapping_registry_manager.py
mooneyme: Lixi - that case is actually something we will want. In that case with the two i's, we would need to aggregate them together.

dsgrid/registry/dimension_mapping_registry_manager.py
daniel-thom: Implemented per our email discussion. Since dimension mappings are not tied to dimension associations, it can cause problems for users.

daniel-thom: An alternative that I'm sure no one will like is to require the dataset contributor to add `from=null, to=X` for all records that do not belong to their data source.

dsgrid/registry/dimension_mapping_registry_manager.py
```diff
@@ -103,6 +107,31 @@ def check_unique_records(self, config: DimensionMappingsConfig, warn_only=False)
...
+            # mapping to a project's dimension for a specific data source, but that information
+            # is not available here.
+            to_dimension = self._dimension_mgr.get_by_id(mapping.to_dimension.dimension_id)
+            allowed_to_records = to_dimension.get_unique_ids()
+            diff = actual_to_records.difference(allowed_to_records)
+            if diff:
+                dim_id = from_dimension.model.dimension_id
+                raise DSGInvalidDimensionMapping(
+                    f"Dimension mapping has invalid 'to' records: dimension_id={dim_id} {diff}"
+                )
```
elainethale, line 133: I think the diff checking may need to be modified slightly to allow for null values? How are nulls going to show up in the actual_from_records and actual_to_records sets? Will they be excluded to begin with, show up as None, or as something else?

elainethale (approved):
One last question about null values. Otherwise ready to go.

dsgrid/registry/dimension_mapping_registry_manager.py
mooneyme: Agreed. I just tested this for NULL values in the to-values and it fails.

tests/test_registry_management.py
```diff
@@ -184,6 +185,42 @@ def test_auto_updates(make_test_project_dir):
...
+        # Invalid 'to' record
+        record_file.write_text(orig_text.replace("CO", "Colorado"))
+        with pytest.raises(DSGInvalidDimensionMapping):
+            dim_mapping_mgr.register(dimension_mapping_file, user, log_message)
+
+        # Valid
+        record_file.write_text(orig_text)
+        dim_mapping_mgr.register(dimension_mapping_file, user, log_message)
+
+
```
mooneyme, line 223: I just tested a valid 'to' record that has NULL values in the test and it fails at line 130 in `dimension_mapping_registry_manager.py` because `diff` is not None. 

This is what I added to the test to check this:
```
        # Valid 'to' record with NULLs
        record_file.write_text(orig_text.replace("CO", ""))
        dim_mapping_mgr.register(dimension_mapping_file, user, log_message)
```

mooneyme (changes requested):
I agree with Elaine that we need to support NULLs in the to-values

tests/test_registry_management.py
mooneyme: @elainethale  can you think of any other things we want to check for with these mappings? Like one thing that comes to mind is: `"What if we have something in `from` mapped to two different `to` values?"`. I think we will eventually support this, but right now we have no way of properly disaggregating/splitting the 'from' across two 'to's.

Anything else come to mind?

dsgrid/registry/dimension_mapping_registry_manager.py
daniel-thom: Fixed

tests/test_registry_management.py
mooneyme: I checked the scenario of `"What if we have something in 'from' mapped to two different 'to' values?"` and it fails in the tests, however I get a really confusing error message that doesn't seem to fit:`dsgrid.exceptions.DSGDuplicateValueRegistered: There are 3 dimension mappings with duplicate content (data files).`

This is the test I applied:
```
        # what if we have Denver mapped to 2 different 'to' values?
        record_file.write_text(orig_text+ "08031,TX\n")
        dim_mapping_mgr.register(dimension_mapping_file, user, log_message)
        ```

Comment (elainethale) [reply tests/test_registry_management.py]: Since to support multiple "to" values we will have to provide more data (e.g., fraction to apply to each "to") I'm glad that fails for now. I agree that the current error message is not enough to go on to figure out the problem. Is it easy to at least list what file is failing? (And I don't know why that message says that 3 mappings have duplicate content.)

I'm failing to think of any other failure conditions we want to check now, but it's also the end-ish of a long day.

Comment (daniel-thom) [reply tests/test_registry_management.py]: @mooneyme Did you do something like this?
```
        record_file.write_text(orig_text.replace("CO", ""))
        dim_mapping_mgr.register(dimension_mapping_file, user, log_message)

        record_file.write_text(orig_text+ "08031,TX\n")
        dim_mapping_mgr.register(dimension_mapping_file, user, log_message)
```
If so, the error message mostly makes sense. You are trying to register the same dimensions multiple times. I think that you actually wanted this (one attempt at registration with your invalid data):
```
        record_file.write_text(orig_text.replace("CO", ""))
        record_file.write_text(orig_text+ "08031,TX\n")
        dim_mapping_mgr.register(dimension_mapping_file, user, log_message)
```
This passes. We aren't checking for duplicate values in the from column. That should be a simple fix.

Comment (daniel-thom) [reply tests/test_registry_management.py]: @mooneyme You wrote this no-op function. I may understand what you mean for multiple to_id values. The user may want to map two different dataset records to the same project record. Why would there be multiple instances of one from record? 
```
    def validate_records(self, config: DimensionMappingsConfig, warn_only=False):
        """Validate dimension mapping records.

        Check:
        - from_id and to_id column names
        - check for duplicate IDs and log a warning if they exist (sometimes we want them to exist if it is an aggregation)
        """
        pass
```

daniel-thom: I think now is a good time for you all to write down the behaviors you would like to see in these cases. I suspect that our current data structures are insufficient to handle all of the cases.

mooneyme: @daniel-thom 

We want to support multiple "to" values to the same "from" value. Examples:
1. Many counties map to a single state
2. TEMPO has higher fidelity of subsectors that we want to collapse at the project level

We do not want to support the SAME "to" value mapped to multiple "from" values. _At least not yet._ We are not ready for this type of disaggregation behavior because would we need to support logic for how to split out the "to" value across multiple "from" values. --> It sounds like currently dsgrid is not checking for this, but we want to fail for this at this point in time.

daniel-thom: To confirm: duplicate to_id values (except null) are not allowed. Duplicate from_ids are allowed. Those are obviously quite prevalent in the base-to-supplemental mappings.

Will we want different logic based on whether it is a dataset base-to-project mapping vs project base-to-supplemental mapping?

mooneyme: Sorry for the confusion, I updated the language to talk about Aggregations vs. Disaggregations. :)

daniel-thom: This is fixed.

mooneyme (approved):
Thanks :)

PR #102, opened 2022-01-24
Full Join Logic for Dimension Mappings

daniel-thom:
Completes JIRA Story DSGRID-186

This PR also fixes handling of dimension associations. The associations registered to existing projects did not work because the data files were not copied to the registry. We can patch this by manually uploading the files to S3.


dsgrid/config/project_config.py
```diff
@@ -49,7 +57,7 @@ class DimensionsModel(DSGBaseModel):
             " one dimension reference per type is allowed.",
         ),
     )
-    supplemental_dimensions: Optional[List[DimensionReferenceModel]] = Field(
+    supplemental_dimensions: List[DimensionReferenceModel] = Field(
```
daniel-thom, line 60: Many of our models unnecessarily define types with `Optional`. If a default value is provided then we don't want them declared `Optional`. That type means that it can be `None` or the real type.

dsgrid/config/project_config.py
```diff
@@ -158,12 +165,11 @@ def check_model_sector(cls, model_sector, values):
 class DimensionMappingsModel(DSGBaseModel):
     """Defines all dimension mappings associated with a dsgrid project, including base-to-base, base-to-supplemental, and dataset-to-project mappings."""
 
-    base_to_base: Optional[List[DimensionMappingReferenceModel]] = Field(
+    base_to_base: List[DimensionMappingReferenceModel] = Field(
```
daniel-thom, line 168: @mooneyme Is this field now defunct?

dsgrid/registry/project_registry_manager.py
```diff
@@ -239,36 +247,84 @@ def _submit_dataset(
...
+        else:
+            raise Exception(f"not supported yet: {dataset_config.model.data_schema}")
+
+        needs_full_join = set()
+        types = [x for x in DimensionType if x not in exclude_dims]
+        for type1, type2 in itertools.product(types, types):
+            if type1 != type2:
+                needs_full_join.add((type1, type2))
+
+        for type1, type2 in needs_full_join:
```
daniel-thom, line 285: @mooneyme Please confirm the logic here.

dsgrid/registry/project_registry_manager.py
```diff
@@ -239,36 +247,84 @@ def _submit_dataset(
...
+                iter(
+                    records.join(lookup, on=[type1.value, type2.value])
+                    .agg(F.count_distinct(type1.value, type2.value))
+                    .collect()[0]
+                    .asDict()
+                    .values()
+                )
+            )
+            if count != records.count():
+                # There is probably a better way to print the rows dropped from the join above.
```
daniel-thom, line 308: If someone has a better suggestion, please raise it.

dsgrid/config/project_config.py
```diff
@@ -289,6 +306,143 @@ def dimension_mapping_manager(self):
...
+        -------
+        DimensionConfig
+
+        """
+        return [v for k, v in self.supplemental_dimensions.items() if k.type == dimension]
+
+    def get_base_to_base_dimension_mapping_by_types(
+        self, from_dimension: DimensionType, to_dimension: DimensionType
+    ):
+        """Return the base-to-base dimension mapping matching the dimension types.
```
elainethale, line 343: Are we calling these association mappings now (instead of base-to-base mappings)? (Maybe related to @daniel-thom 's question to @mooneyme about a possibly defunct-base-to-base field?)

dsgrid/config/project_config.py
```diff
@@ -289,6 +306,143 @@ def dimension_mapping_manager(self):
...
+        """
+        return [
+            x
+            for x in self._base_to_supplemental_mappings.values()
+            if x.model.from_dimension.dimension_type == dimension
+        ]
+
+    def has_base_to_base_dimension_mapping_types(self, from_dimension, to_dimension):
+        """Return True if the config has these base-to-base mappings."""
+        return self._has_mapping(from_dimension, to_dimension, self._base_to_base_mappings)
```
elainethale, line 386: Flagging another instance of base-to-base

elainethale: Looking farther down, we do of course have dataset-base to project-base mappings. Then I think we're calling within project base-to-base mappings "association mappings" instead. Is that correct? 

Without looking a lot closer I can't tell which kind of mapping (of the two I just listed) is meant here.

dsgrid/registry/project_registry_manager.py
```diff
@@ -302,3 +358,22 @@ def remove(self, config_id):
...
+        # This method could be moved elsewhere.
+        for ref in mapping_references:
+            column = ref.from_dimension_type.value
+            if column in df.columns:
+                mapping_config = self._dimension_mapping_mgr.get_by_id(
+                    ref.mapping_id, version=ref.version
+                )
+                records = models_to_dataframe(mapping_config.model.records)
+                df = (
+                    df.join(records, getattr(df, column) == records.from_id)
```
elainethale, line 374: This tells me we're dropping values in df not found in the mapping.

dsgrid/registry/project_registry_manager.py
```diff
@@ -302,3 +358,22 @@ def remove(self, config_id):
...
+                    .drop("from_id")
+                    .drop(column)
+                    .withColumnRenamed("to_id", column)
```
elainethale, line 377: And here we're just switching to using to_id as our identifier. It could be that the mapping represents an aggregation; I agree that for this use case we don't care/don't want to actually deal with duplicates, but if/when this moves elsewhere we'll probably want to encode somewhere (the mapping_config?) whether we want to perform a sum or some other aggregation to collapse duplicates when we are done.

elainethale (approved):
Looks good to me. I have some clarification questions about base-to-base mappings. Otherwise I am super happy to see these features go in.

Note that I have not tried to run the tests myself nor did I go too deep into the queries/logic that is doing the actual checking (although I did spend some time reading them to get a basic sense of the intention and methods).

dsgrid/config/project_config.py
daniel-thom: Here is my interpretation of things:
- A `dimension mapping` maps a `from_id` to a `to_id`. Example: dataset calls a dimension record something slightly different than the project.
- A `dimension association` in the project.toml specifies what combinations of records are required to be present in the load data table. This association can have two or more columns. If the dataset does have a given combination then it must still have a row for the combination, but the ID is null.

I saw that @mooneyme deleted base-to-base mappings in the project files and am assuming that those are now obsolete.

dsgrid/registry/project_registry_manager.py
daniel-thom: My addition in https://github.com/dsgrid/dsgrid/pull/101 requires that a dimension mapping include all records, so this code is currently correct. However, I should have included an assert to make sure the length didn't change.

I actually think my earlier change is incorrect. We don't need to require that all records are covered in the mapping.

dsgrid/registry/project_registry_manager.py
lixiliu: If lookup is a subset of records, some flavor of this might work:
`records[[type1.value, type2.value]].except(lookup[[type1.value, type2.value]])`

dsgrid/registry/project_registry_manager.py
```diff
@@ -239,36 +247,84 @@ def _submit_dataset(
...
+        for type1, type2 in needs_full_join:
+            if project_config.has_dimension_association(type1, type2):
+                records = project_config.get_dimension_association_records(type1, type2)
+            else:
+                pdim1 = project_config.get_base_dimension(type1)
+                pdim1_ids = pdim1.get_unique_ids()
+                pdim2 = project_config.get_base_dimension(type2)
+                pdim2_ids = pdim2.get_unique_ids()
+                records = itertools.product(pdim1_ids, pdim2_ids)
+                records = create_dataframe_from_dimension_ids(
```
lixiliu, line 294: Tip: if `pdim1_ids` and `pdim12_ids` happen to be a col from other dataframes, you can do `pdim1_ids.crossJoin(pdim2_ids)` to get the product dataframe.

dsgrid/config/project_config.py
lixiliu: A project is a combination of datasets minus data it does not want. If we have to generate a `dimension association` from datasets anyway, then I think it's easier and safer for project to specify what it does not want from each dataset and have the association be generated based on that.

mooneyme: That is right Dan.

dsgrid/config/project_config.py
mooneyme: this is now what we are calling an association

dsgrid/registry/project_registry_manager.py
mooneyme: Yes agreed Elaine.

daniel-thom: @elainethale I don't think this comment is relevant to this PR (or at least there are no changes associated with it), so I'm resolving it. I'll assume that you or Meghan will make appropriate plans for any changes to mapping configs.

dsgrid/config/project_config.py
daniel-thom: All references to base-to-base mappings will be deleted in the next push to this PR.

tests/test_dataset_registration.py
```diff
@@ -19,8 +19,12 @@
 from dsgrid.tests.make_us_data_registry import make_test_data_registry, replace_dataset_path
 
 
+PROJECT_ID = "test_efs"
```
mooneyme, line 22: Do these tests belong in the test_dataset_registration.py? Because really we need to test this at dataset submission to project. Right?

daniel-thom: It is running the submit_dataset command. I'll rename the file.

tests/data/dimension_associations/subsector__metric.csv
```diff
@@ -0,0 +1,131 @@
+subsector,metric
+FullServiceRestaurant,cooling
+FullServiceRestaurant,heating
```
mooneyme, line 3: Why do you have associations in dsgrid and not dsgrid-test-data?

daniel-thom: dsgrid-test-data has toy versions of projects and dataset that comprise real registries. This is just a few independent files to test the DimensionAssociation class by itself.

dsgrid/config/dataset_schema_handler_base.py
```diff
@@ -126,3 +217,24 @@ def _check_for_duplicates_in_cols(lst: list, table_name: str):
...
+        for ref in mapping_refs:
+            column = ref.from_dimension_type.value
+            if column in df.columns:
+                mapping_config = self._dimension_mapping_mgr.get_by_id(
+                    ref.mapping_id, version=ref.version
+                )
+                records = models_to_dataframe(mapping_config.model.records)
+                orig = df.count()
+                df = (
+                    df.join(records, df[column] == records.from_id)
```
lixiliu, line 234: I would add how="left". Default is an inner join (https://spark.apache.org/docs/3.1.1/api/python/reference/api/pyspark.sql.DataFrame.join.html), if `from_id` is partial of a col, df.count() will shrink.

dsgrid/config/dataset_schema_handler_one_table.py
```diff
@@ -79,3 +83,12 @@ def _check_one_table_data_consistency(self, config: DatasetConfig, load_data):
...
                 )
+
+    def get_unique_dimension_rows(self, mapping_references):
+        time_dim = self._config.get_dimension(DimensionType.TIME)
+        time_cols = set(time_dim.get_timestamp_load_data_columns())
+        pivoted_cols = set(self.get_pivot_dimension_columns())
+        exclude = time_cols.union(pivoted_cols)
+        dim_cols = [x for x in self._load_data.columns if x not in exclude]
+        df = self._load_data.select(*dim_cols).distinct()
+        return self._remap_dimension_columns(df, mapping_references)
```
lixiliu, line 94: Since there's no collapsing/aggregation in _remap_dimension_columns, this will contain duplicates if to_id in maps contain duplicates, which seem to go against what the name of the func implies

dsgrid/registry/project_registry_manager.py
```diff
@@ -239,36 +279,69 @@ def _submit_dataset(
...
+        records = itertools.product(pdim1_ids, pdim2_ids)
+        return create_dataframe_from_dimension_ids(records, type1, type2)
+
+    @staticmethod
+    def _check_pivot_dimension_columns(project_config, handler, dim_type, mapping_refs):
+        d_dim_ids = set(
+            handler.get_pivot_dimension_columns_mapped_to_project(mapping_refs).values()
+        )
+        p_dim_ids = project_config.get_base_dimension(dim_type).get_unique_ids()
+        diff = d_dim_ids.difference(p_dim_ids)
```
lixiliu, line 342: We want symmetric_difference here right since we are checking that the mapped dimensions match exactly what's expected at project?

lixiliu: If not symmetric_difference, I think it's `diff = p_dim_ids.difference(d_dim_ids)`

dsgrid/config/dataset_schema_handler_base.py
```diff
@@ -31,6 +91,34 @@ def get_pivot_dimension_columns(self):
...
+        """
+        columns = set(self.get_pivot_dimension_columns())
+        dim_type = self.get_pivot_dimension_type()
+        for ref in mapping_references:
+            if ref.from_dimension_type.value == dim_type:
+                mapping_config = self._dimension_mapping_mgr.get_by_id(
+                    ref.mapping_id, version=ref.version
+                )
+                mapping = {
+                    x.from_id: x.to_id for x in mapping_config.model.records if x.to_id is not None
```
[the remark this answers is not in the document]
daniel-thom: If `to_id` is None then the record cannot be present in the pivoted columns. The dict returned from this method will only include actual columns. This is just for the check on line 114. Let me know if I'm missing something.

I do think there is a bug here. `to_id` may be `""` instead of `None` because the `AssociationTableRecordModel` defines the field as a string. I think we need to coerce these values to `None`.

dsgrid/config/dataset_schema_handler_base.py
daniel-thom: In one of our meetings last week we discussed this and decided that we do want an inner join. The mapping should include all records. Note that line 239 guards against this point.

dsgrid/config/dataset_schema_handler_one_table.py
daniel-thom: Yeah, this is a bug. It should be sufficient to append a `.distinct()` to the end of this, right?

dsgrid/config/dataset_schema_handler_base.py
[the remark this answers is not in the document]
lixiliu: I meant to say that `diff = columns.symmetric_difference(mapping)` should be done before anything is removed from the mapping. Also, shouldn't columns be compared against the `from_id` only in the mapping?

dsgrid/config/dataset_schema_handler_base.py
lixiliu: Right I saw line 239. If records are already checked to ensure that all from_id exists, then that makes sense.

dsgrid/registry/project_registry_manager.py
daniel-thom: I'm not sure how I messed this one up. The exact project-required records are known here. I'll fix.

dsgrid/config/project_config.py
```diff
@@ -289,6 +302,79 @@ def dimension_mapping_manager(self):
...
+        dimension_type : DimensionType
+
+        Returns
+        -------
+        DimensionConfig
+
+        """
+        return [v for k, v in self.supplemental_dimensions.items() if k.type == dimension]
+
+    def get_base_to_supplemental_dimension_mappings_by_types(self, dimension: DimensionType):
```
elainethale, line 336: Everywhere we have dimension: DimensionType, shouldn't the name be dimension_type?

dsgrid/config/project_config.py
```diff
@@ -289,6 +302,79 @@ def dimension_mapping_manager(self):
...
+            for x in self._base_to_supplemental_mappings.values()
+            if x.model.from_dimension.dimension_type == dimension
+        ]
+
+    def has_base_to_supplemental_dimension_mapping_types(self, dimension):
+        """Return True if the config has these base-to-supplemental mappings."""
+        return self._has_mapping(dimension, dimension, self._base_to_supplemental_mappings)
+
+    @staticmethod
+    def _has_mapping(from_dimension, to_dimension, mapping):
```
elainethale, line 360: _has_mapping(from_dimension_type, to_dimension_type, mappings):

?

dsgrid/registry/project_registry_manager.py
```diff
@@ -239,36 +279,69 @@ def _submit_dataset(
...
+        pivot_dimension = handler.get_pivot_dimension_type()
+        exclude_dims = set([DimensionType.TIME, DimensionType.DATA_SOURCE, pivot_dimension])
+        types = [x for x in DimensionType if x not in exclude_dims]
+        dimension_pairs = [tuple(sorted((x, y))) for x, y in itertools.combinations(types, 2)]
+
+        data_source = dataset_config.model.data_source
+        associations = project_config.dimension_associations
+        dim_table = handler.get_unique_dimension_rows(mapping_references)
+        for type1, type2 in dimension_pairs:
+            records = associations.get_associations_by_data_source(data_source, type1, type2)
```
elainethale, line 302: Will this grab the right associations if we have something like:

| data_source | sector |
|------------- | ------ |
| resstock | res |

| sector | subsector |
| ------ | ---------- |
| res      | SingleFamily |
| res     | MultiFamiliy |

?

dsgrid/utils/files.py
```diff
@@ -72,6 +72,7 @@ def load_data(filename, **kwargs):
             data = mod.load(f_in)
         except Exception:
             logger.exception("Failed to load data from %s", filename)
+            breakpoint()
```
elainethale, line 75: Seems like this needs to be removed before merging?

tests/data/dimension_associations/model_year__metric.csv
```diff
@@ -0,0 +1,11 @@
...
+2012,cooling
+2012,heating
+2012,water_systems
+2012,exterior_lighting
+2012,fans
+2012,heat_rejection
+2012,heating
+2012,interior_equipment
+2012,interior_lighting
+2012,pumps
```
elainethale, line 11: Why do we need associations for model_years?

tests/data/dimension_associations/scenario__subsector.csv
```diff
@@ -0,0 +1,14 @@
...
+baseline,LargeOffice
+baseline,MediumOffice
+baseline,Outpatient
+baseline,PrimarySchool
+baseline,QuickServiceRestaurant
+baseline,StandaloneRetail
+baseline,StripMall
+baseline,SmallHotel
+baseline,SmallOffice
+baseline,Warehouse
```
elainethale, line 14: This seems redundant with the scenario,data_source and data_source,subsector associations. Is that what's intended?

dsgrid/config/dataset_schema_handler_one_table.py
```diff
@@ -18,28 +17,33 @@
...
     def check_consistency(self):
-        path = Path(self._config.model.path)
-        load_data_df = read_dataframe(check_load_data_filename(path), cache=True)
         with Timer(timer_stats_collector, "check_dataset_time_consistency"):
-            self._check_dataset_time_consistency(self._config, load_data_df)
-        load_data_df = self._config.add_trivial_dimensions(load_data_df)
+            self._check_dataset_time_consistency(self._load_data)
         with Timer(timer_stats_collector, "check_one_table_data_consistency"):
-            self._check_one_table_data_consistency(self._config, load_data_df)
+            self._check_one_table_data_consistency()
```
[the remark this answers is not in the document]
daniel-thom: 1. I think there may have been more than one function call at some point in time.
2. You get to choose the name that shows up in the logs with `Timer`. I may have preferred no underscore. 

I'm open to changing these now.

dsgrid/config/dimension_associations.py
```diff
@@ -0,0 +1,165 @@
...
+
+        return cls(associations)
+
+    def get_associations(self, *dimensions):
+        """Return the records for this dimension association.
+
+        Parameters
+        ----------
+        dimensions : tuple
+            Any number of instances of DimensionType
```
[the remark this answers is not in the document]
daniel-thom: This file does not use `dimension_type`. I used `dimension` here for brevity because there shouldn't be any confusion with `dimension_type` vs `dimension_config`, which could exist in other files. I still prefer the shorter name but am open to changing it.

dsgrid/config/dimension_associations.py
```diff
@@ -0,0 +1,165 @@
...
+        # Try a direct match and then check for a subset.
+        table = self._associations.get(tuple(sorted(dimensions)))
+        if table is not None:
+            return table
+
+        caller_dims = set(dimensions)
+        for association_dims, table in self._associations.items():
+            if not caller_dims.difference(association_dims):
+                # This assumes that if there are multiple matches, all tables have the same data.
+                return table.select(*(x.value for x in dimensions))
```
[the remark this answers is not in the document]
daniel-thom: All keys are already tuples. The point of this code block is the case where the files have 3 dimensions but the caller asks for 2 and those 2 dimensions exist in 2 different files. I'm a bit concern that there is some corner-case problem not covered.

dsgrid/config/dimension_associations.py
```diff
@@ -0,0 +1,165 @@
...
+
+        filter_col = filter_dim.value
+        for dimension in dimensions:
+            tmp = self.get_associations(filter_dim, dimension)
+            if tmp is None:
+                raise DSGInvalidDimensionAssociation(
+                    f"No association is stored for {filter_dim} and {dimension}"
+                )
+            table = (
+                table.join(tmp, dimension.value)
```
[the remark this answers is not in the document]
daniel-thom: Hmm. Would it be an error if any records are dropped by an inner join? I could be missing something but it seems like that shouldn't happen?

dsgrid/config/project_config.py
daniel-thom: I'm not sure I agree with the statement globally, but I do agree here. If the line length limit forced this to be
```
def get_base_to_supplemental_dimension_mappings_by_types(
    self,
    dimension: DimensionType
):
```
then I would want to use the shorter name.

dsgrid/registry/project_registry_manager.py
daniel-thom: Does this answer your question? Also, you can directly load the dimension associations for StandardScenarios without project or datasets if you'd like to experiment.
```
(Pdb) self.get_associations(DimensionType.DATA_SOURCE, DimensionType.SECTOR).show()
+-----------+------+
|data_source|sector|
+-----------+------+
|   comstock|   com|
|   comstock|   res|
+-----------+------+
(Pdb) self.get_associations(DimensionType.SECTOR, DimensionType.SUBSECTOR).show()
+------+--------------------+
|sector|           subsector|
+------+--------------------+
|   com|FullServiceRestau...|
|   com|            Hospital|
|   com|          LargeHotel|
|   com|         LargeOffice|
|   com|        MediumOffice|
|   com|          Outpatient|
|   com|       PrimarySchool|
|   com|QuickServiceResta...|
|   com|          SmallHotel|
|   com|         SmallOffice|
|   com|    StandaloneRetail|
|   com|           StripMall|
|   com|           Warehouse|
|   res|    MidriseApartment|
+------+--------------------+
(Pdb) self.get_associations_by_data_source("comstock", DimensionType.SECTOR, DimensionType.SUBSECTOR).show()
+--------------------+------+
|           subsector|sector|
+--------------------+------+
|FullServiceRestau...|   com|
|            Hospital|   com|
|          LargeHotel|   com|
|         LargeOffice|   com|
|        MediumOffice|   com|
|          Outpatient|   com|
|       PrimarySchool|   com|
|QuickServiceResta...|   com|
|          SmallHotel|   com|
|         SmallOffice|   com|
|    StandaloneRetail|   com|
|           StripMall|   com|
|           Warehouse|   com|
|    MidriseApartment|   res|
+--------------------+------+
```

dsgrid/utils/files.py
daniel-thom: thanks

tests/data/dimension_associations/model_year__metric.csv
daniel-thom: I copied everything from StandardScenarios for test purposes - not necessarily because it makes sense for this project. Maybe I should remove some of the associations.

tests/test_dimension_associations.py
```diff
@@ -0,0 +1,58 @@
...
+
+def test_dimension_associations(spark_session):
+    da = DimensionAssociations.load(Path("."), DATA_DIR.iterdir())
+    assert da.has_associations(DimensionType.SECTOR, DimensionType.SUBSECTOR)
+    assert not da.has_associations(DimensionType.SECTOR, DimensionType.TIME)
+    assert da.get_associations(DimensionType.SECTOR, DimensionType.SUBSECTOR).count() == 27
+    assert (
+        da.get_associations_by_data_source(
+            "comstock", DimensionType.SECTOR, DimensionType.SUBSECTOR
+        ).count()
```
mooneyme, line 20: @daniel-thom what if I want to get all the associations for a specific data_source? Where are we doing the join?
`da.get_associations_by_data_source("comstock").show()` is an empty table.

tests/test_dimension_associations.py
```diff
@@ -0,0 +1,58 @@
...
+from dsgrid.dimension.base_models import DimensionType
+
+
+DATA_DIR = Path("tests") / "data" / "dimension_associations"
+
+
+def test_dimension_associations(spark_session):
+    da = DimensionAssociations.load(Path("."), DATA_DIR.iterdir())
+    assert da.has_associations(DimensionType.SECTOR, DimensionType.SUBSECTOR)
+    assert not da.has_associations(DimensionType.SECTOR, DimensionType.TIME)
```
mooneyme, line 15: Even though we don't provide an association table for this, I still want to be able to see which Sectors are in each Geography (or Time). Remember that if it isn't provided, then we assume a full join between the combos of dimensions.

tests/test_dimension_associations.py
```diff
@@ -0,0 +1,58 @@
+import shutil
```
mooneyme, line 1: In this test script, we need to add some tests for bad associations.

tests/test_register_project.py
```diff
@@ -0,0 +1,47 @@
...
+            finally:
+                if test_dir.exists():
+                    shutil.rmtree(test_dir)
+
+
+def _setup_invalid_dimension_associations(project_dir):
+    association_file = project_dir / "dimension_associations" / "sector__subsector.csv"
+    with open(association_file, "a") as f_out:
+        f_out.write("invalid,invalid\n")
+    return DSGInvalidDimensionAssociation, r"Dimension association for .* has invalid records"
```
mooneyme, line 47: Does it make sense that this is here? the return should be in dsgrid code, not in tests, right?

dsgrid/config/dimension_associations.py
```diff
@@ -0,0 +1,172 @@
...
+            for column in records.columns:
+                tmp = column + "tmp_name"
+                records = records.withColumn(tmp, records[column].cast(StringType())) \
+                    .drop(column) \
+                    .withColumnRenamed(tmp, column)
+            types = tuple(DimensionType(x) for x in sorted(records.columns))
+            associations[types] = records
+            logger.debug("Loaded dimension associations from %s %s", path, records.columns)
+
+        return cls(associations)
```
mooneyme, line 58: Shouldn't we be validating the associations before we return?

mooneyme: Also, where is the full join logic happening?

dsgrid/config/dimension_associations.py
```diff
@@ -172,3 +172,21 @@ def iter_associations(self):
         """Yields the stored dimension types with their association tables as a tuple."""
         for dimensions, table in self._associations.items():
             yield dimensions, table
+
+    def get_full_join_by_data_source(self, data_source):
```
daniel-thom, line 176: @elainethale @mooneyme @lixiliu I think that producing a single table of dimension associations is something like this. I'm unsure about `how="cross"` vs `how="outer"`. Results appear the same in limited testing.

I think that we could implement all checks by keeping this separate from the full combinations of project records. We could still iterate over pairs of combinations in project_registry_manager and check if they exist here. If there was somehow implicit relationships that still need to be accounted for, that wouldn't work.

Let me know what you think. If this seems right to you then I can implement the logic today. Thanks.

mooneyme: @daniel-thom here is what I have been playing around with. Given our current test, this works as we would expect. However I think there are a few case study tests we need to test out to make sure it behaves as expected under every potential scenario.

```
    def get_full_join_by_data_source(self, project_config, data_source):
        tables = []
        for dims, val in self._associations.items():
            if DimensionType.DATA_SOURCE in dims:
                continue
            table = self.get_associations_by_data_source(data_source, *dims)
            tables.append(table)

        table = tables[0]
        if len(tables) > 1:
            for i, other in enumerate(tables[1:]):
                print(i)
                on_columns = list(set(other.columns).intersection(table.columns)) # TODO: is there potentially an issue here with the ordering of the tables to lead to a "non-overlap" on_columns?
                if on_columns:
                    table_columns = table.columns
                    table_distinct_cols_count =  table.select(table.columns).distinct().count() 
                    table = table.join(other, on=on_columns, how="outer")
                    assert table.select(other.columns).distinct().count() == other.select(other.columns).distinct().count() 
                    assert table.select(table_columns).distinct().count() == table_distinct_cols_count 
                else:
                    # TODO: need to test this with the right combinations of supplied associations
                    
        # TODO: in our current test using standard scenarios dimensions, we only have 5 dimensions in the columns, but 8 tables are joined together. This has to do with the full circle chain of associations provided. We may be able to save time and skip these as long as we do some sanity checks?? Unclear ATM.
        
        all_dimensions = set(d.value for d in DimensionType)
        missing_dimensions = all_dimensions.difference(table.columns)
        for dim in missing_dimensions:
            if dim not in ("data_source", "time"): # TODO: how do we want to add in time dimension if we don't have records for them?
                table_count = table.count()
                # TODO: @dtom is there a better/preferred way to get the records of a dimension?
                other = project_config.get_base_dimension(DimensionType(dim)).get_records_dataframe().select("id as geography")
                table = other.crossJoin(table.withColumnRenamed("id", dim))
                assert table.count() == other.count() * table.count()
        
        table = table.withColumn("data_source", lit(data_source))
        return table
```

mooneyme: Also just a note. I found that it is an "outer join" when combining association tables provided that have an overlapping dimension column, however it is a "cross join" when we are joining association tables without an overlapping column or dimensions that were not provided with an association table (e.g., geography).

dsgrid/config/dimension_associations.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def __init__(self, table):
+        self._table = table
+        if table is None:
+            self._dimensions = set()
+        else:
+            self._dimensions = set((DimensionType.from_column(x) for x in table.columns))
+
+    @classmethod
+    def load(cls, path: Path, association_files):
```
daniel-thom, line 27: @mooneyme I refactored this class to store a single table containing all associations. Please check the joins again. I left your asserts out. I'm uncertain if they are needed. I typically use asserts to guard against coding errors and not bad inputs. If you were guarding against bad input then they need to be added back.

I did not include a join of the project config records because I think it's better to keep the code modular. You can still create that table very easily in `project_registry_manager.py` in the `_check_dataset_base_to_project_base_mappings` method. The main advantage of this approach is that dimension_associations.py can be tested without a project.

You will also see that I left the code that performs project/dataset checks by iterating over pairs of dimensions. As discussed, it might be possible to make this faster by comparing one overall table with the dataset table (`dim_table` in the method above).

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -113,3 +203,24 @@ def _show_selected_keys_from_dict(dct, keys: list):
...
+                )
+                records = models_to_dataframe(mapping_config.model.records).filter(
+                    "to_id is not NULL"
+                )
+                df = (
+                    df.join(records, df[column] == records.from_id)
+                    .drop("from_id")
+                    .drop(column)
+                    .withColumnRenamed("to_id", column)
+                )
```
lixiliu, line 225: We need to change how dimension columns get remapped to handle duplicates in from_id.

lixiliu (changes requested):

dsgrid/dataset/dataset_schema_handler_base.py
lixiliu: To be addressed in #108

lixiliu:
@daniel-thom - #108 contains some revision intended for this PR. You should review it (to be merged into this PR) first before making changes.

PR #103, opened 2022-01-25
Add infrastructure for handling config updates

daniel-thom:
Completes JIRA Story DSGRID-81

Implements the following behaviors:
- Allows specification of Pydantic model fields as being non-updateable. Fails update commands if this is violated.
- Changes projects' dataset status to unregistered if dimensions, mappings, or associations are changed.


dsgrid/config/project_config.py
```diff
@@ -131,6 +134,7 @@ class InputDatasetModel(DSGBaseModel):
             "The version string must be in semver format (e.g., '1.0.0') and it must be a valid/"
             "existing version in the registry.",
         ),
+        updateable=False,
```
elainethale, line 137: Seems like the version would be updateable?

dsgrid/config/project_config.py
```diff
@@ -139,6 +143,7 @@ class InputDatasetModel(DSGBaseModel):
         default=DatasetRegistryStatus.UNREGISTERED,
         dsg_internal=True,
         notes=("status is "),
+        updateable=False,
```
elainethale, line 146: Does this mean that status is not directly updatable by users?

dsgrid/config/project_config.py
```diff
@@ -149,6 +154,7 @@ class InputDatasetModel(DSGBaseModel):
         title="model_sector",
```
elainethale, line 154: Do the todos above still require addressing? If so, do we have a plan for that? Should we discuss now and decide on next steps?

dsgrid/config/project_config.py
```diff
@@ -197,6 +203,7 @@ class DimensionMappingsModel(DSGBaseModel):
             " mappings are only supplied if a dataset's dimensions do not match the project's"
             " dimension. ",
         ),
+        updateable=False,
```
elainethale, line 206: Seems like you might update the dataset_to_project mappings. Does this preclude that?

dsgrid/registry/config_update_checker_base.py
```diff
@@ -0,0 +1,61 @@
...
+        """Run all checks.
+
+        Raises
+        ------
+        DSGInvalidOperation
+            Raised if the user is changing an immutable field.
+        DSGInvalidRegistryState
+            Raised if a precondition is violated.
+
+        """
```
elainethale, line 59: Should there be a self.check_preconditions() call here?

dsgrid/registry/project_update_checker.py
```diff
@@ -0,0 +1,35 @@
...
+            raise DSGInvalidRegistryState(
+                f"project status must be one of {self._ALLOWED_UPDATE_STATUSES} in order to update"
+            )
+
+    def handle_postconditions(self):
+        if set(self._REQUIRES_DATASET_UNREGISTRATION).intersection(self._changed_fields):
+            for dataset in self._new_model.datasets:
+                if dataset.status == DatasetRegistryStatus.REGISTERED:
+                    dataset.status = DatasetRegistryStatus.UNREGISTERED
+            logger.info("Set all datasets in %s to unregistered", self._new_model.project_id)
```
elainethale, line 35: Should this log message suggest that the user may want to resubmit the datasets?

Should it say why they were unregistered?

tests/test_registry_management.py
```diff
@@ -231,20 +234,6 @@ def register_dataset(dataset_mgr, config_file, dataset_id, user, log_message):
...
-):
-    project_id = project_config.config_id
-    dataset_id = dataset_config.config_id
-    project_mgr.submit_dataset(
-        project_id,
-        dataset_id,
-        dimension_mapping_refs,
-        user,
-        log_message,
-    )
```
elainethale, line 245: Why is this no longer needed?

dsgrid/config/project_config.py
daniel-thom: Correct. Fixed.

dsgrid/config/project_config.py
daniel-thom: Correct. Updated by dsgrid.

dsgrid/config/project_config.py
daniel-thom: I'll defer to @mooneyme

dsgrid/config/project_config.py
daniel-thom: It's an internal field managed by dsgrid. I'm not sure if it's possible, but I may be able to make all `dsg_internal=True` fields non-updateable.

dsgrid/registry/config_update_checker_base.py
daniel-thom: Doh. Thanks.

tests/test_registry_management.py
daniel-thom: It was unused.

dsgrid/config/project_config.py
mooneyme: I think that is the behavior we want. If it ever comes down to it and we need users to be able to update a status, then we will need to make a CLI command for that. But for now this is good.

dsgrid/registry/common.py
```diff
@@ -62,6 +62,7 @@ class ProjectRegistryStatus(DSGEnum):
     INITIAL_REGISTRATION = "Initial Registration"
     IN_PROGRESS = "In Progress"
     COMPLETE = "Complete"
+    PUBLISHED = "Published"
```
daniel-thom, line 65: @mooneyme @elainethale Is this the right term? Meghan used PUBLISHED-INTERNAL or PUBLISHED-OEDI in Jira DSGRID-45.

lixiliu:
Would it help to add a `registration status` col to `dsgrid registry list`?

daniel-thom:
> Would it help to add a `registration status` col to `dsgrid registry list`?

That would be a good addition. New Jira ticket?

dsgrid/registry/common.py
elainethale: I *think* the type of publication will be indicated by the dsgrid registry location? That is, where we are currently developing is effectively internal and when we publish on OEDI we'll need to make a new dsgrid registry there? If that is the correct interpretation, than I think "Published" is sufficient--the type of publication is indicated by the registry location and who can get to it.

lixiliu:
> > Would it help to add a `registration status` col to `dsgrid registry list`?
> 
> That would be a good addition. New Jira ticket?

Added to JIRA 
https://agile.nrel.gov/browse/DSGRID-229#DSGRID-81

mooneyme (approved):
@elainethale ready to merge?

elainethale (approved):
@mooneyme we're all set.

PR #106, opened 2022-01-27
Support growth rate dataset registration

lixiliu:
Completes JIRA Story DSGRID-205
Closes GitHub Issue #

## Description
- Added "growth_rate" as an `InputDatasetType` option for `dataset_type`, used this field to control a new field called `dataset_type_metadata`, where extra data such as `GrowthRateType` can be stored.
- Added NoOpTime Config and Model to represent NOOP time.
- Refined checks for duplicated columns in data
- Added a new check to assert that  all dimension combinations exist.

Per last Friday's meeting (01/28/2022), 
- linkages between datasets to be specified in a derived dataset config. 
- growth rates to choose from existing metric classes, but `units` would be ignored.

Per email exchanges today (02/01/2022) new config field for growth_rate datasets to be handled in a composition format (similar to data_schema) rather than via inheritance (i.e., new GrowthRateDtasetConfig as a child of DatasetCofig)

Test data: https://github.com/dsgrid/dsgrid-test-data/pull/11

## Checklist

- [x] All tests pass


dsgrid/config/dataset_schema_handler_growth_rate.py
```diff
@@ -0,0 +1,92 @@
...
+    def check_consistency(self):
+        path = Path(self._config.model.path)
+        load_data_df = read_dataframe(check_load_data_filename(path), cache=True)
+        load_data_df = self._config.add_trivial_dimensions(load_data_df)
+        with Timer(timer_stats_collector, "check_one_table_data_consistency"):
+            self._check_one_table_data_consistency(self._config, load_data_df)
+        with Timer(timer_stats_collector, "check_metric_units"):
+            self._check_metric_units(self._config)
+
+    def _check_one_table_data_consistency(self, config: DatasetConfig, load_data):
```
daniel-thom, line 33: Is this the exact same code as OneTableDatasetSchemaHandler? If so, is there a way to share the code?

dsgrid/config/dataset_schema_handler_growth_rate.py
```diff
@@ -0,0 +1,92 @@
...
+from dsgrid.utils.spark import read_dataframe, get_unique_values
+from dsgrid.utils.timing import timer_stats_collector, Timer
+from dsgrid.config.dataset_schema_handler_base import DatasetSchemaHandlerBase
+from dsgrid.dimension.base_models import DimensionType
+from dsgrid.exceptions import DSGInvalidDataset, DSGInvalidDimension
+
+logger = logging.getLogger(__name__)
+
+
+class GrowthRateDatasetSchemaHandler(DatasetSchemaHandlerBase):
```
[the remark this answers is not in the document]
daniel-thom: Yes - unless you have another idea. I expect this kind of refactoring as we grow the codebase. If you call the new folder `datasets` then you should probably move `dsgrid/dataset.py` into it.

dsgrid/config/dataset_schema_handler_growth_rate.py
```diff
@@ -0,0 +1,92 @@
...
+                    f"load_data records do not match dimension records for {name}"
+                )
+
+    def _check_metric_units(self, config: DatasetConfig):
+        metric_records = config.get_dimension(DimensionType.METRIC).get_records()
+        metric_units = {x["unit"] for x in metric_records}
+
+        if len(metric_units) > 1:
+            error = True
+        elif list(metric_units)[0] != "":
```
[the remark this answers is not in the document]
daniel-thom: Irrespective of your question, `dimension_config.get_records()` converts the list of data models into a list of dicts. You're better off accessing `dimension_config.model.records` directly or calling `models_to_dataframe` to get a Spark DataFrame. Perhaps we should add a method to do the later into the dimension config class.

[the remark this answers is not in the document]
daniel-thom: Does `none is not an allowed value` from an enum? Which one? Could you add NONE as a value?

dsgrid/config/dimension_config.py
```diff
@@ -60,3 +60,35 @@ def get_unique_ids(self):
 
         """
         return {x.id for x in self.model.records}
+
+    def get_records(self):
```
daniel-thom, line 64: Why would you want this instead of the typed data models or a dataframe?

lixiliu: You are right, the feature is already here: https://github.com/dsgrid/dsgrid/blob/1d4ecde7aadf73da1fae6e49b76c09e22c09e68c/dsgrid/dimension/dimension_records.py

dsgrid/config/dataset_schema_handler_growth_rate.py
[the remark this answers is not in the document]
lixiliu:
```
E   pydantic.error_wrappers.ValidationError: 1 validation error for DimensionsConfigModel
E   dimensions -> 3 -> records -> unit
E     none is not an allowed value (type=type_error.none.not_allowed)
```

dsgrid/config/dataset_config.py
```diff
@@ -103,6 +103,12 @@ class DataSchemaType(DSGEnum):
         Typically appropriate for small, low-temporal resolution datasets.
         """,
     )
+    GROWTH_RATE = EnumValue(
+        value="growth_rate",
+        description="""
+        One_table data schema with load_data table for unitless and temporally arbitrary data, such as growth rates.
+        """,
+    )
```
elainethale, line 111: I was thinking that growth rates would use the same schemas as before. They'd just contain growth rate (fractional) data instead of, e.g., energy or stock data.

dsgrid/config/dataset_config.py
```diff
@@ -189,6 +195,15 @@ class OneTableDataSchemaModel(DSGBaseModel):
...
+    """ data schema model for growth rate format """
+
+    load_data_column_dimension: DimensionType = Field(
+        title="load_data_column_dimension",
+        description="The data dimension for which its values are in column form (pivoted) in the load_data table.",
+    )
+
+
 class DatasetConfigModel(DSGBaseModel):
     """Represents dataset configurations."""
```
elainethale, line 208: I think there should be a new class GrowthRateDatasetConfigModel(DatasetConfigModel), because the data format for growth rates can largely be the same, its just interpretation of the data that is different--to be fully usable we need to be able to say in the config that this growth rate applies to X other dataset and aligns with that dataset's dimensions in Y way. That's additional information that a regular dataset does not need.

dsgrid/config/dimension_config_factory.py
```diff
@@ -10,6 +11,7 @@
     DimensionType,
     AnnualTimeDimensionModel,
     RepresentativePeriodTimeDimensionModel,
+    TrivialTimeDimensionModel,
```
elainethale, line 14: Per my comment on the test data PR, I don't think you should need a new time type--AnnualTime should work (or be made to work) in this case, I think.

elainethale (changes requested):
I don't think this is headed in the right direction. Growth rate datasets shouldn't need a different data schema or a different time type. What sets them apart is that to use them they have to be associated with another dataset to which they apply. It also needs to be clear that while they relate to, e.g., end-uses, they don't have units of, e.g., kWh. (Which in my mind is accomplished by wrapping them in a GrowthRateDataset class rather than a Dataset class.)

I'm generally looking for a GrowthRateDatasetConfig(DatasetConfig) that specifies what type of growth rate is provided and optionally lets you point to the dataset to which it applies. For now it might be sufficient to just define a trivial GrowthRateDatasetConfig(DatasetConfig) maybe with an enum flag saying what kind of growth rate it is. Then we can figure out how to register the relationship with the dataset to which it applies (e.g., ComStock baseline) in a subsequent PR in which we also work out some of the querying issues.

Does that make sense? Generally we probably need to discuss ... put some time on our calendars for Friday @lixiliu?

dsgrid/config/noop_time_dimension_config.py
```diff
@@ -0,0 +1,41 @@
...
+        ranges = [
+            make_time_range(
+                start=None,
+                end=None,
+                frequency=frequency,
+                leap_day_adjustment=None,
+            )
+        ]
+
+        return ranges  # [[]]
```
daniel-thom, line 35: Nit: stale comment?

dsgrid/utils/spark.py
```diff
@@ -129,20 +129,28 @@ def _post_process_dataframe(df, cache=False, require_unique=None):
...
+    column : str or list of str
 
     Returns
     -------
     set
 
     """
-    return {getattr(x, column) for x in df.select(column).distinct().collect()}
+    dfc = df.select(columns).distinct().collect()
+    if type(columns) == list:
```
daniel-thom, line 146:
```
if isinstance(columns, list):
```

dsgrid/utils/spark.py
```diff
@@ -129,20 +129,28 @@ def _post_process_dataframe(df, cache=False, require_unique=None):
...
+        values = set()
+        for row in dfc:
```
daniel-thom, line 148: Technically, it's more performant to make it a set comprehension. Since this could be large and it's easy to do here, I would.

tests/test_aeo_dataset_registration.py
```diff
@@ -86,25 +91,100 @@ def replace_dataset_path(dataset_config_file, dataset_path):
...
-            manager = make_registry_for_aeo(
-                base_dir,
-                make_test_project_dir,
-                dataset,
-                dataset_path=data_dir,
-                include_datasets=False,
+
+        shutil.copyfile(data_dir / "load_data.csv", data_dir / "load_data_original.csv")
+
+        print("1. normal registration: ")
```
daniel-thom, line 107: Do the print statements help debug on errors? Would it be better to use logging debug statements instead? pytest eats stdout by default and it's easy to enable debug on demand.

dsgrid/dataset/dataset_schema_handler_one_table.py
```diff
@@ -79,3 +86,15 @@ def _check_one_table_data_consistency(self, config: DatasetConfig, load_data):
                 raise DSGInvalidDataset(
                     f"load_data records do not match dimension records for {name}"
                 )
+        dim_records_combo = set(itertools.product(*dim_records_list))
+        data_records_combo = get_unique_values(load_data_df, dim_names)
```
daniel-thom, line 90: I'm wondering if it's better to implement this check in Spark instead of Python. Would it be guaranteed to work for TEMPO data if we used the one-table format for it? I'm expecting 53 million unique time arrays for it. We are configuring Spark with a max result size (output of `df.collect()`) of 4 GiB.

If we're safe, I'm fine with it.

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -53,7 +55,8 @@ def _check_dataset_time_consistency(self, config: DatasetConfig, load_data_df):
         """
         time_dim = config.get_dimension(DimensionType.TIME)
         time_dim.check_dataset_time_consistency(load_data_df)
-        self._check_dataset_time_consistency_by_dimensions(time_dim, load_data_df)
+        if time_dim.model.time_type != TimeDimensionType.NOOP:
```
daniel-thom, line 58: One or two PRs ago I thought about implementing a time handler. This one check doesn't seem bad as-is. Will we need one as we expand the code?

dsgrid/dataset/dataset_schema_handler_one_table.py
lixiliu: Here we check all dimension combinations without adding in time. Time is checked separately in `_check_dataset_time_consistency` where we make sure each of these combinations see the same number of timestamps.

dsgrid/dataset/dataset_schema_handler_base.py
lixiliu: I think it's good enough for our purpose.

lixiliu: Later on we will have another time check for local time.

dsgrid/utils/spark.py
lixiliu: I am following the output format from `itertools.product(*dim_records_list)`, which gives a list/set of tuples. Unless I write a custom func to replace itertools.product()?

dsgrid/config/dataset_config.py
```diff
@@ -452,6 +485,11 @@ def add_trivial_dimensions(self, load_data_lookup):
                 load_data_lookup = load_data_lookup.withColumn(col, F.lit(val))
         return load_data_lookup
 
+    def remove_trivial_dimensions(self, load_data_lookup):
```
daniel-thom, line 488: The dataframe passed in is not necessarily a load_data_lookup, right? More generic name?

dsgrid/utils/spark.py
daniel-thom: I just meant like this:
```
values = {tuple(getattr(row, col) for col in columns) for row in dfc}
```

tests/test_aeo_dataset_registration.py
```diff
@@ -86,25 +91,100 @@ def replace_dataset_path(dataset_config_file, dataset_path):
...
+    remove_geography_subsector=None,
+    export_index=False,
+):
+    df_data = pd.read_csv(data_dir / "load_data_original.csv")
+
+    if duplicate_col is not None:
+        df_data[f"{duplicate_col}_dup"] = df_data[duplicate_col]
+        df_data = df_data.rename(columns={f"{duplicate_col}_dup": duplicate_col})
+    if remove_geography_subsector is not None:
+        if type(remove_geography_subsector) != tuple:
```
daniel-thom, line 175: I wouldn't bother with a type check. It will fail in a similar way if someone passes something else in.

dsgrid/dataset/dataset_schema_handler_one_table.py
daniel-thom: How does the math work out? Max number of time arrays * number of dimension columns * size-per-cell

dsgrid/config/dataset_config.py
```diff
@@ -84,6 +84,7 @@ class InputDatasetType(DSGEnum):
     SECTOR_MODEL = "sector_model"
     HISTORICAL = "historical"
     BENCHMARK = "benchmark"
+    GROWTH_RATE = "growth_rate"
```
elainethale, line 87: I think GROWTH_RATE is distinct from InputDatasetType, that is, you could provide a growth_rate dataset based on a sector model, historical data, or benchmarke data.

How about a new enum like:

```
class DatasetQualifiers(DSGEnum):
    GROWTH_RATE = "growth_rate"
```

I imagine we may eventually want to add others. Right now "normalized" comes to mind.

dsgrid/config/dataset_config.py
```diff
@@ -181,14 +183,25 @@ class StandardDataSchemaModel(DSGBaseModel):
...
+    EXPONENTIAL_ANNUAL = "exponential_annual"
+    EXPONENTIAL_MONTHLY = "exponential_monthly"
+
+
+class GrowthRateModel(DSGBaseModel):
+    growth_rate_type: GrowthRateType = Field(
+        title="growth_rate_type",
+        description="Type of growth rates, e.g., exponential_annual",
+        options=GrowthRateType.format_for_docs(),
+    )
```
elainethale, line 202: I like these

dsgrid/config/dataset_config.py
```diff
@@ -209,6 +222,10 @@ class DatasetConfigModel(DSGBaseModel):
     )
     # TODO: This must be validated against the project's dimension records for data_source
     # TODO: This must also be validated against the project_config
+    dataset_type_metadata: Optional[Any] = Field(
+        title="dataset_type_metadata",
+        description="Additional metadata to include related to dataset_type",
+    )
```
elainethale, line 228: So instead of overloading `dataset_type` I'd suggest having:

```
qualifiers: List[DatasetQualifiers]
```

and either

```
qualifier_data: List[Union(GrowthRateModel)]
```

or just

```
growth_rate_data: Optional[GrowthRateModel]
```

In either case, pydantic would need to enforce the presence of GrowthRateModel if qualifiers includes "growth_rate".

dsgrid/dataset/dataset_schema_handler_standard.py
```diff
@@ -73,11 +79,26 @@ def _check_lookup_data_consistency(self, config: DatasetConfig, load_data_lookup
...
+        if dim_records_combo != data_records_combo:
+            missing = dim_records_combo.difference(data_records_combo)
+            logger.error(
+                "load_data_lookup is missing %s dimension combination(s): \n%s",
+                len(missing),
+                missing,
+            )
+            raise DSGInvalidDataset(
+                f"load_data_lookup records do not match dimension records for dimension combinations = {dim_names}"
+            )
```
elainethale, line 96: This code is probably also subject to Dan's questions about whether we should be doing this kind of checking in Spark

dsgrid/utils/spark_partition.py
```diff
@@ -10,21 +10,37 @@ class SparkPartition:
     def __init__(self):
         return
 
-    def get_data_size(self, df, bytes_per_cell=64):
+    def get_data_size(self, df, bytes_per_cell=8):
         """ approximate dataset size """
```
elainethale, line 14: Add a full doc string. I don't know what bytes_per_cell means (and therefore can't follow why the default is being changed from 64 to 8), and right now you have to look at the source code to know what's going to be returned and what the data size units are.

dsgrid/utils/spark_partition.py
```diff
@@ -10,21 +10,37 @@ class SparkPartition:
...
     @timed_info
-    def get_optimal_number_of_files(self, df, MB_per_file=128):
-        """ calculate *optimal* number of files """
+    def get_optimal_number_of_files(self, df, MB_per_cmp_file=128, cmp_ratio=0.18):
+        """calculate *optimal* number of files
+        Parameters
+        ----------
+        df : DataFrame
+        MB_per_cmp_file : float
+            Size of compressed file in MB
```
elainethale, line 27: Change to, "Desired size of compressed file on disk in MB"?

Also, how would one determine cmp_ratio? Would it be possible to make a temp file that serializes a small portion of the data in order to calculate the cmp_ratio?

@daniel-thom

elainethale (changes requested):
I don't think we should overload InputDatasetType with "growth_rate". An alternative is suggested in my detailed comments.

Otherwise looks good!

dsgrid/utils/spark_partition.py
daniel-thom: If the data has floats without repeating values, this is likely pretty good. I can imagine getting a better ratio with 8-byte floats over 4-byte floats, so that could be checked. If the data has large runs of repeated values then the ratio will be huge. IIRC, at least one version of the comstock data exhibits this. If we use this function on the one-table format then this applies to the dimension columns.

All that being said, I think the effort to optimize this probably isn't worth it.

dsgrid/config/dataset_config.py
lixiliu: I don't think this should be a list as we only want one type of data in a given dataset, right? E.g., consumption data would never be stored in the same dataframe as growth rate data

dsgrid/utils/spark_partition.py
lixiliu: We could do that but it's not super necessary because testing compression on a small snippet of the data is still a guessing game and compression ratio is near the ballpark of the default in most cases. The function is meant to be used interactively and iteratively until the actual saved output is of an acceptable size.

dsgrid/dataset/dataset_schema_handler_one_table.py
lixiliu: Per conversation on 02/02/2022, the combo check will not be necessary at the dataset level.

dsgrid/config/dataset_config.py
```diff
@@ -181,14 +182,30 @@ class StandardDataSchemaModel(DSGBaseModel):
...
-    """ data schema model for one table load data format """
-
     load_data_column_dimension: DimensionType = Field(
         title="load_data_column_dimension",
         description="The data dimension for which its values are in column form (pivoted) in the load_data table.",
     )
 
 
+class DatasetQualifierType(DSGEnum):
+    INDEPENDENT = "independent"
```
lixiliu, line 192: `Independent` (default) as in these datasets represent data that would be taken as an "independent variable" in an equation, contracts the fact that growth_rates are parameters in an equation, and derived_datasets would be like dependent variables/output.

Other terms we can consider are: `input` (as in input variable) or `primary_input`. "growth_rate" would be a process-level dataset in a input->process->output paradigm. Up for other recommendations too.

dsgrid/utils/spark.py
```diff
@@ -129,20 +129,26 @@ def _post_process_dataframe(df, cache=False, require_unique=None):
                     raise DSGInvalidField(f"DataFrame has duplicate entries for {column}")
 
 
-def get_unique_values(df, column):
-    """Return the unique values of a dataframe in one column.
+def get_unique_values(df, columns: [str, list]):
```
daniel-thom, line 132: Are you sure this is correct? I think it might be
```
from typing import AnyStr, List, Union
def get_unique_values(df, columns: Union[AnyStr, List]):
```

dsgrid/config/dataset_config.py
elainethale: I'm okay assuming there will be exactly one qualifier for now.

Should the type be Optional[GrowthRateModel] for now?

dsgrid/config/dataset_config.py
```diff
@@ -308,6 +335,18 @@ class DatasetConfigModel(DSGBaseModel):
         ),
     )
 
+    @validator("dataset_qualifier_metadata", pre=True)
+    def check_dataset_qualifier_metadata(cls, metadata, values):
+        if values["dataset_qualifier"] == DatasetQualifierType.INDEPENDENT:
+            pass
```
elainethale, line 341: Explicitly check for empty metadata?

dsgrid/utils/spark_partition.py
```diff
@@ -10,21 +10,55 @@ class SparkPartition:
...
+        Returns
+        -------
+        n_rows : int
+            Number of rows in df
+        n_cols : int
+            Number of columns in df
+        data_MB : float
+            Estimated size of df in memory in MB
+
+        """
```
elainethale, line 32: Thanks!

tests/test_dataset_registration.py
```diff
@@ -57,6 +54,7 @@ def test_invalid_datasets(make_test_project_dir, make_test_data_dir):
             try:
                 # Create a new directory because there are collisions with cached
                 # Spark load_data_lookup dataframes.
+                print(f"> test {i}...")
```
elainethale, line 57: logger.info?

elainethale (approved):
A couple of small things to clean up perhaps, but this looks good overall

lixiliu:
https://github.com/dsgrid/dsgrid/pull/106#discussion_r798832978

"Quantity" would break down when Metric Class = EnergyEfficiency.

If qualifier is an important discriminator, I think it's easy to give it a default value, because it's easier to check that qualifier == some_value than to check that qualifier exists.

dsgrid/config/dataset_config.py
lixiliu: Thanks.

"Quantity" would break down when Metric Class = EnergyEfficiency. And you can still have a growth rate for EnergyEfficiency.

If qualifier is an important discriminator, I think it's easy to give it a default value, because it's easier to check that qualifier == some_value than to check if qualifier exists for a dataset.

PR #108, opened 2022-02-10
Support duplicates in from_id in dimension mappings

lixiliu:
Completes JIRA Story DSGRID-
Addresses Item#1 in GitHub Issue: https://github.com/dsgrid/dsgrid-project-StandardScenarios/issues/33

## Description
Support different types of dimension mapping types and checks (at registration) based on type. Modified apply map functions accordingly. Tested submit_dataset ResStock and ComStock to CONUS_2022.

- Added `mapping_type` enum to dimension_mapping config. Based on the "mapping_type", a DSG internal variable `archetype` is assigned to enable specific checks for duplicate id records and fraction groupie sums. 
- Added `from_fraction` as an optional col (i.e. with default value=1) in dimension mapping records.
- Renamed AssociationTable (Record) Config and Model to MappingTable equivalents.
- Test data includes mapping_type=duplication "model_year_to_model_year" mapping.

Test data PR: https://github.com/dsgrid/dsgrid-test-data/pull/12
SS PR: https://github.com/dsgrid/dsgrid-project-StandardScenarios/pull/34
EFS PR: https://github.com/dsgrid/dsgrid-project-EFS/pull/30

## Checklist

- [x] All tests pass


dsgrid/config/association_tables.py
```diff
@@ -29,6 +29,10 @@ class AssociationTableRecordModel(DSGBaseModel):
         title="to_id",
         description="Destination mapping",
     )
+    fraction: Optional[float] = Field(
+        title="fraction",
+        description="Fraction of from_id to map to to_id",
+    )
```
elainethale, line 35: I don't think users should have to fill a column out if the fraction is always 1.

dsgrid/registry/dimension_mapping_registry_manager.py
```diff
@@ -147,8 +146,52 @@ def validate_records(self, config: DimensionMappingsConfig, warn_only=False):
         Check:
         - from_id and to_id column names
         - check for duplicate IDs and log a warning if they exist (sometimes we want them to exist if it is an aggregation)
+
+        Implemented:
+        - allow duplicates in from_id if fraction col is explicitly filled out, log warning when fraction sums > 1;
+        - duplicates in to_id is always allowed but logged when found.
```
elainethale, line 152: I don't think expected conditions should warn. What happened to adding some meta-data to the DimensionMappingConfig so we know what to expect?

elainethale (changes requested):
We should add meta-data to the dimension mapping config file so we know what to expect and users don't have to fill out a column of all ones.

lixiliu:
@elainethale @mooneyme - PR ready for review

dsgrid/config/association_tables.py
```diff
@@ -31,6 +31,11 @@ class AssociationTableRecordModel(DSGBaseModel):
         title="to_id",
         description="Destination mapping",
     )
```
elainethale, line 33: General question: Is this the data model for dimension mappings, dimension associations, or both?

dsgrid/config/dimension_mapping_base.py
```diff
@@ -66,7 +156,7 @@ class DimensionMappingReferenceModel(DSGBaseModel):
     )
     required_for_validation: Optional[bool] = Field(
         title="version",
-        description="Set to False if a given base-to-base dimension mapping is NOT required for input dataset validation; default is True",
+        description="Set to False if a given base-to-base dimension mapping is NOT required for input dataset validation; default is True",  # outdated, FIXME
```
elainethale, line 159: Can we fix as part of this PR?

dsgrid/registry/dimension_mapping_registry_manager.py
```diff
@@ -144,17 +140,92 @@ def _check_records_against_dimension_records(self, config):
...
+                allow_dup_to_records,
+                "to_id",
+                mapping.filename,
+                mapping.mapping_type.value,
+            )
+
+            if "fraction_sum_eq1" in mapping.archetype.value:
+                self._check_fraction_sum(
+                    mapping.records, mapping.filename, mapping.mapping_type.value
+                )
```
elainethale, line 189: If had SUM_GT1 and SUM_LT1 archetypes could also check for that here when applicable

elainethale (commented):
This looks to be really close--thanks, @lixiliu! Just a few questions/comments from me ...

dsgrid/config/association_tables.py
lixiliu: Looks like it's just for dimension_mappings. I know it's confusing. Association_tables (DimensionAssociations) are just in dict form:
https://github.com/dsgrid/dsgrid/blob/1d4ecde7aadf73da1fae6e49b76c09e22c09e68c/dsgrid/submit/submit_dataset.py#L67-L70

And if I am seeing this correctly, DimensionAssociations are not checked for completeness:
https://github.com/dsgrid/dsgrid/blob/98e95326f83a66d7327962374defbaa2451acf60/dsgrid/config/dimension_associations.py

dsgrid/config/dimension_mapping_base.py
```diff
@@ -6,17 +6,66 @@
...
+    # input from_fraction col, FRACTION_SUM_EQ1
+    MANY_TO_MANY_AGGREGATION = "many_to_many_aggregation"
+    ONE_TO_MANY_DISAGGREGATION = "one_to_many_disaggregation"
+    MANY_TO_MANY_DISAGGREGATION = "many_to_many_disaggregation"
+
+    # input from_fraction col, FRACTION_SUM_NOT_EQ1
+    ONE_TO_ONE_MULTIPLICATION = "one_to_one_multiplication"
+    ONE_TO_MANY_MULTIPLICATION = "one_to_many_multiplication"
+    MANY_TO_ONE_MULTIPLICATION = "many_to_one_multiplication"
+    MANY_TO_MANY_MULTIPLICATION = "many_to_many_multiplication"
```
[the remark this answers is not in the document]
lixiliu: We might need this for scaling up ComStock results so that the floor area matches floor area from AEO. Recall ComStock only models a subset of all comm. building types. We can do this scaling in the form of a dim mapping or we do this directly in the ComStock data so the data submitted to project is already scaled.

dsgrid/config/dimension_mapping_base.py
```diff
@@ -6,17 +6,66 @@
...
+
+    ONE_TO_ONE_MAP_FRACTION_SUM_EQ1 = "one_to_one_map_fraction_sum_eq1"  # unique from and to
+    ONE_TO_MANY_MAP_FRACTION_SUM_EQ1 = "one_to_many_map_fraction_sum_eq1"  # dup from, unique to
+    MANY_TO_ONE_MAP_FRACTION_SUM_EQ1 = "many_to_one_map_fraction_sum_eq1"  # unique from, dup to
+    MANY_TO_MANY_MAP_FRACTION_SUM_EQ1 = "many_to_many_map_fraction_sum_eq1"  # dup from and to
+
+    ONE_TO_ONE_MAP_FRACTION_SUM_NOT_EQ1 = "one_to_one_map_fraction_sum_not_eq1"
+    ONE_TO_MANY_MAP_FRACTION_SUM_NOT_EQ1 = "one_to_many_map_fraction_sum_not_eq1"
+    MANY_TO_ONE_MAP_FRACTION_SUM_NOT_EQ1 = "many_to_one_map_fraction_sum_not_eq1"
+    MANY_TO_MANY_MAP_FRACTION_SUM_NOT_EQ1 = "many_to_many_map_fraction_sum_not_eq1"
```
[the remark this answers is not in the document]
lixiliu: I was back and forth about this. Ultimately I chose NOT_EQ1 to reduce the # of archetypes we have. I can't find a case where we care and need to explicitly check the fraction sum is < or > 1. Right now we only check the fraction sum is 1 for "_EQ1" archetypes. "_NOT_EQ1" archetypes are not checked for fraction sum.

dsgrid/config/dimension_mapping_base.py
lixiliu: What is the new replacement term for base-to-base dimension mapping? Dimension associations?

dsgrid/config/association_tables.py
elainethale: Should we change the data model class name to MappingTableRecordModel (or DimensionMappingTableRecordModel)? @lixiliu @mooneyme

tests/test_registry_management.py
```diff
@@ -202,26 +203,29 @@ def test_invalid_dimension_mapping(make_test_project_dir):
...
-        record_file.write_text(orig_text + "invalid county,CO\n")
+        record_file.write_text(orig_text + "\ninvalid county,1,CO\n")
         with pytest.raises(DSGInvalidDimensionMapping):
             dim_mapping_mgr.register(dimension_mapping_file, user, log_message)
 
         # Invalid 'from' record - nulls aren't allowd
-        record_file.write_text(orig_text + ",CO\n")
-        with pytest.raises(DSGInvalidDimensionMapping):
+        record_file.write_text(orig_text + ",1.2,CO\n")
+        with pytest.raises(ValidationError):
```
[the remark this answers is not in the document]
lixiliu: It's a pydantic catch now since from_id only allows str as input.

dsgrid/config/dimension_mapping_base.py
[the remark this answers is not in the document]
elainethale: This makes sense to me now.

elainethale (commented):
Just one new idea compared to yesterday: EXPLICIT instead of MULTIPLICATION?

lixiliu:
Only the recently registered mapping_id=conus_2022-resstock_energy_enduses_kwh__conus-2022-detailed-end-uses-kwh__5abd7f24-3ed9-4457-a609-2b20d1b63e5f contains the "mapping_type" and "archetype" fields in the registry. Do we need to retroactively add these fields to dim mapping configs registered before this PR? `dsgrid registry list` does not check_records so there would not be an error if we don't modify the configs for now.

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -210,17 +212,52 @@ def _remap_dimension_columns(self, df):
...
-                    df.join(records, df[column] == records.from_id)
+                    df.join(records, df[column] == records.from_id, "cross")
                     .drop("from_id")
                     .drop(column)
                     .withColumnRenamed("to_id", column)
+                ).filter(f"{column} is not NULL")
+                non_fraction_cols = [
+                    x for x in df.columns if x not in {"fraction", "from_fraction"}
+                ]
+                df = df.fillna(1, ["from_fraction"]).selectExpr(
```
daniel-thom, line 256: `subset=`

dsgrid/registry/dimension_mapping_registry_manager.py
```diff
@@ -144,17 +140,103 @@ def _check_records_against_dimension_records(self, config):
...
-        - from_id and to_id column names
-        - check for duplicate IDs and log a warning if they exist (sometimes we want them to exist if it is an aggregation)
+        - duplicate records in from_id and to_id columns per mapping archetype
+        - sum of from_fraction by from_id per mapping archetype
+        - special check for mapping_type=duplication
+
         """
-        pass
+        for mapping in config.model.mappings:
+            if "one_to_one" in mapping.archetype.value:
```
daniel-thom, line 157: This relies on these enum values producing multiple matches, right? Seems a bit non-straightforward. Python enums are very flexible. You could probably add a method that returns allow_dup_from/to_records.

dsgrid/registry/project_registry_manager.py
```diff
@@ -292,34 +294,57 @@ def _check_dataset_base_to_project_base_mappings(
...
+            x.id for x in dataset_config.dimensions if x.type == DimensionType.DATA_SOURCE
+        ][0]
+        data_source = [
+            x.id for x in self._dimension_mgr.get_by_id(data_source_dim_id).model.records
+        ][
+            0
+        ]  # this assumes only only data_source per dataset
         dim_table = handler.get_unique_dimension_rows().drop("id")
+        for col in dim_table.columns:
+            assert dim_table.filter(F.col(col).isNull()).count() == 0
```
daniel-thom, line 307: Should this be an exception instead of assert? Seems like we could use this as a helper function.

tests/test_registry_management.py
```diff
@@ -202,26 +203,29 @@ def test_invalid_dimension_mapping(make_test_project_dir):
         orig_text = record_file.read_text()
 
         # Invalid 'from' record
-        record_file.write_text(orig_text + "invalid county,CO\n")
+        record_file.write_text(orig_text + "\ninvalid county,1,CO\n")
```
daniel-thom, line 206: This kind of problem happens when a file doesn't end with a newline character, which is a convention. I thought I had fixed this in the data. Note that git and GitHub complain about this.

daniel-thom (commented):
Looks great. Thanks for the fixes in addition to the new content.

dsgrid/config/mapping_tables.py
```diff
@@ -31,6 +31,11 @@ class AssociationTableRecordModel(DSGBaseModel):
         title="to_id",
         description="Destination mapping",
     )
+    from_fraction: Union[float, None] = Field(
```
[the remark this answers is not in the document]
daniel-thom: Based on our chat conversation I think this should be `Optional[float]`. I'm torn on the default value. Probably `None`.

tests/test_registry_management.py
lixiliu: I might have ruined your fix in the test data.

dsgrid/config/project_config.py
```diff
@@ -461,7 +470,9 @@ def add_dataset_dimension_mappings(self, dataset_config, references):
     def check_dataset_dimension_mappings(
```
lixiliu, line 470: Is this still needed?

daniel-thom: I say delete it. The other checks that look at all records will catch anything caught here.

elainethale (approved):
Looks good. Dan has some good suggestions. I approve once those are addressed.

dsgrid/config/project_config.py
```diff
@@ -175,7 +175,7 @@ def check_model_sector(cls, model_sector, values):
 
 
 class DimensionMappingsModel(DSGBaseModel):
-    """Defines all dimension mappings associated with a dsgrid project, including base-to-base, base-to-supplemental, and dataset-to-project mappings."""
+    """Defines all dimension mappings associated with a dsgrid project, including dimension associations and base-to-supplemental dimension mappings."""
```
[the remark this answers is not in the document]
lixiliu: I thought the conventional dimension mappings are always dataset-to-project. Right? That's why I removed it.

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -103,8 +103,9 @@ def get_pivot_dimension_columns_mapped_to_project(self):
         """
         columns = set(self.get_pivot_dimension_columns())
         dim_type = self.get_pivot_dimension_type()
+        mapped_pivot_columns = columns
```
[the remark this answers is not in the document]
lixiliu: No, I think it's right, `mapped_pivot_columns = columns` unless a ref mapping remaps them, in which case it becomes: `mapped_pivot_columns = set(mapping.values())`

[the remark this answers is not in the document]
lixiliu: `columns` is called in for-loop remapping process.

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -210,17 +212,52 @@ def _remap_dimension_columns(self, df):
...
+
+    def _map_and_reduce_dimension(self, df, column, records):
+        """ Map and partial reduce a dimension """
+        if column in df.columns:
+            if column == self.get_pivot_dimension_type().value:
+                # map and consolidate pivoted dimension columns
+                nonvalue_cols = list(
+                    set(df.columns).difference(set(self.get_pivot_dimension_columns()))
                 )
+                to_ids = sorted(list({x.to_id for x in records if x.to_id is not None}))
```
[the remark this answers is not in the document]
lixiliu: This is for facilitating sorted pivot columns after remapping the columns and consolidating them in the query.

dsgrid/config/project_config.py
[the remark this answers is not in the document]
daniel-thom: It's not very important, but my point was that this model can be used to map dataset base dimensions to project base dimensions, like countyX to countyY, and that doesn't seem to be covered any more.

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -103,8 +103,9 @@ def get_pivot_dimension_columns_mapped_to_project(self):
         """
         columns = set(self.get_pivot_dimension_columns())
         dim_type = self.get_pivot_dimension_type()
+        mapped_pivot_columns = columns
         for ref in self._mapping_references:
-            if ref.from_dimension_type.value == dim_type:
+            if ref.from_dimension_type == dim_type:
```
daniel-thom, line 108: This doesn't check for multiple matches, which, if it occurred, it would be a bug. We need to identify a place to check for that. Assuming there is only one match, this if clause could end with a `break`.

PR #112, opened 2022-02-28
Support registration of TEMPO dataset

daniel-thom:
This PR allows TEMPO datasets to be registered and submitted to the StandardScenarios project with checks enabled.

There are still some issues related to Spark memory management for which I don't have a reliable solution. The in-memory checks of dimension combinations are extremely problematic. I have been able to run through each piece of code to submit the TEMPO dataset, but not all at once.

I added lots of timer stats in this PR to see how long operations take. I do expect that we will eventually make these debug log messages or at least only log them to the file (no console). For now, I think we need them.

### DIFF
--- a/dsgrid/config/representative_period_time_dimension_config.py
+++ b/dsgrid/config/representative_period_time_dimension_config.py
@@ -1,13 +1,27 @@
 import abc
+import calendar
+import logging
+import sys
 from collections import namedtuple
-from datetime import timedelta
+from datetime import datetime, timedelta
 
+import pandas as pd
 import pyspark.sql.functions as F
 
-from dsgrid.dimension.time import RepresentativePeriodFormat, TimeZone
+from dsgrid.dimension.time import (
+    RepresentativePeriodFormat,
+    TimeZone,
+    DatetimeRange,
+    LeapDayAdjustmentType,
+)
 from dsgrid.exceptions import DSGInvalidDataset
+from dsgrid.time.types import OneWeekPerMonthByHourType
+from dsgrid.utils.timing import Timer, track_timing, timer_stats_collector
 from .dimensions import RepresentativePeriodTimeDimensionModel
 from .time_dimension_base_config import TimeDimensionBaseConfig
+
+
+logger = logging.getLogger(__name__)
 
 
 class RepresentativePeriodTimeDimensionConfig(TimeDimensionBaseConfig):
@@ -28,9 +42,10 @@ class RepresentativePeriodTimeDimensionConfig(TimeDimensionBaseConfig):
     def model_class():
         return RepresentativePeriodTimeDimensionModel
 
+    @track_timing(timer_stats_collector)
     def check_dataset_time_consistency(self, load_data_df):
         self._format_handler.check_dataset_time_consistency(
-            self._format_handler.get_expected_timestamps(self.model.ranges), load_data_df
+            self._format_handler.list_expected_dataset_timestamps(self.model.ranges), load_data_df
         )
 
     def convert_dataframe(self, df):
@@ -41,13 +56,16 @@ class RepresentativePeriodTimeDimensionConfig(TimeDimensionBaseConfig):
         return self._format_handler.get_frequency()
 
     def get_time_ranges(self):
-        return self._format_handler.get_time_ranges(self.model.ranges)
+        return self._format_handler.get_time_ranges(self.model.ranges, self.get_tzinfo())
 
     def get_timestamp_load_data_columns(self):
         return self._format_handler.get_timestamp_load_data_columns()
 
     def get_tzinfo(self):
         return None
+
+    def list_expected_dataset_timestamps(self):
+        return self._format_handler.list_expected_dataset_timestamps(self.model.ranges)
 
 
 class RepresentativeTimeFormatHandlerBase(abc.ABC):
@@ -65,21 +83,6 @@ class RepresentativeTimeFormatHandlerBase(abc.ABC):
         Raises
         ------
         DSGInvalidDataset
-
-        """
-
-    @abc.abstractmethod
-    def get_expected_timestamps(self, ranges):
-        """Return a list of times required to be in the data.
-
-        Parameters
-        ----------
-        ranges : list
-
-        Returns
-        -------
-        list
-            list of datetime
 
         """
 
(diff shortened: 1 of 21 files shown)


dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -268,16 +267,16 @@ def _map_and_reduce_dimension(self, df, column, records):
...
     def _check_null_value_in_unique_dimension_rows(dim_table):
-        dim_with_null = {}
+        dim_with_null = set()
         cols_to_check = {x for x in dim_table.columns if x != "id"}
 
         for col in cols_to_check:
-            null_count = dim_table.filter(F.col(col).isNull()).count()
-            if null_count > 0:
-                dim_with_null[col] = null_count
+            if not dim_table.select(col).filter(f"{col} is NULL").rdd.isEmpty():
```
daniel-thom, line 276: This is a problematic piece of code. `dim_table` for TEMPO has 27 million rows.  Spark frequently fails with memory errors toward the end of this for loop. It appears to be a garbage collection issue. Many online sources suggest enabling the G1GC garbage collector (Java feature). It did fix the issue for me, at least a few times. I feel uncertain about making this the default behavior in Jade. Running this code in Spark local-mode (instead of a standalone cluster) also seems to run better. I'm reluctant to make changes in this area because I don't expect this type of operation to be common. We will normally be processing large on-disk dataframes, not these large in-memory tables.

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -169,27 +168,25 @@ def _check_dataset_time_consistency_by_dimensions(self, time_dim, load_data_df):
             .agg(F.countDistinct(*time_col).alias("distinct_timestamps"))
         )
 
-        distinct_counts = timestamps_by_dims.select("distinct_timestamps").distinct()
-        if distinct_counts.count() > 1:
+        distinct_counts = timestamps_by_dims.select("distinct_timestamps").distinct().collect()
```
lixiliu, line 171: If this is taking too long for TEMPO, we can consider an alternative check. Instead of checking that each dim combo have the same timestamp count and that the count is expected, we can check that each timestamp has the same count (i.e.count df groupy timestamp). Maybe it would be faster considering TEMPO has many dim combos. And we already check in the time class func that timestamps match time range specified.

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -136,30 +135,30 @@ def get_columns_for_unique_arrays(self, time_dim, load_data_df):
...
         Check dataset time consistency such that all dimension combinations return the same set of time ranges.
         """
-        time_ranges = time_dim.get_time_ranges()
-        assert len(time_ranges) == 1, len(time_ranges)
-        time_range = time_ranges[0]
-        # TODO: need to support validation of multiple time ranges: DSGRID-173
-        expected_timestamps = time_range.list_time_range()
-        expected_count = len(expected_timestamps)
+        logger.info("Check dataset time consistency by dimensions.")
+        expected_count = len(time_dim.list_expected_dataset_timestamps())
```
lixiliu, line 161: If this is slow, we can calculate expected_count from the time range info: (start-end)/freq+1

daniel-thom: This one is fast.

dsgrid/dataset/dataset_schema_handler_base.py
lixiliu: Good revision of the code. I think when I wrote this func, it was more for internally checking that the mappings get applied correctly and the df gets consolidated correctly afterward. The consolidation piece of the code ( in `_map_and_reduce_dimension`) is not fully fleshed out yet. I think we can skip it once that's done.

dsgrid/dataset/dataset_schema_handler_standard.py
```diff
@@ -93,8 +87,10 @@ def _check_lookup_data_consistency(self):
                     f"load_data_lookup records do not match dimension records for {name}"
                 )
 
+    @track_timing(timer_stats_collector)
     def _check_dataset_internal_consistency(self):
         """ Check load_data dimensions and id series. """
+        logger.info("Check dataset internal consistency.")
         self._check_load_data_columns()
         data_ids = []
         for row in self._load_data.select("id").distinct().sort("id").collect():
```
lixiliu, line 96: Can we strip id directly without a for-loop?

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -136,30 +135,30 @@ def get_columns_for_unique_arrays(self, time_dim, load_data_df):
...
 
+    @track_timing(timer_stats_collector)
     def _check_dataset_time_consistency(self, load_data_df):
         """Check dataset time consistency such that:
         1. time range(s) match time config record;
         2. all dimension combinations return the same set of time range(s).
 
         """
         if os.environ.get("__DSGRID_SKIP_DATASET_TIME_CONSISTENCY_CHECKS__"):
+            logger.info("Skip dataset time consistency checks.")
```
elainethale, line 146: Should this be a warning?

lixiliu (commented):
See Comments in sections

elainethale (commented):
Looks good to me. I made a minor comment and did not attempt to review the guts. I definitely appreciate that we may need more verbose logging and timing for a while ...

dsgrid/config/representative_period_time_dimension_config.py
```diff
@@ -104,78 +107,106 @@ def get_timestamp_load_data_columns():
...
+            mismatch = sorted(
+                set(expected_timestamps).symmetric_difference(set(actual_timestamps))
+            )
+            raise DSGInvalidDataset(
+                f"load_data timestamps do not match expected times. mismatch={mismatch}"
+            )
+        logger.info("Verified that expected_timestamps equal actual_timestamps")
+
+        lengths = (
+            load_data_df.select("month", "day_of_week", "hour", "id")
```
lixiliu, line 155: Presence of `id` implies it's for Standard SchemaHandler only

lixiliu: Many check instead df.count group by distinct_timestamps is the same for each timestamp. I.e. lengths = df.groupBy("distinct_timestamps").count().

lixiliu (commented):
Comments

dsgrid/dataset/dataset_schema_handler_base.py
daniel-thom: I agree here. The smarter check being performed in the representative_time_config should be moved up here. This code can group by the set of timestamp columns instead of dimension combinations.

lixiliu: FYI, Dimension combinations is just `id` col in the Standard handler b/c load_data is passed into the dataset time check. But yeah, we should just check count by timestamps.
https://github.com/dsgrid/dsgrid/blob/88529273139803da98482d56a51e3f4235cc0b36/dsgrid/dataset/dataset_schema_handler_standard.py#L41

dsgrid/dataset/dataset_schema_handler_base.py
daniel-thom: Now we check for NULLs in both load_data and load_data_lookup.

elainethale (approved):

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -136,75 +135,40 @@ def get_columns_for_unique_arrays(self, time_dim, load_data_df):
...
+        time_cols = time_dim.get_timestamp_load_data_columns()
+        counts = load_data_df.groupBy(*time_cols).count().select("count")
+        distinct_counts = counts.select("count").distinct().collect()
+        if len(distinct_counts) != 1:
+            raise DSGInvalidDataset(
+                f"All time arrays must have the same times: unique timestamp counts = {len(distinct_counts)}"
+            )
+        unique_ta_counts = load_data_df.select(*unique_array_cols).distinct().count()
+        if unique_ta_counts != distinct_counts[0]["count"]:
+            raise DSGInvalidDataset("dataset has invalid counts")
```
lixiliu, line 169: NIT, but more specific msg is better for debugging. sth like this?
f"dataset expects {unique_ta_counts} count for each timestamp in {time_col[0]}: actual={distinct_counts[0]["count"]}"

lixiliu (approved):
NIT

dsgrid/registry/project_registry_manager.py
```diff
@@ -302,20 +305,58 @@ def _check_dataset_base_to_project_base_mappings(
...
+        if diff != 0:
+            msg = (
+                f"Dataset {dataset_config.config_id} is missing required dimension records "
+                f"from project {project_config.config_id}: {diff}"
+            )
+            logger.error(msg)
+            self._find_missing_dimension_records(
+                project_config, dataset_config, dim_table, data_source, exclude_dims
+            )
+            raise DSGInvalidDataset(msg)
```
daniel-thom, line 331: This latest commit reduces the time to check dimension records in half, so it is a no-brainer. I left the old code for reporting purposes, but that can be optimized later. This needs review @elainethale  @lixiliu

lixiliu: Symmetric_difference equivalent for dataframe:
```
dim_table_unique = dim_table.select(*cols).distinct()
project_table.unionAll(dim_table_unique).except(
    project_table.intersect(dim_table_unique)
)
```

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -268,17 +236,33 @@ def _map_and_reduce_dimension(self, df, column, records):
...
+                        dims_with_null.add(col)
+                assert dims_with_null, "Did not find any dimensions with NULL values"
+
+                raise DSGInvalidDimensionMapping(
+                    "Invalid dimension mapping application. "
+                    "Combination of remapped dataset dimensions contain NULL value(s) for "
+                    f"dimension(s): \n{dims_with_null}"
+                )
+        finally:
+            sql("DROP VIEW dim_table")
```
daniel-thom, line 268: @lixiliu This reduced the time to process comstock in half. And it seems to prevent the GC error when running locally.

lixiliu: It's cut in half because the code didn't get past `if not nulls.rdd.isEmpty()` for ComStock right? Though the for-loop should be faster now when prompted since it would be looping through only the null-containing subset.

Does it make sense to drop view right after `nulls` is created and before the for-loop since it's not used again?

daniel-thom: I ran a full test with all three datasets. The time actually dropped from 987s to 373s. The main issue is that starting a single Spark job has high overhead. The for loop triggers that overhead many times. This illustrates the importance of squashing things into one job even when it's harder to write.

The `nulls` dataframe is built on top of `dim_table`, so I'm not sure what would happen I dropped it before the for loop. To me it makes sense visually to see that `dim_table` is used across the code block. I doubt there would be any performance or memory difference.

One possible objection is that this for loop isn't tested. I don't know how to create a scenario where this would be entered. Any ideas?

lixiliu: Don't drop null in _map_and_reduce() might work for ComStock since project only calls for a subset of ComStock geography.

daniel-thom: Yeah, I meant testing it without hacking the code. Testing once is different than testing at every invocation of pytest. The only way we have to do it with current infrastructure is to manipulate the test data. I've seen test infrastructure that allows arbitrarily changing values at runtime, but not with Python. I did perform the one-time test.

lixiliu: Agreed with the test limitations, I don't think we can hack the test data. The null check is meant to catch errors when mapping/data reduction are not applied correctly, so it's checking for internal consistency in the code, not the data.

PR #114, opened 2022-03-21
Fix handling of dataset path

daniel-thom:
Fixes an issue where the dataset path is not correctly handled. We were not correctly handling the case the user passes an absolute path to the local registry when registering a dataset in offline mode.

I also disabled path checking for all cases except when registering the dataset, which should be the only time it is required.


daniel-thom:
This is not the right fix.

PR #115, opened 2022-03-22
Dt/fix dataset path handling

daniel-thom:
Fixes an issue where the dataset path is not correctly handled. We were not correctly handling the case the user passes an absolute path to the local registry when registering a dataset in offline mode.

While trying to fix this issue I came to the conclusion that we made a design mistake by including the path to the load_data.parquet inside the dataset.toml file. This was very tricky to deal with in particular scenarios, especially when handling updates. We don't actually need the path in the file. Instead:
- When the user registers the dataset they can pass the path in the CLI command.
- Once the dataset is registered our code can look up the path. The DatasetConfigModel specifies `dataset_id` and `dataset_version`. This is sufficient to get us to the correct registry path.

### DIFF
--- a/dsgrid/config/dataset_config.py
+++ b/dsgrid/config/dataset_config.py
@@ -401,9 +401,9 @@ class DatasetConfig(ConfigBase):
         return DatasetConfigModel
 
     @classmethod
-    def load_from_registry(cls, config_file, dimension_manager, registry_path):
+    def load_from_registry(cls, config_file, dimension_manager, registry_data_path):
         config = cls._load(config_file)
-        dataset_path = registry_path / config.config_id / config.model.dataset_version
+        dataset_path = registry_data_path / config.config_id / config.model.dataset_version
         return cls.load(config, dimension_manager, dataset_path)
 
     @classmethod
--- a/dsgrid/config/dataset_config.py
+++ b/dsgrid/config/dataset_config.py
@@ -9,6 +9,7 @@ from pydantic import validator, root_validator
 from semver import VersionInfo
 
 from dsgrid.dimension.base_models import DimensionType
+from dsgrid.exceptions import DSGInvalidParameter
 from dsgrid.registry.common import check_config_id_strict
 from .config_base import ConfigBase
 from .dimensions import (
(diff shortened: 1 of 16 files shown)


mooneyme (approved):
looks good to me!

dsgrid/dataset/dataset_schema_handler_one_table.py
```diff
@@ -24,7 +24,7 @@ def __init__(self, load_data_df, *args, **kwargs):
 
     @classmethod
     def load(cls, config: DatasetConfig, *args, **kwargs):
-        path = Path(config.model.path)
+        path = Path(config.dataset_path)
         load_data_df = read_dataframe(check_load_data_filename(path))
```
lixiliu, line 28: Can we just use `load_data_df = read_dataframe(config.load_data_path)` here to be consistent with standard schema_handler?

dsgrid/config/dataset_config.py
```diff
@@ -424,32 +401,50 @@ def model_class():
         return DatasetConfigModel
 
     @classmethod
-    def load(cls, config_file, dimension_manager):
+    def load_from_registry(cls, config_file, dimension_manager, registry_path):
```
lixiliu, line 404: NIT: `registry_data_path` instead of `registry_path`? When I see registry_path, I think of `.dsgrid-registry` but it's really one level down.

dsgrid/registry/dataset_registry_manager.py
```diff
@@ -191,16 +187,18 @@ def update(self, config, update_type, log_message, submitter=None):
...
         version = self._update_config(config, submitter, update_type, log_message)
-        new_key = ConfigKey(config.config_id, version)
+        new_key = ConfigKey(config_id, version)
         self._datasets.pop(old_key, None)
         self._datasets[new_key] = config
 
         if not self.offline_mode:
-            self.sync_push(self.get_registry_directory(config.config_id))
-            self.sync_push(self.get_registry_data_directory(config.config_id))
+            self.sync_push(self.get_registry_directory(config_id))
```
lixiliu, line 200: This confused me because in `def _register()`, `get_registry_directory()` uses `config.model.data_id`, then I realized config.config_id = config.model.data_id for dataset_manager.

lixiliu (approved):
Looks good, minor optional suggestions.

dsgrid/registry/dataset_registry_manager.py
daniel-thom: Yes, the parent class requires that `config_id` be a method so that it can have a consistent way of getting project_id/dataset_id/dimension_id/dimension_mapping_id. I'll make them consistent here.

PR #116, opened 2022-03-22
Show tables in HTML when running in a notebook

daniel-thom:


lixiliu (approved):
Thanks for updating all of the .show() funcs.

PR #117, opened 2022-03-26
Auto-register dependent configs

daniel-thom:
Completes JIRA Story DSGRID-232

This PR is a refactor of the registration process to allow automatic registration of dimensions and dimension mappings without ever having to specify UUIDs in project.toml and dataset.toml (unless you are sharing an already-registered config). That is achieved.

The best way to see a snapshot of the workflow is to look at tests/cli/test_registry.py::test_register_project_and_dataset.

The changes are somewhat substantial. The key aspect is that there is a new class called `RegistrationContext` that gets created anytime someone initiates a registration process. It aggregates all config IDs that get registered along the way. Whenever the process finishes it should sync to AWS on success and delete any intermediate configs on failure.

Tied to test data PR: https://github.com/dsgrid/dsgrid-test-data/pull/15


docs/source/tutorial.rst
```diff
@@ -6,67 +6,55 @@ Tutorial
 Submit a dataset
 ****************
```
elainethale, line 7: Okay to not be super-clear yet, but this doesn't seem to mention that dimensions can now be directly included in the dataset.toml? (You don't necessarily need a dimensions.toml.) Possible to at least mention that so we'll remember to clean up and make more robust documentation later?

dsgrid/tests/make_us_data_registry.py
```diff
@@ -56,9 +56,6 @@ def make_test_data_registry(
     dataset_path = Path(dataset_path)
     path = create_local_test_registry(registry_path)
     dataset_dir = Path("datasets/sector_models/comstock")
-    project_dimension_mapping_config = src_dir / "dimension_mappings.toml"
-    dimension_mapping_config = src_dir / dataset_dir / "dimension_mappings.toml"
-    dimension_mapping_refs = src_dir / dataset_dir / "dimension_mapping_references.toml"
```
elainethale, line 61: While it still exists we should continue to test the old, slow way of registering everything. Is that still happening?

elainethale (approved):
Two questions/comments. Once those are addressed, please merge (I don't need to re-review).

docs/source/tutorial.rst
daniel-thom: I did actually refer to dataset.toml on line 17.

I do agree that this entire page needs enhancement.

dsgrid/tests/make_us_data_registry.py
daniel-thom: There are tests that explicitly call dimension_manager.register_from_file and dimension_mapping_manager.register_from_file, but we do not test that entire workflow. The code paths are shared and so I didn't think it was necessary. But I'm open to it if you think we need a full example.

elainethale: I think that would be best/least regrests to have two parallel examples, one with each workflow.

docs/source/tutorial.rst
elainethale: Maybe add TODOs that describe what kinds of updates need to be made? (Since I think it's higher priority right now just to get this in and get data to ReEDS?)

PR #121, opened 2022-04-07
Improve UI look-and-feel

daniel-thom:
The full workflow is now supported and tested:
- Register project
- Register and submit dataset

It now configures logging and so you can see progress in the notebook. However, it gets cluttered and that needs work in a future PR.

### DIFF
--- a/dev/README.md
+++ b/dev/README.md
@@ -388,6 +388,20 @@ jade submit-jobs config.json -R periodic -r1
 - After all jobs finish `<output-dir>/stats` will interactive resource utilization plots.
 - In the future we will have Spark metrics recorded in JSON files.
 
+## Connecting a Jupyter notebook to a Spark cluster on Eagle
+
+Make sure you have already installed the dsgrid notebooks in your scratch directory. That installs
+notebooks as well as a script that will start a Jupyter notebook on Eagle. It is called `start_notebook.sh`.
+
+1. Create a JADE configuration per the above steps. However, the command passed to JADE must be
+`bash dsgrid-notebooks/start_notebook.sh`.
+2. Consider whether you want to run the notebook server from the container or from your local conda environment.
+Set the JADE config parameter appropriately in `config.json`.
+3. Submit the JADE job. Once the job starts run `tail -f <output-dir>/*.e`. It will eventually show
+the Jupyter notebook URL as well as the the command you need to run to open an SSH tunnel.
+4. Open the SSH tunnel.
+5. Connect to the notebook.
+
 ### Executing scripts
 There are two basic ways to submit scripts to a Spark cluster.
 
--- a/dev/README.md
+++ b/dev/README.md
@@ -242,6 +242,41 @@ $ export SPARK_CLUSTER=spark://<hostname>:7077
 
 Start Jupyter and open `registration.ipynb`.
 
+Handling of stdout and stderr needs improvement. If the notebook gets cluttered, the best solution is to
+re-execute the cell that creates the UI. You can pass default values for all text box fields in order to
+avoid having to re-enter them every time.
+
+```
+app = RegistrationGui(
+    defaults={
+       "local_registry": "/my-local-registry",
+        "project_file": "/repos/dsgrid-project-StandardScenarios/dsgrid_project/project.toml",
+        "dataset_file": "/repos/dsgrid-project-StandardScenarios/dsgrid_project/datasets/sector_models/comstock/dataset.toml",
+        "dimension_mapping_file": "/repos/dsgrid-project-StandardScenarios/dsgrid_project/datasets/sector_models/comstock/dimension_mappings.toml",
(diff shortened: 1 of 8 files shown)


dev/README.md
```diff
@@ -353,6 +388,20 @@ jade submit-jobs config.json -R periodic -r1
...
 
+## Connecting a Jupyter notebook to a Spark cluster on Eagle
+
+Make sure you have already installed the dsgrid notebooks in your scratch directory. That installs
+notebooks as well as a script that will start a Jupyter notebook on Eagle. It is called `start_notebook.sh`.
+
+1. Create a JADE configuration per the above steps. However, the command passed to JADE must be
+`bash dsgrid-notebooks/start_notebook.sh`.
+2. Consider whether you want to run the notebook server from the container or from your local conda environment.
+Set the JADE config parameter appropriately in `config.json`.
```
elainethale, line 399: How do we know what "appropriately" means here? Is there something "appropriately" can link to?

dev/README.md
```diff
@@ -353,6 +388,20 @@ jade submit-jobs config.json -R periodic -r1
...
+3. Submit the JADE job. Once the job starts run `tail -f <output-dir>/*.e`. It will eventually show
+the Jupyter notebook URL as well as the the command you need to run to open an SSH tunnel.
+4. Open the SSH tunnel.
+5. Connect to the notebook.
```
elainethale, line 403: I don't see anything here about multiple nodes ... Should that be an addendum to these basic instructions?

dsgrid/notebooks/start_notebook.sh
```diff
@@ -0,0 +1,11 @@
...
+unset XDG_RUNTIME_DIR
+export SPARK_CLUSTER=$1
+export DSGRID_LOG_FILE_PATH=`pwd`/$2
+echo "Spark cluster is running at ${SPARK_CLUSTER}" >&2
+echo "JADE output directory is ${DSGRID_LOG_FILE_PATH}" >&2
+mkdir -p $DSGRID_LOG_FILE_PATH
+jupyter notebook --no-browser --ip=0.0.0.0 --port 8889 &
+sleep 10
+echo "Create an ssh tunnel with this command: ssh -L 8889:${HOSTNAME}:8889 ${USER}@el1.hpc.nrel.gov" >&2
+wait
```
elainethale, line 11: Drawbacks here compared to https://github.com/dsgrid/dsgrid-coordination/blob/main/prep_spark_cluster_notebook.py (and instructions on that README):
- Not obvious how to start and restart the jupyter notebook if that becomes necessary/desired
- Jade job won't finish elegantly and so won't copy performance stats back to the run folder
- No links to Spark UI (although I still don't know how to use that anyway)

elainethale (commented):
I generally really like start_notebooks.sh and how simple that makes the instructions. I left a few questions/thoughts that might be helpful or just annoying.

dev/README.md
daniel-thom: Step 1 refers to the "above" steps which talk generically about acquiring nodes. I would say that the question of how to choose the number of nodes belongs elsewhere. We need one source to refer to for this question. We are a bit scattered now.

dsgrid/notebooks/start_notebook.sh
daniel-thom: Good point about a clean shutdown. I'll add instructions. The Spark UI discussion is another case where we need one main site to link against.

daniel-thom: Please refer to the README instructions that I added.

tests/cli/test_install_notebooks.py
```diff
@@ -6,13 +6,15 @@
 
 def test_install_notebooks():
     with TemporaryDirectory() as tmpdir:
-        expected_filename = "registration.ipynb"
+        expected_filenames = ["registration.ipynb", "start_notebook.sh"]
         check_run_command(f"dsgrid install-notebooks --path={tmpdir}")
         files = list((Path(tmpdir) / "dsgrid-notebooks").iterdir())
-        assert files
-        assert files[0].name == expected_filename
+        # TODO: Why does this fail only on CI?
```
daniel-thom, line 12: This passes on my local system as well as on Eagle. If you have an idea about what is happening, let me know.

lixiliu:
The GUI doesn't appear to allow interaction to prompt. When there's a lock file in the registry, it would just err out. Can we fix that?

daniel-thom:
> The GUI doesn't appear to allow interaction to prompt. When there's a lock file in the registry, it would just err out. Can we fix that?

When you click a button I'm pretty sure IPython runs the code in a background thread and so input() does not work. We would somehow have to rewire that function or decide how to handle locks globally in the UI. For example, we could catch lock exceptions specifically. Or we could add a widget option that allows overriding locks.

lixiliu (approved):

PR #122, opened 2022-04-12
Add support for dimension display and query names

daniel-thom:


dsgrid/config/project_config.py
```diff
@@ -386,6 +376,26 @@ def get_base_dimension(self, dimension_type: DimensionType):
                 return dim_config
         assert False, dimension_type
 
+    def get_dimension_records(self, dimension_type: DimensionType, query_name: str):
```
daniel-thom, line 379: @lixiliu I'm open to different names. I didn't include `dataframe` for brevity - and because I assume that anyone calling this will want a dataframe. That being said, we could return the list of data models instead.

dsgrid/registry/registry_manager_base.py
```diff
@@ -526,9 +526,6 @@ def remove(self, config_id):
         # TODO: Do we want to handle specific versions? This removes all configs.
 
     def _remove(self, config_id):
-        if not self.offline_mode:
-            # TODO: DSGRID-145
-            raise Exception("sync-push of config removal is currently not supported")
```
daniel-thom, line 531: This was a bug. If you hit an error while registering something in online mode, you would get this. Your local registry would be left in an odd state. In theory, it would get cleaned up in the next sync. However, I think this is the best thing to do.
- `remove` is not available in the CLI
- We don't perform any syncing on error.

dsgrid/config/project_config.py
lixiliu: dataframe is better

dsgrid/config/dimensions.py
```diff
@@ -49,6 +49,21 @@ class DimensionBaseModel(DSGBaseModel):
         ),
         updateable=False,
     )
+    display_name: str = Field(
+        title="display_name",
+        description="Display name. Source for auto-generated query_name.",
+        note="Dimension display names should be singular, concise names that are common across "
+        "projects and datasets. Must be unique within a project or dataset.",
+        notes=(
+            "Only alphanumeric characters and spaces are supported (no dashes or special characters).",
```
lixiliu, line 58: Why spaces here? Aren't underscores better? If used in query, we'll always have to wrap names in "" in sql if with spaces.

dsgrid/config/dimensions.py
```diff
@@ -139,6 +149,22 @@ def check_name(cls, name):
...
+            raise ValueError(f"display_name={display_name} does not meet the requirements")
+        return display_name
+
+    @validator("query_name")
+    def check_query_name(cls, query_name, values):
+        if "display_name" not in values:
+            return query_name
+        if query_name is not None:
+            return query_name
+        return values["display_name"].lower().replace(" ", "_")
```
lixiliu, line 166: I see what you did here, :)

daniel-thom: We do. In this regular expression `\w` allows letters, numbers, and underscores. There is probably some debate about whether underscore is alphanumeric. In Python it is.
```
REGEX_VALID_REGISTRY_DISPLAY_NAME = re.compile(r"^[\w ]+$")
```

dsgrid/config/dimensions.py
lixiliu: " Symbols like *, & and @ are also considered alphanumeric characters." Do we want to support common symbols too?

lixiliu (commented):
comments

lixiliu:
Do we have companion PRs yet to add display_name in SS and EFS? @daniel-thom

dsgrid/config/dimensions.py
daniel-thom: I'm not sure those characters are generally considered alphanumeric. I can just add "and underscores" to remove ambiguity.

daniel-thom:
> Do we have companion PRs yet to add display_name in SS and EFS? @daniel-thom

yes

dsgrid/config/dataset_config.py
```diff
@@ -498,6 +493,10 @@ def load_dimensions(self, dimension_manager):
 
         """
         self._dimensions.update(dimension_manager.load_dimensions(self.model.dimension_references))
+        check_uniqueness((x.model.name for x in self._dimensions.values()), "dimension name")
+        check_uniqueness(
+            (x.model.query_name for x in self._dimensions.values()), "dimension query name"
+        )
```
elainethale, line 499: Can we exclude trivial dimensions from this uniqueness check? When we aggregate to "all sectors" or "all subsectors" I'd like the drop-down for resolution to say "None," and similarly you would never say "where all-sectors = X" or "where all-sectors IN (X)".

dsgrid/config/dimensions.py
elainethale: I am wondering if we should support dashes and convert them to underscores in the query name. Personally, I'd rather not have dashes or underscores in display names, but if we do have them I would prefer dashes to underscores

dsgrid/config/project_config.py
```diff
@@ -460,15 +470,16 @@ def load_dimensions(self, dimension_manager: DimensionRegistryManager):
         supplemental_dimensions = dimension_manager.load_dimensions(
             self.model.dimensions.supplemental_dimension_references
         )
-        dims = list(itertools.chain(base_dimensions.values(), supplemental_dimensions.values()))
+        self._base_dimensions.update(base_dimensions)
+        self._supplemental_dimensions.update(supplemental_dimensions)
+
+        dims = list(self.iter_dimensions())
         check_uniqueness((x.model.name for x in dims), "dimension name")
+        check_uniqueness((x.model.query_name for x in dims), "dimension query name")
```
elainethale, line 478: This will need to change if we are able to ignore uniqueness for trivial dimensions

elainethale: Another way to deal with my issue would be to only force uniqueness within a DimensionType, but I'm not sure if that would work with the way the queries are set up.

tests/data/dimension_models/minimal/dimension_test_time.toml
```diff
@@ -23,6 +24,7 @@ Annual time 2018-2050
 """
 frequency = "P1D" #"P365D"
 name = "annual_time_2018_to_2020"
+display_name = "Timestamp"
 time_interval_type = "instantaneous"
 ranges = [{start = "2018-01-01", end = "2020-01-01"}]
```
elainethale, line 29: Is this meant to cover two years of time or should end be 2019-01-01?

elainethale (commented):
Thanks, @daniel-thom and apologies for the delay. 

I added a couple of comments related to one detail we should iron out in the meeting tomorrow. Otherwise I think this is ready.

dsgrid/config/dimensions.py
```diff
@@ -49,6 +49,29 @@ class DimensionBaseModel(DSGBaseModel):
...
+    display_name: str = Field(
+        title="display_name",
+        description="Display name. Source for auto-generated query_name.",
+        note="Dimension display names should be singular noun phrases that are concise and "
+        "distinguish the dimension across all dimensions within a project, inclusive of dataset "
+        "dimensions, project base dimensions, and project supplemental dimensions. This uniqueness "
+        "requirement applies unless the dimension is trivial, that is, contains only a single "
+        "record. For trivial dimensions, if the dimension represents an aggregation of everything, "
+        "e.g., all of the sectors, then the convention is to list the display name as 'None'; "
+        "if the dimension represents a single slice of the data, e.g., all electricity use or "
```
daniel-thom, line 61: I'm OK with the options listed in the PR comments: only require unique display names within a dimension, skip the check for trivial dimensions, allow dashes.

I think I'm still hung up on one item that may be important for the queries. When I look at the SS project.toml I see that end uses are registered as a dimension. When I look inside the enduses_kw.csv I see that one column is `fuel_id`. One typical value is `electricity`. As a software developer this tells me that if I want to find all `electricity` records I should read this table and filter it. But then I see that there is a supplemental dimension called `all-electric-end-uses`. The mapping record file for that dimension seems to have all of the rows in the previous table with `fuel_id = electricity` (at least I hope it does).

Why do we have what are essentially duplicate mappings? Are we checking to ensure they match? Which one do we use for queries? Should we auto-generate the second table during registration? 

This same idea applies to counties and states. Maybe others, too.

elainethale: I like the idea of auto-generating the second table during registration. That would support (a) the dimension records we have now that are convenient for users that want to concisely communicate mapping information, and (b) the ability to use these sorts of auto-generated mappings in the same way as we use explicitly provided mappings.

daniel-thom: Let's have a discussion on how dimension record fields are to be used generally. It's possible that we can use them to auto-generate mappings, but from chatting with @lixiliu, it doesn't seem likely.

What we definitely could do is add an additional section to `ProjectConfigModel` called something like `all_in_one_supplemental_dimensions`. Instead of a bunch of dimensions and mappings like this:
```
[[dimensions.supplemental_dimensions]]
class = "Sector"
description = """
All Sectors
"""
file = "dimensions/supplemental/all_sectors.csv"
module = "dsgrid.dimension.standard"
name = "all-sectors"
display_name = "All Sector"
type = "sector"
```
```
[[dimension_mappings.base_to_supplemental]]
description = "Aggregation map for Com-Res-Tra sectors"
file = "dimension_mappings/base_to_supplemental/lookup_sectors_to_all_sectors.csv"
mapping_type = "many_to_one_aggregation"
[dimension_mappings.base_to_supplemental.from_dimension]
name = "Sectors - Com-Res-Tra-Only"
type = "sector"
[dimension_mappings.base_to_supplemental.to_dimension]
name = "all-sectors"
type = "sector"
```
we have this one field
```
[dimensions.all_in_one_supplemental_dimensions]
types = ["sector", "subsector", "data_source"]
```
When the registration code sees this entry it auto-generates the dimensions and mappings that are currently generated.

daniel-thom:
This was supplanted by #125

Issue #105, opened 2022-01-26
Add import package cleaning to Black


lixiliu:
NOT urgent, but it would be great if Black can check (and auto-remove if possible) unused imported packages/functions/variables.

daniel-thom:
flake8 reports unused imports. This package will do what you ask, but I'm not sure it's best: https://pypi.org/project/autoflake/. The default behavior does appear to be safe.

elainethale:
Complete

PR #124, opened 2022-04-20
add either pylint or flake8 for format check

lixiliu:
Resolves Issue #105

Completes JIRA Story DSGRID-
Closes GitHub Issue #

## Checklist

- [ ] All tests pass


lixiliu:
When running `pre-commit run --all-files` pylint seems to catch a lot more errors in our codebase than flake8.

elainethale (commented):
Are you proposing to run both flake8 and pylint for now?

Are there any pylint errors we need to ignore?

lixiliu:
> Are you proposing to run both flake8 and pylint for now?
> 
> Are there any pylint errors we need to ignore?

No, we should pick one - a bit of errors to parse through. Here are the logs if you wanted to take a look:

- [pylint_errors.log](https://github.com/dsgrid/dsgrid/files/8525268/pylint_errors.log)
- [flake8_errors.log](https://github.com/dsgrid/dsgrid/files/8525269/flake8_errors.log)

daniel-thom:
> > Are you proposing to run both flake8 and pylint for now?
> > Are there any pylint errors we need to ignore?
> 
> No, we should pick one - a bit of errors to parse through. Here are the logs if you wanted to take a look:
> 
> * [pylint_errors.log](https://github.com/dsgrid/dsgrid/files/8525268/pylint_errors.log)
> * [flake8_errors.log](https://github.com/dsgrid/dsgrid/files/8525269/flake8_errors.log)

1. Was pylint run in the right environment? Those import errors don't make sense to me.
2. It looks like flake8 plays better with click and pydantic. We would have to do add some dynamic disabling of pylint checks.

lixiliu:
> > > Are you proposing to run both flake8 and pylint for now?
> > > Are there any pylint errors we need to ignore?
> > 
> > 
> > No, we should pick one - a bit of errors to parse through. Here are the logs if you wanted to take a look:
> > 
> > * [pylint_errors.log](https://github.com/dsgrid/dsgrid/files/8525268/pylint_errors.log)
> > * [flake8_errors.log](https://github.com/dsgrid/dsgrid/files/8525269/flake8_errors.log)
> 
> 1. Was pylint run in the right environment? Those import errors don't make sense to me.
> 2. It looks like flake8 plays better with click and pydantic. We would have to do add some dynamic disabling of pylint checks.

1. Yes, I suspect pylint has a special way of checking paths that could create false errors (e.g., someone complained here: https://stackoverflow.com/questions/51095449/pylint-false-positive-e0401-import-errors-in-vscode-while-using-venv)
2. Are we leaning more towards flake8? Black and flake8 could work together: https://sbarnea.com/lint/black/

daniel-thom:
> > > > Are you proposing to run both flake8 and pylint for now?
> > > > Are there any pylint errors we need to ignore?
> > > 
> > > 
> > > No, we should pick one - a bit of errors to parse through. Here are the logs if you wanted to take a look:
> > > 
> > > * [pylint_errors.log](https://github.com/dsgrid/dsgrid/files/8525268/pylint_errors.log)
> > > * [flake8_errors.log](https://github.com/dsgrid/dsgrid/files/8525269/flake8_errors.log)
> > 
> > 
> > 
> > 1. Was pylint run in the right environment? Those import errors don't make sense to me.
> > 2. It looks like flake8 plays better with click and pydantic. We would have to do add some dynamic disabling of pylint checks.
> 
> 1. Yes, I suspect pylint has a special way of checking paths that could create false errors (e.g., someone complained here: https://stackoverflow.com/questions/51095449/pylint-false-positive-e0401-import-errors-in-vscode-while-using-venv)
> 2. Are we leaning more towards flake8? Black and flake8 could work together: https://sbarnea.com/lint/black/

My guess is that flake8 will be easier to configure to produce "clean" code but pylint _might_ be more thorough. The examples here show that. Let's start with flake8 and then maybe we'll add both later on.

dsgrid/config/dimension_config.py
```diff
@@ -1,20 +1,8 @@
 import abc
 import logging
-import os
```
lixiliu, line 3: LOL all these unused imports, must have been legacy from before refactoring?

dsgrid/__init__.py
```diff
@@ -1,2 +1,2 @@
-from ._version import __version__, __author__, __copyright__
-from dsgrid.utils.timing import timer_stats_collector
+from dsgrid.utils.timing import timer_stats_collector  # noqa: F401
+from ._version import __version__, __author__, __copyright__  # noqa: F401
```
lixiliu, line 2: why # noqa on these imports?

dsgrid/project.py
```diff
@@ -110,39 +105,9 @@ def unload_dataset(self, dataset_id):
...
-
-    def get_geography(self):
-        # TODO: is name right? cls?
-        return self.get_project_dimension(DimensionType.GEOGRAPHY).name
-
-    def get_dimension_mappings(self, dimension_type, mapping_type):
-        if isinstance(mapping_type, str):
-            mapping_type = MappingType(mapping_type)
-        dimension = self.get_project_dimension(dimension_type)
-        return getattr(dimension, mapping_type.value)
```
lixiliu, line 148: Why are we removing so much code?

dsgrid/tests/common.py
```diff
@@ -24,41 +22,6 @@
...
-    tmpdir = Path(gettempdir()) / "test_data"
-    if os.path.exists(tmpdir):
-        shutil.rmtree(tmpdir)
-    os.mkdir(tmpdir)
-    dst_path = tmpdir / "datasets"
-    shutil.copytree(Path(TEST_DATASET_DIRECTORY), dst_path)
-    yield dst_path
-    shutil.rmtree(tmpdir)
-
-
```
lixiliu, line 61: so much code removal

lixiliu (commented):
Thanks for cleaning it up!

dsgrid/config/dimension_config.py
daniel-thom: It is very difficult to remember to delete these after you remove their use. I blame Python.

dsgrid/__init__.py
daniel-thom: flake8 complains because they are unused in this file. We use them elsewhere, as is `dsgrid.__version__`. So, disable the warning.

dsgrid/project.py
daniel-thom: This is old stuff that we will never use. I didn't want to mess with configuring flake8 to ignore them.

daniel-thom (approved):

dsgrid/project.py
lixiliu: But some of these codes we can use in project query. I definitely used a few of them

daniel-thom: I searched the repo and didn't see any usages. Let's take them one-by-one:
- get_input_dataset: This is actually confusing because it returns a DatasetConfig. There is a get_dataset method that returns a Dataset. I'm improving that one in my query branch. I think that I should go further and either delete or rename list_datasets and _iter_datasets. But let me know how (if) you were using this one.
- get_geography: This is too specific. The ProjectConfig class has `get_base_dimension` which should be used instead.
- get_mapping_type: I think this is stale and completely incorrect. There is no MappingType in our repo.

lixiliu: Yeah you are right, the one I used was get_dataset. I even had a TODO in my branch to delete get_geography for the same reason, and yeah, get_mapping_type/get_dimension_mappings is not working.

lixiliu: I do use get_project_dimension,  and _iter_base_dim is part of that.

lixiliu: Okay then :)

PR #126, opened 2022-05-05
Ll/fix association check

lixiliu:
Completes JIRA Story DSGRID-
Closes GitHub Issue #

## Description
Fix dimension association so that it only checks for and require the minimal set of pair associations.
Works with Standard Scenario branch: https://github.com/dsgrid/dsgrid-project-StandardScenarios/tree/dt/autogen-dims

## Checklist

- [ ] All tests pass

### DIFF
--- a/dsgrid/config/dimension_associations.py
+++ b/dsgrid/config/dimension_associations.py
@@ -4,7 +4,7 @@ from pathlib import Path
 from pyspark.sql.types import StringType
 
 from dsgrid.dimension.base_models import DimensionType
-from dsgrid.utils.spark import read_dataframe
+from dsgrid.utils.spark import read_dataframe, check_for_nulls
 
 
 logger = logging.getLogger(__name__)
@@ -189,4 +189,5 @@ def _join_associations(associations):
             else:
                 table = table.crossJoin(other)
 
+    check_for_nulls(table)
     return table
--- a/dsgrid/config/dimension_associations.py
+++ b/dsgrid/config/dimension_associations.py
@@ -70,8 +70,29 @@ class DimensionAssociations:
                 "Every supplied dimension must have an association with data_source. "
                 f"has_data_source = {dims_with_data_source} all = {all_dims}"
             )
-
         table = _join_associations_by_data_source(associations, associations_by_data_source)
+
+        # # new #
+        # all_dims = set()
+        # associations = {}
+        # for association_file in association_files:
+        #     filename = path / association_file
+        #     table = read_dataframe(filename, cache=True)
+        #     for column in table.columns:
+        #         tmp = column + "tmp_name"
+        #         table = (
+        #             table.withColumn(tmp, table[column].cast(StringType()))
+        #             .drop(column)
+        #             .withColumnRenamed(tmp, column)
+        #         )
+        #     dims = set((DimensionType.from_column(x) for x in table.columns))
+        #     all_dims.update(dims)
+        #     key = _make_key(dims)
+        #     associations[key] = table
+        #     logger.debug("Loaded dimension associations from %s %s", path, table.columns)
+
+        # table = _join_associations(associations)
+
         return cls(table)
 
     @property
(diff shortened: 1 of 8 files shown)


dsgrid/config/dimension_associations.py
```diff
@@ -108,7 +96,7 @@ def get_associations(self, *dimensions, data_source=None):
         ds_column = DimensionType.DATA_SOURCE.value
         columns = [ds_column] + [x.value for x in dimensions]
         table = self._table.select(*columns)
-        if data_source is not None:
+        if data_source is not None and ds_column in table.columns:
```
daniel-thom, line 99: Wouldn't it be an error if data_source is specified but there is no "data_source" in the table?

lixiliu: It would return the entire table down-selected to the specified dimensions, isn't it? If data_source is not specified in any paired association, we assume the table applies to all data_sources, right?

elainethale (commented):
It makes sense to me that data_source can be treated like any other dimension type. data_source being special really just falls out of it being commonly called out in association tables. 

I don't really have any opinion on the logic/implementation details, but it probably would make sense to add a test that would fail with the old code and passes with the new code, if that's possible.

elainethale:
It looks like there is currently a failing test, so maybe that's where this change can be documented.

daniel-thom (approved):

tests/test_project.py
```diff
@@ -64,6 +66,108 @@ def test_dataset_load():
...
+                value_operations.append(operation)
+
+            load_data_sum = load_data_sum.selectExpr(*value_operations).toPandas()
+
+            # 2B.3 check that the newly mapped load_data_sum = mapped_load_data_sum within tolerance
+            decimal_tolerance = 3
+            load_data_diff = (
+                (load_data_sum - mapped_load_data_sum).round(decimal_tolerance).iloc[0]
+            )  # pd.series
+            assert len(load_data_diff[load_data_diff != 0]) == 0
```
daniel-thom, line 165: pandas.testing.assert_series_equal or pandas.testing.assert_frame_equal?

tests/test_project.py
```diff
@@ -46,11 +47,12 @@ def test_dataset_load():
     project = Project.load(PROJECT_ID, offline_mode=True, registry_path=TEST_REGISTRY)
     project.load_dataset(DATASET_ID)
     dataset = project.get_dataset(DATASET_ID)
+
     assert isinstance(dataset, Dataset)
     spark = SparkSession.getActiveSession()
     data = spark.sql(f"select * from {DATASET_ID}__load_data")
     assert "timestamp" in data.columns
-    assert "fans" in data.columns
+    assert "com_fans" in data.columns
```
daniel-thom, line 55: Is there a new test data branch?

tests/test_project.py
```diff
@@ -64,6 +67,108 @@ def test_dataset_load():
...
+                value_operations.append(operation)
+
+            load_data_sum = load_data_sum.selectExpr(*value_operations).toPandas()
+
+            # 2B.3 check that the newly mapped load_data_sum = mapped_load_data_sum within tolerance
+            decimal_tolerance = 3
+            assert pd.testing.assert_frame_equal(
+                load_data_sum.round(decimal_tolerance),
+                mapped_load_data_sum.round(decimal_tolerance),
+            )
```
daniel-thom, line 166: You should be able to do this:
```
assert pd.testing.assert_frame_equal(
                load_data_sum,
                mapped_load_data_sum,
                check_less_precise=True,  # or set atol
            )
```

dsgrid/config/dimension_associations.py
```diff
@@ -215,4 +189,5 @@ def _join_associations_by_data_source(associations, associations_by_data_source)
             else:
                 table = table.crossJoin(other)
 
+    check_for_nulls(table)
```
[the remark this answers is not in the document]
lixiliu: In raw data tables, NULL IDs are okay

Issue #123, opened 2022-04-20
Add dataset syncing to dsgrid registration GUI notebook


lixiliu:
Other issues I am seeing:

- I can still submit-and-register dataset without a log msg.
- status msgs pile on and make it hard to go between the GUI and the latest status

PR #128, opened 2022-05-25
Fix minor issues

daniel-thom:
Fixes these issues:
- DSGRID-244: Make full dimension association table available to users
- One bug listed in #123 

I removed the change to enforce model/weather years as integers because it's not compatible with our design.

### DIFF
--- a/dsgrid/config/project_config.py
+++ b/dsgrid/config/project_config.py
@@ -613,6 +613,48 @@ class ProjectConfig(ConfigWithDataFilesBase):
             if dataset.status == status:
                 yield dataset
 
+    @track_timing(timer_stats_collector)
+    def make_dimension_association_table(self, data_source=None):
+        """Return a table with associations across all dimensions.
+
+        Parameters
+        ----------
+        data_source : DimensionType, optional
+            Filter table by data_source, by default None.
+
+        Returns
+        -------
+        pyspark.sql.DataFrame
+
+        """
+        ds = DimensionType.DATA_SOURCE.value
+        table = self.dimension_associations.table
+        table_columns = set()
+        # table is None when the project doesn't define any dimension associations.
+        if table is not None:
+            if ds in table.columns and data_source is not None:
+                table = table.filter(f"{ds} = '{data_source}'").drop(ds)
+            table_columns.update(table.columns)
+
+        exclude = set((DimensionType.TIME, DimensionType.DATA_SOURCE))
+        all_dimensions = set(d.value for d in DimensionType if d not in exclude)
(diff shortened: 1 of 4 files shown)


elainethale (approved):

dsgrid/config/project_config.py
```diff
@@ -613,6 +613,48 @@ def _iter_datasets_by_status(self, status: DatasetRegistryStatus) -> InputDatase
...
+        data_source : DimensionType, optional
+            Filter table by data_source, by default None.
+
+        Returns
+        -------
+        pyspark.sql.DataFrame
+
+        """
+        ds = DimensionType.DATA_SOURCE.value
+        table = self.dimension_associations.table
```
lixiliu, line 631: I have trouble visualizing this

dsgrid/config/project_config.py
```diff
@@ -613,6 +613,48 @@ def _iter_datasets_by_status(self, status: DatasetRegistryStatus) -> InputDatase
...
+        table_columns = set()
+        # table is None when the project doesn't define any dimension associations.
+        if table is not None:
+            if ds in table.columns and data_source is not None:
+                table = table.filter(f"{ds} = '{data_source}'").drop(ds)
+            table_columns.update(table.columns)
+
+        exclude = set((DimensionType.TIME, DimensionType.DATA_SOURCE))
```
lixiliu, line 639: if data_source is None, we should not exclude `data_source` right?

lixiliu (commented):
See comments. I don't remember where we landed in terms of combining the pairwise association tables.

dsgrid/config/project_config.py
daniel-thom: Yeah, good catch. In the location where I copied this code, data_source was always set.

dsgrid/config/project_config.py
daniel-thom: I think that we decided that since we can't guarantee anything, it's up to the project coordinator to specify the dimension order in the project config.

dsgrid/config/project_config.py
```diff
@@ -636,16 +636,17 @@ def make_dimension_association_table(self, data_source=None):
                 table = table.filter(f"{ds} = '{data_source}'").drop(ds)
             table_columns.update(table.columns)
 
-        exclude = set((DimensionType.TIME, DimensionType.DATA_SOURCE))
+        exclude = {DimensionType.TIME}
+        if data_source is not None:
```
lixiliu, line 640: I rather not drop data_source at all in either cases. Having data_source in the table helps with sanity check.

lixiliu (approved):

PR #131, opened 2022-06-16
Allow access to remote data paths

daniel-thom:
This will allow AWS EC2 nodes to read Parquet files directly from S3. It will also allows HPC users to skip data-syncing.

Completes JIRA Story DSGRID-248

### DIFF
--- a/dsgrid/config/dataset_config.py
+++ b/dsgrid/config/dataset_config.py
@@ -21,9 +21,11 @@ from .dimensions import (
 )
 
 
+# Note that there is special handling for S3 at use sites.
 ALLOWED_LOAD_DATA_FILENAMES = ("load_data.parquet", "load_data.csv")
 ALLOWED_LOAD_DATA_LOOKUP_FILENAMES = (
     "load_data_lookup.parquet",
+    # The next two are only used for test data.
     "load_data_lookup.csv",
     "load_data_lookup.json",
 )
@@ -49,6 +51,7 @@ def check_load_data_filename(path: str):
 
     """
     if path.startswith("s3://"):
+        # Only Parquet is supported on AWS.
         return path + "/load_data.parquet"
 
     for allowed_name in ALLOWED_LOAD_DATA_FILENAMES:
@@ -78,6 +81,7 @@ def check_load_data_lookup_filename(path: str):
 
     """
     if path.startswith("s3://"):
+        # Only Parquet is supported on AWS.
         return path + "/load_data_lookup.parquet"
 
     for allowed_name in ALLOWED_LOAD_DATA_LOOKUP_FILENAMES:
--- a/dsgrid/config/dataset_config.py
+++ b/dsgrid/config/dataset_config.py
@@ -31,16 +31,16 @@ ALLOWED_LOAD_DATA_LOOKUP_FILENAMES = (
 logger = logging.getLogger(__name__)
 
 
-def check_load_data_filename(path: Path):
+def check_load_data_filename(path: str):
     """Return the load_data filename in path. Supports Parquet and CSV.
 
     Parameters
     ----------
-    path : Path
+    path : str
 
     Returns
     -------
-    Path
+    str
 
     Raises
     ------
@@ -48,25 +48,28 @@ def check_load_data_filename(path: Path):
         Raised if no supported load data filename exists.
 
     """
+    if path.startswith("s3://"):
+        return path + "load_data.parquet"
+
     for allowed_name in ALLOWED_LOAD_DATA_FILENAMES:
-        filename = path / allowed_name
+        filename = Path(path) / allowed_name
         if filename.exists():
-            return filename
+            return str(filename)
 
     # Use ValueError because this gets called in Pydantic model validation.
     raise ValueError(f"no load_data file exists in {path}")
 
 
-def check_load_data_lookup_filename(path: Path):
+def check_load_data_lookup_filename(path: str):
     """Return the load_data_lookup filename in path. Supports Parquet, CSV, and JSON.
 
     Parameters
     ----------
-    path : Path
+    path : str
 
     Returns
     -------
-    Path
+    str
 
     Raises
     ------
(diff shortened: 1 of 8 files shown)


dsgrid/config/dataset_config.py
```diff
@@ -31,53 +31,59 @@
...
+    str
 
     Raises
     ------
     ValueError
         Raised if no supported load data lookup filename exists.
 
     """
+    if path.startswith("s3://"):
+        return path + "/load_data_lookup.parquet"
```
elainethale, line 81: Why wouldn't this also use ALLOWED_LOAD_DATA_LOOKUP_FILENAMES? 

I also recall us creating a filesystem interface that had both local and cloud instantiations--should we be using that here instead of introducing if-thens?

dsgrid/config/dimension_associations.py
```diff
@@ -42,7 +42,7 @@ def load(cls, path: Path, association_files):
         associations = {}
         for association_file in association_files:
             filename = path / association_file
-            table = read_dataframe(filename, cache=True)
+            table = read_dataframe(filename, cache=True, read_with_spark=False)
```
elainethale, line 45: Should read_with_spark be an option for the load method? (I.e., is it always always going to be the case that we want read_with_spark to be False here?)

dsgrid/registry/registry_manager.py
```diff
@@ -514,3 +527,33 @@ def get_registry_path(registry_path=None):
...
     return _REGISTRY_TYPE_TO_CLASS[registry_type]
+
+
+def _should_use_remote_data():
+    if sys.platform in ("darwin", "win32"):
+        # Local systems need to sync all load data files.
+        use_remote_data = False
+    elif "NREL_CLUSTER" in os.environ:
+        # All Eagle compute nodes have access to the shared load data files.
+        logger.info("Use remote data on NREL_CLUSTER %s", os.environ["NREL_CLUSTER"])
```
elainethale, line 538: I don't recall this environment variable. Where should it be documented?

dsgrid/utils/spark.py
```diff
@@ -63,6 +63,12 @@ def read_dataframe(filename, cache=False, require_unique=None, read_with_spark=T
     Supported formats when read_with_spark=True: .csv, .json, .parquet
     Supported formats when read_with_spark=False: .csv, .json
 
+    When reading CSV files on AWS read_with_spark should be set to False because the
+    files would need to be present on local storage for all workers. The master node
+    will sync the config files from S3, read them with standard filesystem system calls,
+    and then convert the data to Spark dataframes. This could change if we ever decide
+    to read CSV files with Spark directly from S3.
```
elainethale, line 70: This might be the answer to one of my questions above, but I'll let you confirm @daniel-thom.

dsgrid/utils/spark.py
```diff
@@ -85,36 +91,38 @@ def read_dataframe(filename, cache=False, require_unique=None, read_with_spark=T
...
     _post_process_dataframe(df, cache=cache, require_unique=require_unique)
     return df
 
 
 def _read_with_spark(filename):
     spark = SparkSession.getActiveSession()
-    path = str(filename)
-    if filename.suffix == ".csv":
-        df = spark.read.csv(path, inferSchema=True, header=True)
+    if filename.endswith(".csv"):
```
lixiliu, line 102: Does `filename` here need to be str for spark.read.csv()? If not, use `Path(filename).suffix`?

dsgrid/utils/spark.py
```diff
@@ -85,36 +91,38 @@ def read_dataframe(filename, cache=False, require_unique=None, read_with_spark=T
...
+        # with open(filename, encoding="utf-8-sig") as f_in:
+        #     rows = [Row(**x) for x in csv.DictReader(f_in)]
+        obj = pd.read_csv(filename)
     elif Path(filename).suffix == ".json":
-        rows = load_data(filename)
+        obj = load_data(filename)
     else:
         assert False, f"Unsupported file extension: {filename}"
-    return SparkSession.getActiveSession().createDataFrame(rows)
+    return SparkSession.getActiveSession().createDataFrame(obj)
```
lixiliu, line 125: Is this just `spark.createDataFrame(obj)` since spark is instantiated?

dsgrid/project.py
```diff
@@ -24,19 +25,37 @@ def __init__(self, config, dataset_configs, dimension_mgr, dimension_mapping_mgr
...
+        offline_mode=False,
+        remote_path=REMOTE_REGISTRY,
+        use_remote_data=None,
+    ):
         """Load a project from the registry.
         Parameters
         ----------
         project_id : str
         registry_path : str | None
+        remote_path: str, optional
```
lixiliu, line 42: doc str ordering different than argument ordering

dsgrid/registry/registry_manager.py
lixiliu: Maybe this is supplied by Jade but this is how we determine that the env is on Eagle.

lixiliu (approved):
minor comments

dsgrid/registry/registry_manager.py
lixiliu: In common.py, we used `if os.environ.get("NREL_CLUSTER") is not None:`, is the elif statement equivalent?

dsgrid/config/dimension_associations.py
daniel-thom: I think we always want it to be False. On local systems it really doesn't matter. On EC2 systems it will fail if we try to read with Spark.

dsgrid/registry/registry_manager.py
daniel-thom: This is set by the NREL HPC software. It's documented on their website.

dsgrid/utils/spark.py
daniel-thom: Yes.

dsgrid/registry/registry_manager.py
daniel-thom: > In common.py, we used `if os.environ.get("NREL_CLUSTER") is not None:`, is the elif statement equivalent?

Yes, equivalent

dsgrid/utils/spark.py
daniel-thom: We don't have a `spark` variable in this file or method. We are relying on the global variable in SparkSession.

elainethale (approved):

PR #133, opened 2022-07-09
Add functionality to make a filtered registry for testing

daniel-thom:
The purpose of this PR is to create datasets that are small enough to allow fast tests of queries.

There is a test that demonstrates the functionality in tests/test_filtered_registry.py. I have a more complete example on Eagle at /projects/dsgrid/filtered-registries/simple-standard-scenarios/. That directory has a config file that controls the filtering as well as a filtered version of the StandardScenarios project. The data reduction is ~60 GB -> 6 MB.

### DIFF
--- a/dsgrid/cli/dsgrid_admin.py
+++ b/dsgrid/cli/dsgrid_admin.py
@@ -181,12 +181,20 @@ def _path_cb(_, __, val):
 @click.argument("dst_registry_path", type=click.Path(exists=False), callback=_path_cb)
 @click.argument("config_file", type=click.Path(exists=True), callback=_path_cb)
 @click.option(
+    "--data-symlinks/--no-data-symlinks",
+    default=False,
+    is_flag=True,
+    show_default=True,
+    help="Make symlinks to data files (no big copy operations).",
+)
+@click.option(
     "-r",
     "--use-rsync",
     default=False,
     is_flag=True,
     show_default=True,
-    help="Use rsync instead of copy. Useful when calling many times. Not available on Windows.",
+    help="Use rsync instead of copy. Mutually exclusive with --data-symlinks. Not available on "
+    "Windows.",
 )
 @click.option(
     "-f",
@@ -197,13 +205,23 @@ def _path_cb(_, __, val):
     help="Overwrite dst_registry_path if it already exists. Only applies if use_rsync is False.",
 )
 def make_filtered_registry(
-    src_registry_path: Path, dst_registry_path: Path, config_file: Path, use_rsync, force
+    src_registry_path: Path,
+    dst_registry_path: Path,
+    config_file: Path,
+    data_symlinks,
+    use_rsync,
+    force,
 ):
     """Make a filtered registry for testing purposes."""
+    if use_rsync and data_symlinks:
+        print("use_rsync is not supported with data_symlinks", file=sys.stderr)
+        sys.exit(1)
+
     simple_model = RegistrySimpleModel(**load_data(config_file))
     RegistryManager.copy(
         src_registry_path,
         dst_registry_path,
+        make_data_symlinks=data_symlinks,
         use_rsync=use_rsync,
         force=force,
     )
--- a/dsgrid/cli/dsgrid_admin.py
+++ b/dsgrid/cli/dsgrid_admin.py
@@ -8,8 +8,11 @@ from pathlib import Path
 import click
 
 from dsgrid.common import LOCAL_REGISTRY, REMOTE_REGISTRY
+from dsgrid.config.simple_models import RegistrySimpleModel
 from dsgrid.loggers import setup_logging, check_log_file_size
 from dsgrid.registry.registry_manager import RegistryManager
+from dsgrid.registry.filter_registry_manager import FilterRegistryManager
+from dsgrid.utils.files import load_data
 
 
 logger = logging.getLogger(__name__)
@@ -169,8 +172,48 @@ def remove_dataset(registry_manager, dataset_id):
     registry_manager.dataset_manager.remove(dataset_id)
 
 
+def _path_cb(_, __, val):
+    return Path(val)
+
+
+@click.command()
+@click.argument("src_registry_path", type=click.Path(exists=True), callback=_path_cb)
+@click.argument("dst_registry_path", type=click.Path(exists=False), callback=_path_cb)
+@click.argument("config_file", type=click.Path(exists=True), callback=_path_cb)
+@click.option(
+    "-r",
+    "--use-rsync",
+    default=False,
+    is_flag=True,
+    show_default=True,
+    help="Use rsync instead of copy. Useful when calling many times. Not available on Windows.",
+)
+@click.option(
+    "-f",
+    "--force",
+    default=False,
+    is_flag=True,
+    show_default=True,
+    help="Overwrite dst_registry_path if it already exists. Only applies if use_rsync is False.",
+)
+def make_filtered_registry(
+    src_registry_path: Path, dst_registry_path: Path, config_file: Path, use_rsync, force
+):
+    """Make a filtered registry for testing purposes."""
+    simple_model = RegistrySimpleModel(**load_data(config_file))
+    RegistryManager.copy(
+        src_registry_path,
+        dst_registry_path,
+        use_rsync=use_rsync,
+        force=force,
+    )
+    mgr = FilterRegistryManager.load(dst_registry_path, offline_mode=True)
+    mgr.filter(simple_model=simple_model)
+
+
 cli.add_command(registry)
 cli.add_command(create_registry)
+cli.add_command(make_filtered_registry)
 
 registry.add_command(dimensions)
 registry.add_command(dimension_mappings)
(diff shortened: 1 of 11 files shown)


elainethale (approved):
Looks good to me!

dsgrid/cli/dsgrid_admin.py
```diff
@@ -169,8 +172,66 @@ def remove_dataset(registry_manager, dataset_id):
...
+    help="Use rsync instead of copy. Mutually exclusive with --data-symlinks. Not available on "
+    "Windows.",
+)
+@click.option(
+    "-f",
+    "--force",
+    default=False,
+    is_flag=True,
+    show_default=True,
+    help="Overwrite dst_registry_path if it already exists. Only applies if use_rsync is False.",
```
lixiliu, line 205: NIT, if I am interpreting this correctly, you can either call:
`--data-symlinks` or `--use-rsync`+ optionally`--no-data-symlinks`? And `-f` only works with the former case (i.e. with `--data-symlinks`)? If so, what's the use of having both `--data-symlinks` and `--no-data-symlinks`?

dsgrid/cli/dsgrid_admin.py
```diff
@@ -169,8 +172,66 @@ def remove_dataset(registry_manager, dataset_id):
...
+def make_filtered_registry(
+    src_registry_path: Path,
+    dst_registry_path: Path,
+    config_file: Path,
+    data_symlinks,
+    use_rsync,
+    force,
+):
+    """Make a filtered registry for testing purposes."""
+    if use_rsync and data_symlinks:
```
lixiliu, line 216: What if both False and False?

dsgrid/config/simple_models.py
```diff
@@ -0,0 +1,54 @@
...
+class DimensionsSimpleModel(DSGBaseModel):
+
+    base_dimensions: List[DimensionSimpleModel]
+    supplemental_dimensions: List[DimensionSimpleModel] = Field(default=[])
+
+    @validator("base_dimensions")
+    def check_base_dimensions(cls, base_dimensions):
+        dimension_types = {x.dimension_type for x in base_dimensions}
+        if len(dimension_types) != len(base_dimensions):
+            raise ValueError("base_dimensions cannot contain duplicate dimension types")
```
lixiliu, line 27: Optional, might be good to show what the duplicated dimension(s) is(are).

dsgrid/dataset/dataset_schema_handler_one_table.py
```diff
@@ -96,3 +98,7 @@ def get_unique_dimension_rows(self):
         self._check_null_value_in_unique_dimension_rows(dim_table)
 
         return dim_table
+
+    @track_timing(timer_stats_collector)
+    def filter_data(self, dimensions: List[DatasetSimpleModel]):
+        assert False, "not supported yet"
```
lixiliu, line 104: This would be more or less similar to the two-table. Based on your code:
```
    @track_timing(timer_stats_collector)
     def filter_data(self, dimensions: List[DatasetSimpleModel]):
         load_data = self._load_data
         pivot_dimension_type = self.get_pivot_dimension_type()
         pivoted_columns = set(self.get_pivot_dimension_columns())
         pivoted_columns_to_keep = set()
         load_data_columns = set(load_data.columns)
         for dim in dimensions:
             column = dim.dimension_type.value
             if column in load_data_columns:
                 load_data = load_data.filter(load_data[column].isin(dim.record_ids))
             elif dim.dimension_type == pivot_dimension_type:
                 pivoted_columns_to_keep.update(set(dim.record_ids))
             # else trivial dimension

         # Bring it all into memory so that we can delete the original file.
         load_data2 = load_data.coalesce(1).cache()
         load_data2.count()
         overwrite_dataframe_file(self._config.load_data_lookup_path, lookup2)
         logger.info("Rewrote simplified %s", self._config.load_data_lookup_path)
         ids = next(iter(lookup2.select("id").distinct().select(F.collect_list("id")).first()))

         load_df = self._load_data.filter(self._load_data.id.isin(ids))
         pivoted_columns_to_remove = list(pivoted_columns.difference(pivoted_columns_to_keep))
         load_df = load_df.drop(*pivoted_columns_to_remove)
         overwrite_dataframe_file(self._config.load_data_path, load_df)
         logger.info("Rewrote simplified %s", self._config.load_data_path)
```

dsgrid/registry/filter_registry_manager.py
```diff
@@ -0,0 +1,96 @@
...
+        # Note: Use pandas to write CSVs because Spark produces directories.
+
+        def handle_dimension(simple_dim, dim):
+            records = dim.get_records_dataframe()
+            filename = dim.src_dir / dim.model.filename
+            df = records.filter(records.id.isin(simple_dim.record_ids))
+            df.toPandas().to_csv(filename, index=False)
+            modified_dims.add(dim.model.dimension_id)
+            modified_dim_records[dim.model.dimension_id] = {
+                x.id for x in df.select("id").distinct().collect()
```
lixiliu, line 47: Didn't you come up with a faster way to turn distinct value collection into a list? I can't remember.

dsgrid/registry/registry_manager.py
```diff
@@ -494,6 +496,72 @@ def _update_projects_with_datasets(self, updated_datasets, updated_projects):
...
+        ------
+        DSGInvalidParameter
+            Raised if src is not a valid registry.
+            Raised if dst exists, use_rsync is False, and force is False.
+
+        """
+        if not {x.name for x in src.iterdir()}.issuperset({"configs", "data"}):
+            raise DSGInvalidParameter(f"{src} is not a valid registry")
+
+        if use_rsync:
```
lixiliu, line 527: It looks like, the valid option combos for rsync and symlinks are:
- `use_rsync` (True + Whatever for symlinks)
   (In reality it is True + False only since True + True is restricted in code above)
- `no rsync` + `make_data_symlinks` (False + True)
- `no rsync` + `not make_data_symlinks` (False + False)

lixiliu (approved):
This is reviewed on a very late night, let me know if you find my comments unhelpful.

dsgrid/cli/dsgrid_admin.py
daniel-thom: There should be 3 options:
1. (default) data_symlinks=false use_rsync=false. Copy all data to a new directory. Takes a long time.
2. data_symlinks=true use_rsync=false. Copy all configs to a new directory. Create symlinks to the data files. Very fast.
3. data_symlinks=false use_rsync=true. Run `rsync src dst`. This will skip files that already exist. Slow the first time, then fast. But also kinda pointless if you are filtering data on the copy.

dsgrid/cli/dsgrid_admin.py
daniel-thom: Copy all files with `shutil.copytree`

dsgrid/dataset/dataset_schema_handler_one_table.py
daniel-thom: Yeah, I know it should be straightforward. I didn't want to add code that I couldn't test. Let's add this as a separate effort when the data can be tested.

dsgrid/registry/filter_registry_manager.py
daniel-thom: Yes, and it's a bit awkward. I couldn't remember and it doesn't matter in this case because the records are small.

dsgrid/registry/registry_manager.py
```diff
@@ -506,8 +506,9 @@ def copy(src: Path, dst: Path, make_data_symlinks=False, use_rsync=False, force=
         dst : Path
         simple_model : RegistrySimpleModel
             Filter all configs and data according to this model.
-        make_data_symlinks : bool
-            If True, make symlinks to data files.
+        mode : str
+            Controls whether to copy all data, make symlinks to data files, or sync data with the
+            rsync utility (not available on Windows). Options: 'copy', 'data-symlinks', 'rsync'
         use_rsync : bool
```
lixiliu, line 512: remove?

lixiliu: modify force flag description too

dsgrid/registry/registry_manager.py
```diff
@@ -524,24 +525,26 @@ def copy(src: Path, dst: Path, make_data_symlinks=False, use_rsync=False, force=
...
             logger.info("Copy data from source registry %s", src)
-            if make_data_symlinks:
+            if mode == "data-symlinks":
                 (dst).mkdir()
                 shutil.copytree(src / "configs", dst / "configs")
                 _make_data_symlinks(src, dst)
             else:
                 shutil.copytree(src, dst, symlinks=True)
+        else:
+            raise DSGInvalidParameter(f"mode={mode} is not supported")
```
lixiliu, line 544: optional - could be cleaner if you move this mode validation to `make_filtered_registry`

daniel-thom: click does the validation in the CLI handler. This is independent and needed extra validation.

PR #135, opened 2022-07-12
Fix EMR/bootstrap package version issues

daniel-thom:
The combination of packages combined with conda-forge resulted in Python
3.7 getting installed instead of 3.9. That caused version conflicts with
the Jinja package. Fixed by listing 3.9 explicitly. Also removed unused
packages.


lixiliu (approved):
Looks good

PR #137, opened 2022-07-29
Change query_name rule for projects

daniel-thom:
Require that query_name be unique across all dimensions in a project
instead of per dimension type. This is needed to allow query
functionality.

### DIFF
--- a/dsgrid/config/project_config.py
+++ b/dsgrid/config/project_config.py
@@ -9,8 +9,7 @@ from semver import VersionInfo
 
 from dsgrid.data_models import DSGBaseModel
 from dsgrid.dimension.base_models import check_required_dimensions
-
-from dsgrid.exceptions import DSGInvalidField, DSGInvalidDimension
+from dsgrid.exceptions import DSGInvalidField, DSGInvalidDimension, DSGInvalidParameter
 from dsgrid.registry.common import (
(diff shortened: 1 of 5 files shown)


dsgrid/config/project_config.py
```diff
@@ -436,6 +447,30 @@ def get_base_to_supplemental_dimension_mappings_by_types(self, dimension_type: D
...
+        query_name : str
+
+        Returns
+        -------
+        pyspark.sql.DataFrame
+
+        """
+        dim = self.get_dimension(query_name)
+        dimension_type = dim.model.dimension_type
+        base_dim = self.get_base_dimension(dimension_type)
```
lixiliu, line 464: This is the same as the following right?
```
if query_name in self.get_base_dimension_query_names():
    raise DSGInvalidParameter(f"Cannot pass base dimension: {dimension_type}/{query_name}")
```

lixiliu (approved):
I like it, being able to retrieve config, model, and records directly from query_name is much simpler. 👍

dsgrid/config/project_config.py
daniel-thom: That produces the same result, but what I have is more efficient code-wise. You example has to iterate across each dimension.

Issue #139, opened 2022-08-02
Platform inference error


lixiliu:
Currently we check for the existence of "NREL_CLUSTER" to deduce when a user is on Eagle. However, when the user launches a cluster and ssh directly into the master node, "NREL_CLUSTER" does not get assigned automatically. Instead, try to check for key env vars like SLURM_JOB_ID.

Additionally, when "NREL_CLUSTER" is detected, USE_REMOTE_DATA (or sth similar) is set automatically to TRUE. This is non-sensical when the user is trying to register a dataset and the data do NOT exist remotely yet, and it introduces a dataset path bug where the input path to the datasets (for registration) are overriden with S3 path-equivalent.

Issue #138, opened 2022-08-02
registration copies entire dataset directory


lixiliu:
Current registration process copies the entire parent directory of the dataset. Instead, dsgrid should only copy files with validated filenames, i.e. load_data and load_data_lookup, so irrelevant files that happen to exist in the dataset dir are not copied

PR #140, opened 2022-08-02
Add support for queries of project datasets

daniel-thom:
This PR introduces functionality to do the following:
- Concatenate datasets that have been mapped to project dimensions (except time).
- Filter data according to dimension filter interface.
- Filter data according to user-specified supplemental dimensions.
- Reduce pivot columns according to user-specified supplemental dimensions.
- Perform independent aggregations on query results.
- Cache results - and load from cache on future requests.
- Create composite datasets with a specific query model.
- Query composite datasets.

Testing the code requires the filtered registry in https://github.com/dsgrid/dsgrid-test-data/pull/21. Refer to test code in tests/test_queries.py.

This PR also fixes
- #138 
- #139 
- Minor improvements to the registry-filtering code

Limitations, features excluded from this PR:
- Pre-filter time before processing each dataset.
- Convert dataset time to project time dimension, especially affects TEMPO.
- Filter result dimensions
- Complex filtering of dimensions. No combinations of "and" and "or" within a dimension. No time "between". No combinations of "and" and "or" across dimension types.
- `excluded_dataset_ids`, `include_dsgrid_dataset_components` in ProjectConstraintsModel are not implemented.
- When load_data is joined with load_data_lookup, need to apply the fraction. Lixi will add this from her notebook in a future PR.

This PR contains lots of ideas and contributions from @lixiliu. Thanks Lixi!


tests/test_queries.py
```diff
@@ -0,0 +1,890 @@
...
+                dataset_ids=[
+                    "conus_2022_reference_comstock",
+                    "conus_2022_reference_resstock",
+                    # "tempo_conus_2022",
+                ],
+                # TODO: need to allow min/max/sum of metric on other supplemental dimensions
+                dimension_filters=[
+                    DimensionFilterExpressionModel(
+                        dimension_type=DimensionType.GEOGRAPHY,
+                        query_name="county",
```
mooneyme, line 234: Is this also a `dimension_query_name`? What is the difference between `query_name` here and `dimension_query_name` in the `MetricReductionModel`?

dsgrid/dataset/dimension_filters.py
```diff
@@ -0,0 +1,188 @@
...
+    Example:
+        DimensionFilterExpressionModel(
+            dimension_type=DimensionType.GEOGRAPHY,
+            query_name="county",
+            operator="=="
+            value="06037",
+        ),
+    is equivalent to
+        df.filter("county == '06037'")
+
```
mooneyme, line 99: This example is the same example that you have in `DimensionFilterValueModel`. I am a little confused in the difference between the two, so adding another example (or modifying this one) to distinguish the two classes would be helpful. 

... Also it looks like Expression has the `operator` while Value does not. Is that because Value is an implied `==` operator?

tests/test_queries.py
```diff
@@ -0,0 +1,890 @@
...
+                        operator="==",
+                        value="06037",
+                    ),
+                ],
+                metric_reductions=[
+                    MetricReductionModel(
+                        dimension_query_name="electricity",
+                        operation=self._operation,
+                    ),
+                ],
```
mooneyme, line 244: 1. I am a little confused generally on the `metric_reductions` and the `MetricReductionModel`.  It sort of seems to me like they basically are both filters and aggregations? For example, in this specific query, you are filtering out NG end uses AND aggregating the electricity end uses together into one value.

2. Also, why do we only have metric reductions? What if we wanted to do a subsector reduction? For example, what if we wanted to filter for multi-family building types and sum the electricity values together? Would this be a "reduction" of some form? What would that query look like?

tests/test_queries.py
```diff
@@ -0,0 +1,890 @@
...
+                    DimensionFilterExpressionModel(
+                        dimension_type=DimensionType.GEOGRAPHY,
+                        query_name="county",
+                        operator="==",
+                        value="06037",
+                    ),
+                ],
+                metric_reductions=[
+                    MetricReductionModel(
+                        dimension_query_name="electricity",
```
mooneyme, line 241: Where are we getting this `dimension_query_name` from?

dsgrid/query/models.py
```diff
@@ -0,0 +1,305 @@
...
+            raise ValueError(f"length of ChainedAggregationModel must be at least: {length}")
+        if aggregations[-1].name is None:
+            raise ValueError("ChainedAggregationModel requires its last model to define a name.")
+        return aggregations
+
+
+# TODO: Deserializing JSON into QueryModel probably won't work for all cases, notably the unions.
+
+
+class ProjectConstraintsModel(DSGBaseModel):
```
mooneyme, line 103: This name isn't super intuitive to me.

tests/test_queries.py
```diff
@@ -0,0 +1,890 @@
...
+                include_dsgrid_dataset_components=True,
+                dataset_ids=[
+                    "conus_2022_reference_comstock",
+                    "conus_2022_reference_resstock",
+                    # "tempo_conus_2022",
+                ],
+                dimension_filters=[
+                    DimensionFilterColumnOperatorModel(
+                        dimension_type=DimensionType.METRIC,
+                        query_name="electricity",
```
mooneyme, line 295: Where does this come from? How does this differ from `dimension_query_name` = electricity?

tests/test_queries.py
```diff
@@ -0,0 +1,890 @@
...
+                        operator="like",
+                        value="%",
```
mooneyme, line 297: What is this? Why %? Is this equivelent to `col like "%electricity"`? If so, does the prefix/suffix placement of the % matter here?

tests/test_queries.py
```diff
@@ -0,0 +1,890 @@
...
+                    group_by_columns=["county"],
+                    aggregation_function=F.sum,
+                    name="sum",
+                ),
+                AggregationModel(
+                    group_by_columns=["county"],
+                    aggregation_function=F.max,
+                    name="max",
+                ),
+            ],
```
mooneyme, line 391: Currently, it looks like this just create two output files under the "electricity-use-by-county" query dir -- one for max and one for sum. These do not look chained (which you mentioned as something not being implemented yet). So does this mean that these are independent queries under the same root query name?

tests/test_queries.py
```diff
@@ -0,0 +1,890 @@
...
+                    "conus_2022_reference_resstock",
+                    # "tempo_conus_2022",
+                ],
+                metric_reductions=[
+                    MetricReductionModel(dimension_query_name="electricity", operation="max")
+                ],
+            ),
+            aggregations=[
+                AggregationModel(
+                    group_by_columns=["state"],
```
mooneyme, line 425: Does this use the state county to state dimension mapping? Or does it just use the state field in the county dimension records? Do we have an example of a query that uses a dimension mapping?

tests/test_queries.py
```diff
@@ -0,0 +1,890 @@
...
+    ProjectQuerySubmitter(project, output_dir).submit(
+        query.make_query(),
+        persist_intermediate_table=persist_intermediate_table,
+        load_cached_table=load_cached_table,
+    )
+    result = query.validate()
+    print(f"Result of query {query.name} = {result}")
+
+
+def run_derived_dataset(
```
mooneyme, line 855: What is a derived_dataset in this context? In the queries dir, there is a derived_dataset folder and a cached_tables folder. What is the difference here?

mooneyme (commented):
@daniel-thom I took some time to explore the test_queries.py and figure out what the code is doing/ what files are being generated and what they look like. I still have some more exploration I want to do, especially with the run_derived_dataset, but ran out of time for today. These are my current comments and questions. More to come later.

dsgrid/dataset/dimension_filters.py
daniel-thom: DimensionFilterValueModel is a shortcut for `==`. I can remove it if it's confusing. You can use any operator in DimensionFilterExpressionModel.

dsgrid/query/models.py
daniel-thom: This model defines the constraints that will be used to transform the project's data. This is separate from the parameters that define how to manipulate and store the results. This model uniquely identifies a query and so is used for caching.

I'm open to suggestions on a better name.

tests/test_queries.py
daniel-thom: They are different models that have different fields. Per our meeting I added the `dimension_query_name` alias to `MetricReductionModel` but didn't think to add it to all dimension filter models. I can do that.

tests/test_queries.py
daniel-thom: It represents this supplemental dimension:
```
[[dimensions.supplemental_dimensions]]
class = "EnergyEndUse"
description = """
All Electric End Uses
"""
file = "dimensions/supplemental/all_electric_enduses.csv"
module = "dsgrid.dimension.standard"
name = "all-electric-end-uses"
display_name = "Electricity"
type = "metric"
```

tests/test_queries.py
daniel-thom: 1. Lixi and I debated the term endlessly. Still open to suggestions. Another way of saying it is that MetricReductionModel is used to aggregate (or reduce) the metric values for each set of unique dimension combinations (which is the electricity columns at each row in this case.)

3. You would specify a dimension filter such that only multi-family building types are included. Then you would define an AggregationModel with `group_by_columns=["subsector"]` and `aggregation_function=F.sum`.

tests/test_queries.py
daniel-thom: This filter is one way to saw, "Filter out all columns that are not electricity end uses. Keep the end use column names. Do not perform any reduction/aggregation." The Spark SQL `like` was supposed to match any string - just `"%"`. However, this SQL like statement isn't actually used here. Instead the code filters out any dimension record ID that does not have a mapping. natural_gas_heating does not map to electricity, so it gets filtered out. I need to change this interface.

tests/test_queries.py
daniel-thom: Yes

tests/test_queries.py
daniel-thom: `group_by_columns` is a list of query names. If the query_name does not exist, it will fail. The query_name can be either a base dimension or a supplemental dimension. If it is a supplemental dimension we will look up the dimension mapping and apply it.

tests/test_queries.py
daniel-thom: The contents are the same. The difference is that a derived dataset has a dataset_id and can be directly queried. I think it provides a more explicit workflow. Suppose a user wants to do this:
1. Create a dataset that includes comstock, resstock, and tempo. Don't perform any aggregations. This is a long, slow process and the user only wants to run it once.
2. Explore the data.
3. Perform additional queries to answer specific questions.

The user could achieve this by running a project query with no aggregation, then run additional queries with aggregations. It would be fast because of the caching, but they might not understand why. It also wouldn't be obvious on how to view the cached table.

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -1,17 +1,14 @@
...
-import pyspark.sql.functions as F
-
 from dsgrid.config.dataset_config import DatasetConfig
 from dsgrid.config.simple_models import DatasetSimpleModel
 from dsgrid.dimension.base_models import DimensionType
-from dsgrid.exceptions import DSGInvalidDataset, DSGInvalidDimensionMapping, DSGInvalidField
+from dsgrid.exceptions import DSGInvalidDataset
 from dsgrid.dimension.time import TimeDimensionType
-from dsgrid.utils.spark import models_to_dataframe, check_for_nulls
+from dsgrid.utils.dataset import map_and_reduce_non_pivot_dimension, map_and_reduce_pivot_dimension
```
lixiliu, line 11: How about `pivoted` and `stacked` dimensions?

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -180,91 +189,67 @@ def _check_dataset_time_consistency_by_time_array(self, time_dim, load_data_df):
...
-
-        else:
-            pass
+            records = mapping_config.get_records_dataframe()
+            if column in df.columns:
+                df = map_and_reduce_non_pivot_dimension(df, records, column)
+            elif column == self.get_pivot_dimension_type().value and not pivot_columns.difference(
+                df.columns
+            ):
+                # TODO: Do we want operation to be configurable?
```
lixiliu, line 230: Yes, I would make operation an input para to func

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -180,91 +189,67 @@ def _check_dataset_time_consistency_by_time_array(self, time_dim, load_data_df):
...
-                value_operations.append(operation)
-
-            df = df.selectExpr(*nonvalue_cols, *value_operations)
-
-        else:
-            pass
+            records = mapping_config.get_records_dataframe()
+            if column in df.columns:
+                df = map_and_reduce_non_pivot_dimension(df, records, column)
+            elif column == self.get_pivot_dimension_type().value and not pivot_columns.difference(
```
lixiliu, line 227: I remember what this is checking now. This is for when the mapping is about the pivoted dimension and that the df (since load_data_lookup can be passed in) being passed in contains the pivoted columns.

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -180,91 +189,67 @@ def _check_dataset_time_consistency_by_time_array(self, time_dim, load_data_df):
...
+                # TODO: Do we want operation to be configurable?
+                operation = "sum"
+                df, _ = map_and_reduce_pivot_dimension(
+                    df,
+                    records,
+                    pivot_columns,
+                    operation,
+                    rename=False,
+                )
+            elif column == self.get_pivot_dimension_type().value and pivot_columns.intersection(
```
lixiliu, line 239: This should be `column == self.get_pivot_dimension_type().value and not pivot_columns.intersection(df.columns)`, no exception, just passing.

Because if the mapping is about a pivoted dimension but the df does not have the pivoted dim, then it does nothing.

dsgrid/dataset/dataset_schema_handler_standard.py
```diff
@@ -40,10 +51,123 @@ def get_unique_dimension_rows(self):
...
         dim_table = self._remap_dimension_columns(self._load_data_lookup).distinct()
-        self._check_null_value_in_unique_dimension_rows(dim_table)
-
+        check_null_value_in_unique_dimension_rows(dim_table)
         return dim_table
 
+    def make_project_dataframe(
+        self, load_data_df=None, lookup_df=None, query_context: QueryContext = None
+    ):
+        ld_df = load_data_df or self._load_data
```
lixiliu, line 60: what happens if both are available?

dsgrid/dataset/dataset_schema_handler_standard.py
```diff
@@ -40,10 +51,123 @@ def get_unique_dimension_rows(self):
...
+        ld_df = self._remap_dimension_columns(
+            ld_df,
+            # Some pivot columns may have been removed.
+            pivot_columns=set(ld_df.columns).intersection(self.get_pivot_dimension_columns()),
+        )
+
+        if query_context is not None:
+            ld_df = self._process_metric_aggregations(query_context, ld_df)
+
+        ld_df = ld_df.join(lk_df.drop("fraction"), on="id").drop("id")
```
lixiliu, line 76: Did "fraction" get worked into the load_data table at this point? If not, we cannot drop col.

dsgrid/dataset/dataset_schema_handler_standard.py
```diff
@@ -40,10 +51,123 @@ def get_unique_dimension_rows(self):
...
+            else:
+                dataset_record_ids = (
+                    dataset_mapping.withColumnRenamed("from_id", "dataset_record_id")
+                    .join(
+                        project_record_ids,
+                        on=dataset_mapping.to_id == project_record_ids.id,
+                    )
+                    .select("dataset_record_id")
+                    .distinct()
+                    .withColumnRenamed("dataset_record_id", "id")
```
lixiliu, line 110: Is it necessary to rename it twice? Can you do sth like selectExpr("from_id as id") and remove both withColumnRenamed() in the func?

dsgrid/dataset/dataset_schema_handler_standard.py
```diff
@@ -40,10 +51,123 @@ def get_unique_dimension_rows(self):
...
+        return ld_df.select(*final_columns)
+
+    def _prefilter_dataset(self, context: QueryContext, lk_df, ld_df):
+        # TODO: prefilter time
+        if context.model.project.drop_dimensions:
+            columns = [x.value for x in context.model.project.drop_dimensions]
+            lk_df = lk_df.drop(*columns)
+
+        for dim_type, project_record_ids in context.iter_record_ids_by_dimension_type():
+            dataset_mapping = self._get_dataset_to_project_mapping_records(dim_type)
```
lixiliu, line 98: Does this only return the simple downselection mappings without multipliers (i.e. with default from fractions) ?

dsgrid/dataset/dataset_schema_handler_standard.py
```diff
@@ -40,10 +51,123 @@ def get_unique_dimension_rows(self):
...
+            if dataset_mapping is None:
+                dataset_record_ids = project_record_ids
```
lixiliu, line 100: I get this is a more systematic approach to prefifltering, but would it be more efficient to have a down-selection flag instead of this? I.e., for dimension without mapping, do not apply join and down-selection?

dsgrid/dataset/dataset_schema_handler_standard.py
```diff
@@ -40,10 +51,123 @@ def get_unique_dimension_rows(self):
...
+            if dim_type == self.get_pivot_dimension_type():
+                # Drop columns that don't match requested project record IDs.
+                cols_to_keep = {x.id for x in dataset_record_ids.collect()}
+                cols_to_drop = set(self.get_pivot_dimension_columns()).difference(cols_to_keep)
+                if cols_to_drop:
+                    ld_df = ld_df.drop(*cols_to_drop)
+            else:
+                # Drop rows that don't match requested project record IDs.
+                tmp = dataset_record_ids.withColumnRenamed("id", "dataset_record_id")
+                lk_df = lk_df.join(
```
lixiliu, line 121: Is it possible for ld_df to contain more than just time and pivoted dimension?

lixiliu (commented):
For now

lixiliu (commented):
test 2

dsgrid/dataset/dataset_schema_handler_standard.py
daniel-thom: load_data_df takes precedence. self._load_data is always available.

dsgrid/dataset/dataset_schema_handler_standard.py
daniel-thom: That should have happened in `_remap_dimension_columns` and/or `_process_metric_aggregations`, right? Perhaps it should be dropped as soon as it is applied.

dsgrid/dataset/dataset_schema_handler_standard.py
daniel-thom: There is no mapping here. It should only remove rows or columns that the project is not asking for.

dsgrid/dataset/dataset_schema_handler_standard.py
daniel-thom: Not sure I follow. If there is no mapping provided then the dataset record IDs are the same as the project's. We still may need to remove rows and columns from the dataset.

dsgrid/dataset/dataset_schema_handler_standard.py
daniel-thom: Not to my knowledge. Is that a bad assumption?

dsgrid/dataset/dataset_schema_handler_base.py
daniel-thom: Sure. I'll make the change at the end of the PR. It will probably change lots of files.

dsgrid/dataset/dataset_schema_handler_base.py
daniel-thom: Where would we store the operation?

dsgrid/dataset/dataset_schema_handler_standard.py
lixiliu: I thought if project only wants a subset of what's available, a mapping is always provided. So isn't it that if dataset_records_ids = project_record_ids, no down-selection map-join is required?

dsgrid/dataset/dataset_schema_handler_standard.py
lixiliu: yeah, looks like standard schema handler is hard-coded to include only time, pivoted dim, and id.

dsgrid/dataset/dataset_schema_handler_standard.py
daniel-thom: Project record IDs are filtered before we get to this step in the dataset filter processing. Suppose a user asks for only a single county and the dataset uses the same county IDs. There will be no mapping and this code will still filter the dataset for that one county ID (which is in `project_record_ids`).

dsgrid/utils/dataset.py
```diff
@@ -0,0 +1,117 @@
...
+            val = ",".join(expr)
+            expr = f"least({val}) AS {column}"
+        elif operation in ("avg", "mean"):
+            val = "(" + "+".join(expr) + f") / {len(expr)}"
+            expr = f"{val} AS {column}"
+        else:
+            raise Exception(f"Unsupported operation: {operation}")
+        exprs.append(expr)
+
+    return df.selectExpr(*nonvalue_cols, *exprs), sorted(final_columns)
```
lixiliu, line 99: *exprs is not sorted but final_columns is? Maybe sort both?

dsgrid/dataset/dataset_schema_handler_standard.py
```diff
@@ -40,10 +51,123 @@ def get_unique_dimension_rows(self):
...
+            ld_df, columns = map_and_reduce_pivot_dimension(
+                ld_df,
+                records,
+                pivot_columns,
+                aggregation.operation,
+                rename=True,
+            )
+            if context.metric_columns is None:
+                context.metric_columns = columns
+            elif context.metric_columns != columns:
```
lixiliu, line 149: I don't fully follow the purpose of this check.

dsgrid/query/models.py
lixiliu: ProjectQueryModel?

dsgrid/query/query_submitter.py
```diff
@@ -0,0 +1,367 @@
...
+class ProjectBasedQuerySubmitter(QuerySubmitterBase):
+    def __init__(self, project, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._project = project
+
+    def _process_metric_reductions(self, context, df):
+        assert (
+            len(context.model.metric_reductions) <= 1
+        ), "More than one metric aggregation is not supported yet"
+        cols = set(context.metric_columns)  # TODO DT: same as pivot columns?
```
lixiliu, line 80: Maybe to avoid confusion, use the name value_columns instead of metric_columns? These columns represent some values or quantities but the resolution and column names are dictated by the pivoted dimension.

lixiliu (commented):
For now II.

dsgrid/dataset/dataset_schema_handler_standard.py
lixiliu: In the standard dataset schema, the fraction col often only exists in the lookup table, although I suppose the load_data table could have one too if a mapping with scaling factors is applied to the time column. When you join the two tables in the standard schema, the fraction(s) need to be applied to the value columns and be consolidated that way. Though we can leave fraction as is and apply during the final aggregation step.

dsgrid/utils/dataset.py
daniel-thom: I'm not sure about sorting expressions. The table columns do get properly ordered at the end of StandardDatasetSchemaHandler.make_project_dataframe.

dsgrid/dataset/dataset_schema_handler_standard.py
lixiliu: Got it, I was thinking apples when it was about oranges. Thanks for clarifying

dsgrid/dataset/dataset_schema_handler_standard.py
daniel-thom: We need to store the value of the metric columns. They could be changed during mapping/aggregation/reduction. Every dataset should produce the same metric columns, so assignment only has to happen once. Then every other iteration should be checked to ensure they match.

A case I'm not sure about is one where the user asks for all end uses. Clearly, resstock and tempo will be missing each other's end uses. What happens to the not-applicable columns? Do we fill with null? We've said "no null values for load", but that was really only for individual datasets. Seems like the final dataset can have them.

dsgrid/query/query_submitter.py
daniel-thom: I'm open. If the subsector dimension is pivoted I could be OK with calling those the "metric columns." I could also go with "value_columns;" it just feels a bit generic. Maybe we should stick with "pivoted_columns," as stated in my note.

dsgrid/dataset/dataset_schema_handler_standard.py
lixiliu: If a query asks for both ResStock and Tempo, then yes, we will have to fill some of the columns with null, it's an artifact of having pivoted column. In the case of ResStock only, I think we want to show the null columns. I remember in the very beginning of dsgrid that we wanted this type of behaviors so users know they are indeed zeros.

dsgrid/project.py
```diff
@@ -130,3 +168,115 @@ def _iter_datasets(self):
...
+            dim_type = dim_filter.dimension_type
+            df = record_ids.get(dim_type, default_records(dim_type))
+            if dim_filter.query_name == base_dim_query_names[dim_type]:
+                df = dim_filter.apply_filter(df, column="id")
+            else:
+                mapping_records = self._config.get_base_to_supplemental_mapping_records(
+                    dim_filter.query_name
+                )
+                required_ids = (
+                    mapping_records.filter("to_id is not NULL").select("from_id").distinct()
```
lixiliu, line 220: Doesn't `self._config.get_base_to_supplemental_mapping_records(dim_filter.query_name)` get rid of null to_id from the mapping_records?

dsgrid/project.py
```diff
@@ -130,3 +168,115 @@ def _iter_datasets(self):
...
+            dim_type = dim.model.dimension_type
+            mapping_records = self._config.get_base_to_supplemental_mapping_records(
+                aggregation.query_name
+            ).filter("to_id is not NULL")
+            required_ids = mapping_records.select("from_id").distinct()
+            df = record_ids.get(dim_type, default_records(dim_type))
+            df = df.join(required_ids, on=df.id == required_ids.from_id).drop("from_id")
+            # TODO: We could truncate the mapping records with any record that just got removed
+            # for the life of this query. Save in the context.
+            record_ids[dim_type] = df
```
lixiliu, line 236: Looks like you could refactor here since the steps are similar between `project.dimension_filters` and `project.metric_reductions`.

dsgrid/query/models.py
```diff
@@ -0,0 +1,305 @@
...
+
+
+# class CreateStandaloneDerivedDatasetQueryModel(QueryBaseModel):
+# """Represents a user query to create a Standalone Derived Dataset. This dataset contains
+# all dimension records and dimension mapping records. It does not need a Project.
+# """
+# TODO: create the format (config / .toml) that will define this object
+
+
+class DerivedDatasetQueryModel(QueryResultBaseModel):
```
lixiliu, line 295: Are there equivalencies between these terms? If so, I think we should rename for consistency. Which one is the standalone dataset query interface (i.e. without a dependency on project)?

- ProjectConstraintModel ~ CreateDerivedDatasetQueryModel
- ProjectQueryResultModel ~ DerivedDatasetQueryModel

lixiliu (commented):
Last round of comments, thanks Dan!

dsgrid/dataset/dataset_schema_handler_standard.py
daniel-thom: Should the value be 0 instead of null? I think it's fair to say that the TEMPO dataset has 0 load for electricity_cooling. The columns could be added at various places. I need to think about it.

dsgrid/query/models.py
daniel-thom: ProjectConstraintsModel is used for both project queries (ProjectQueryResultModel) and derived-dataset-creation queries (CreateDerivedDatasetQueryModel). It defines the constraints that are used to filter the data.

The output of ProjectQueryResultModel cannot be queried; it is a final result.
The output of CreateDerivedDatasetQueryModel can be queried - with DerivedDatasetQueryModel. That query produces a final result.

I would like to have a group discussion on these workflows to ensure they are ideal. They work in some cases but probably won't in all. Then we'll have a better idea on naming conventions.

dsgrid/utils/dataset.py
lixiliu: This is just a sort on to_ids before the for-loop if you want to sort both expressions and final_columns

dsgrid/dataset/dataset_schema_handler_base.py
lixiliu: I am going back on what I said. Since this is an operation to map datasets to project, I think the operation should depend on metric type and value_type (I think denoted in the time dimension).

tests/test_queries.py
daniel-thom: This is now fixed. We select only the dimensions that map this supplemental dimension and then apply the filter. Since the filter value is "%", we keep all. If you wanted to keep only columns with "cooling" in the name, `value="%cooling"`. 

This solution may not be immediately intuitive for users and I'm open to a change.

dsgrid/dataset/dataset_schema_handler_standard.py
daniel-thom: I think the double rename is required because we can't collide with "id" in project_record_ids.

tests/test_queries.py
elainethale: Per our meeting, it's non-intuitive that the filter is being applied on the "electricity" query name but the column selections are on the detailed end-uses. Maybe these are really two different operations/filters on two different query names? Maybe default behavior when joining a new query name is to keep the old one around unless it is aggregated out of existence? Is that possible?

tests/test_queries.py
mooneyme: Is this something we want to encourage? Or is this just a placeholder for when we get to chained aggregations?

dsgrid/query/models.py
```diff
@@ -0,0 +1,305 @@
...
+        # ]
+        # ] = Field(
+        title="filters",
+        description="Dimension filters to apply to the dataset'",
+    )
+
+
+class MetricReductionModel(DSGBaseModel):
+    """Reduces the metric values for each set of unique dimension combinations, excluding
+    query_name."""
```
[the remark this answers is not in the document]
daniel-thom: I was trying to express the concept as generically as possible, such that it would apply to a data format with pivoted columns as well as a long, narrow format. Also, with or without time. For each combination of dimension (excluding the one referred to by query_name) there will be one or more metric values that can be reduced with an operation. Does that make sense? If so, how would you write it?:

dsgrid/query/models.py
```diff
@@ -0,0 +1,305 @@
...
+
+    query_name: str = Field(
+        title="query_name", alias="dimension_query_name", description="Dimension query name"
+    )
```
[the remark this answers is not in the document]
daniel-thom: No. It corresponds to the dimension that contains metric values. I was also trying to account for the fact that we may want to support long, narrow format eventually.

dsgrid/query/models.py
```diff
@@ -0,0 +1,305 @@
...
+    excluded_dataset_ids: List[str] = Field(
+        title="excluded_dataset_ids", description="Datasets to exclude from query", default=[]
+    )
+    include_dsgrid_dataset_components: bool = Field(
+        title="include_dsgrid_dataset_components", description=""
+    )
+    metric_reductions: List[MetricReductionModel] = Field(
+        title="metric_reductions",
+        description="Specifies how metric values should be reduced.",
+        default=[],
```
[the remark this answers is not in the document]
daniel-thom: It is an aggregation/reduction of the dimension that contains metric values. No one has liked any term that has been used here.

dsgrid/query/models.py
```diff
@@ -0,0 +1,305 @@
...
+    )
+    drop_dimensions: List[DimensionType] = Field(
+        title="drop_dimensions",
+        description="Drop columns for these dimensions.",
+        default=[],
```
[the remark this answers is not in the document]
daniel-thom: I forget who asked for this. I think the idea was that the user doesn't want to see a particular dimension in the final table because they don't care about it. In that case, there is a bug because this may produce duplicate rows that are not being dropped.

dsgrid/query/models.py
```diff
@@ -0,0 +1,305 @@
...
+    )
+    dimension_filters: List[Any] = Field(
+        # List[
+        # Union[
+        #     DimensionFilterValueModel,
+        #     DimensionFilterExpressionModel,
+        #     DimensionFilterExpressionRawModel,
+        #     DimensionFilterColumnOperatorModel,
+        # ]
```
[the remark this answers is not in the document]
daniel-thom: This has to be fixed for the serialization-deserialization process. Pydantic doesn't work well with unions. I deferred the work.

dsgrid/query/query_context.py
```diff
@@ -0,0 +1,65 @@
...
+
+    @property
+    def model(self):
+        return self._model
+
+    def get_record_ids_by_dimension_type(self, dimension_type):
+        return self._record_ids_by_dimension_type[dimension_type]
+
+    def set_record_ids_by_dimension_type(self, dimension_type, record_ids):
+        self._record_ids_by_dimension_type[dimension_type] = record_ids
```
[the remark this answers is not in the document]
daniel-thom: This is for filtering project records. As of now, the user can only ask for one filter per dimension type. Do we need to support one filter on a base dimension and a second filter on a supplemental dimension?

dsgrid/query/query_submitter.py
daniel-thom: Tied to reductions of the dimension that contains metric values.

dsgrid/query/query_submitter.py
```diff
@@ -0,0 +1,367 @@
...
+    def _query_filename(path: Path):
+        return path / "query.json"
+
+    @staticmethod
+    def _cached_table_filename(path: Path):
+        return path / "table.parquet"
+
+
+class ProjectQuerySubmitter(ProjectBasedQuerySubmitter):
+    """Submits queries for a project."""
```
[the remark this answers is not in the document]
daniel-thom: Yes

tests/test_queries.py
```diff
@@ -0,0 +1,890 @@
...
+        registry_path=REGISTRY_PATH,
+    )
+    try:
+        query = QueryTestElectricityValuesDerivedDataset(
+            REGISTRY_PATH, project, output_dir=output_dir
+        )
+        DerivedDatasetQuerySubmitter(project, output_dir).create_dataset(
+            query.make_query(),
+            persist_intermediate_table=True,
+        )
```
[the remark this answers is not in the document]
daniel-thom: `create_dataset` creates a derived dataset that a user can query. I think that is different in concept than an intermediate table (but the same in content).

tests/test_queries.py
```diff
@@ -0,0 +1,890 @@
...
+        assert query.validate()
```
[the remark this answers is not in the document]
daniel-thom: Yes

tests/test_queries.py
daniel-thom: In a past meeting we agreed that we should support independent aggregations, and so I implemented it. Isn't this something a user would want to do? Like find the max load by county and total load by county?

dsgrid/dataset/dataset_schema_handler_base.py
daniel-thom: We want `pivot_columns.issubset(df.columns)`

dsgrid/dataset/dataset_schema_handler_base.py
daniel-thom: Keep as-is because this will catch a bug if it happens.

dsgrid/dataset/dataset_schema_handler_standard.py
```diff
@@ -40,10 +51,123 @@ def get_unique_dimension_rows(self):
...
         Check each col in combination for null value."""
         dim_table = self._remap_dimension_columns(self._load_data_lookup).distinct()
-        self._check_null_value_in_unique_dimension_rows(dim_table)
-
+        check_null_value_in_unique_dimension_rows(dim_table)
         return dim_table
 
+    def make_project_dataframe(
+        self, load_data_df=None, lookup_df=None, query_context: QueryContext = None
+    ):
```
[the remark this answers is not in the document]
daniel-thom: It's a public method on an internal class.

dsgrid/dataset/dataset_schema_handler_standard.py
daniel-thom: Deferred from this PR.

dsgrid/dataset/dataset_schema_handler_standard.py
daniel-thom: Will insert nulls for now. Make a note that we have to do something with nulls when the user asks for mean.

dsgrid/dataset/dimension_filters.py
daniel-thom: Remove DimensionFilterValueModel

dsgrid/project.py
daniel-thom: Lines 230-2 can be put into a function.

dsgrid/query/models.py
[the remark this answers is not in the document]
daniel-thom: This concept will go away. This PR needs to combine `MetricReductionModel` and `AggregationModel`. The code needs to detect whether a groupBy column requires row-wise or column-wise aggregation and do the right thing.

dsgrid/query/models.py
daniel-thom: `ProjectQueryModel` is the current winner

dsgrid/query/models.py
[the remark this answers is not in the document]
daniel-thom: Will leave this unimplemented for now. Add to TODO list.

dsgrid/query/models.py
daniel-thom: DerivedDatasetQueryModel should include the word `Result`

tests/test_queries.py
```diff
@@ -0,0 +1,890 @@
...
+
+    def get_expected_ny_sum_electricity(self):
+        by_county = self.get_raw_stats()["by_county"]
+        return (
+            by_county["36047"]["comstock_resstock"]["sum"]["electricity"]
+            + by_county["36081"]["comstock_resstock"]["sum"]["electricity"]
+        )
+
+
+class QueryTestElectricityAgg(QueryTestBase):
```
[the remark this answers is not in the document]
daniel-thom: I used this order because I want to see the tests at the top of the file. I run one test at a time while developing and this helps me find them quickly. Let me know if you still disagree.

dsgrid/dataset/dataset.py
```diff
@@ -1,16 +1,18 @@
...
 from dsgrid.config.dataset_schema_handler_factory import make_dataset_schema_handler
+from dsgrid.query.query_context import QueryContext
 
 logger = logging.getLogger(__name__)
 
 
-class Dataset:
-    """Contains metadata and data for a sector."""
+class DatasetBase(abc.ABC):
+    """Base clase for datasets"""
```
lixiliu, line 15: typo

dsgrid/dataset/dataset_schema_handler_standard.py
```diff
@@ -40,10 +54,121 @@ def get_unique_dimension_rows(self):
         """Get distinct combinations of remapped dimensions, including id.
         Check each col in combination for null value."""
         dim_table = self._remap_dimension_columns(self._load_data_lookup).distinct()
-        self._check_null_value_in_unique_dimension_rows(dim_table)
-
+        check_null_value_in_unique_dimension_rows(dim_table)
         return dim_table
 
+    def make_project_dataframe(self):
+        # TODO: Can we remove NULLs at registration time?
```
lixiliu, line 61: Not sure if we can completely remove null values give pivoted dim structure. Imagine a dataset where a county does not have a particular end use. We can always replace remaining null values with 0s. But it might affect mean calculation.

lixiliu: Worth removing where id is null though.

dsgrid/dataset/dataset_schema_handler_standard.py
```diff
@@ -40,10 +54,121 @@ def get_unique_dimension_rows(self):
...
+
+        self._check_aggregations(context)
+        lk_df, ld_df = self._prefilter_dataset(context, lk_df, ld_df)
+
+        lk_df = self._remap_dimension_columns(lk_df)
+        # Some pivoted columns may have been removed in pre-filtering.
+        pivoted_columns = set(ld_df.columns).intersection(self.get_pivoted_dimension_columns())
+        ld_df = self._remap_dimension_columns(ld_df, pivoted_columns=pivoted_columns)
+
+        pivoted_columns = set(ld_df.columns).intersection(
```
lixiliu, line 90: I think you can just use `self.get_pivoted_dimension_columns_mapped_to_project()` here.

dsgrid/dataset/pivoted_table.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def __init__(self, project_config, dataset_id=None):
+        super().__init__(project_config, dataset_id=dataset_id)
+
+    def add_columns(self, df, query_columns, context: QueryContext, aggregation_allowed):
+        columns = set(df.columns)
+        base_query_names = self.project_config.get_base_dimension_query_names()
+        for column in query_columns:
+            query_name = column.dimension_query_name
+            if query_name in base_query_names:
```
lixiliu, line 30: Still need these checks? Feels like you want to see that `query_name` is in both `base_query_names` and `columns`.

dsgrid/dataset/pivoted_table.py
```diff
@@ -0,0 +1,197 @@
...
+                assert query_name in columns, f"{query_name} {columns}"
```
lixiliu, line 31: Still need this check? Feels like you could get away with if `query_name in columns` for both.

dsgrid/dataset/pivoted_table.py
```diff
@@ -0,0 +1,197 @@
...
+                continue
+            elif query_name in columns:
+                continue
+            supp_dim = self._project_config.get_dimension(query_name)
+            base_dim = self._project_config.get_base_dimension(supp_dim.model.dimension_type)
+            base_query_name = base_dim.model.dimension_query_name
+            records = self._project_config.get_base_to_supplemental_mapping_records(query_name)
+            if not aggregation_allowed:
+                to_ids = records.groupBy("from_id").agg(F.count("to_id").alias("count_to_id"))
```
lixiliu, line 40: Seems like you are only allowing for many-to-one and one-to-one mappings. I would check against mapping archetypes (e.g., ONE_TO_ONE_MAP) instead since we already did those checks in registration.

dsgrid/dataset/dataset_schema_handler_standard.py
daniel-thom: In the standard format for input datasets:
- load_data cannot have nulls
- load_data_lookup cannot have nulls for dimension columns (right?). If the data ID column has a null, that row has been intentionally excluded.

dsgrid/dataset/dataset_schema_handler_standard.py
daniel-thom: Agreed. I can assert that there is a full match.

dsgrid/dataset/pivoted_table.py
daniel-thom: Case 1 is a check that the base dimension is already present. If that isn't there, something is very wrong. Case 2 is a redundant addition, which can happen. I would prefer to keep the extra check.

dsgrid/dataset/pivoted_table.py
```diff
@@ -0,0 +1,197 @@
...
+                query_name = column.dimension_query_name
+                # No work is required if the user requested the base dimension for the pivoted
+                # dimension.
+                if (
+                    query_name not in base_query_names
+                    and query_name not in context.get_dimension_query_names(pivoted_dim_type)
+                ):
+                    if column.function is not None:
+                        # TODO: Do we need to support this?
+                        raise Exception("column function cannot be set on {column}")
```
lixiliu, line 113: missing f string
if column.function supported here, it would only act on the column names (imagine the pivoted dim as non-pivoted, the column names become the fields, a column function would operate on those fields. If we are supporting it, we should change name to feature_extraction (i.e., time-related or str-related operations) or sth.

dsgrid/dataset/dataset_schema_handler_standard.py
lixiliu: right, the question is whether we should drop the id=null rows from load_data_lookup even though we painstakingly put them there previously

dsgrid/dataset/dataset_schema_handler_standard.py
lixiliu: I think `self.get_pivoted_dimension_columns_mapped_to_project()` already does the intersection stuff

dsgrid/dataset/pivoted_table.py
```diff
@@ -0,0 +1,197 @@
...
+        new_pivoted_columns = []  # Will be the pivoted columns after the last aggregation.
+        pivoted_dim_type = context.get_pivoted_dimension_type(dataset_id=self.dataset_id)
+        for agg in aggregations:
+            base_query_names = self.project_config.get_base_dimension_query_names()
+            dimension_query_name = None
+            for _, column in agg.iter_dimensions_to_keep():
+                query_name = column.dimension_query_name
+                # No work is required if the user requested the base dimension for the pivoted
+                # dimension.
+                if (
```
lixiliu, line 107: Is this cond checking if query_name is supplemental?

lixiliu: I feel like we can refactor these checks to return whether a query_name is base vs. supplemental, and stacked vs. pivoted.

dsgrid/dataset/pivoted_table.py
```diff
@@ -0,0 +1,197 @@
...
+            for dim_type, column in agg.iter_dimensions_to_keep():
+                if dim_type != pivoted_dimension_type:
+                    columns.append(column)
+            if not columns:
+                continue
+            df = self.add_columns(df, columns, context, True)
+            pivoted_columns = set(context.get_pivoted_columns(dataset_id=self.dataset_id))
+            group_by_cols = []
+            for column in columns:
+                if column.dimension_query_name not in pivoted_columns:
```
lixiliu, line 174: I don't think you need this because `columns` append only dims of non-pivoted dim type.

dsgrid/dataset/pivoted_table.py
```diff
@@ -0,0 +1,197 @@
...
+        pivoted_dimension_type = context.get_pivoted_dimension_type(dataset_id=self.dataset_id)
+        for agg in aggregations:
+            columns = []
+            for dim_type, column in agg.iter_dimensions_to_keep():
+                if dim_type != pivoted_dimension_type:
+                    columns.append(column)
+            if not columns:
+                continue
+            df = self.add_columns(df, columns, context, True)
+            pivoted_columns = set(context.get_pivoted_columns(dataset_id=self.dataset_id))
```
lixiliu, line 171: you can move this outside of the for-loop

dsgrid/dataset/pivoted_table.py
```diff
@@ -0,0 +1,197 @@
...
+                    if column.function is None:
+                        expr = column.dimension_query_name
+                    else:
+                        expr = column.function(column.dimension_query_name)
+                        if column.alias is not None:
+                            expr = expr.alias(column.alias)
+                    group_by_cols.append(expr)
+                    group_by_query_names.append(column.dimension_query_name)
+            op = agg.aggregation_function
+            agg_expr = [op(x).alias(x) for x in pivoted_columns]
```
lixiliu, line 184: since pivoted_columns is set here, it might change the ordering right? Do we care? And when do you do deep copy (you did in process_pivoted_aggregations) and when do you not?

dsgrid/dataset/table_format_handler_base.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def convert_columns_to_query_names(self, df):
+        # All columns start off as base dimension names but need to be query names.
+        base_to_query_name_mapping = self.project_config.get_base_dimension_to_query_name_mapping()
+        columns = set(df.columns)
+        for dim_type in DimensionType:
+            if dim_type == DimensionType.TIME:
+                time_dim = self._project_config.get_base_dimension(dim_type)
+                time_cols = time_dim.get_timestamp_load_data_columns()
+                # TODO DT: Should we enforce that projects can only have one time column?
```
lixiliu, line 66: we can't right because of Tempo? Unless we always explode time cols to timestamps

dsgrid/dimension/dimension_filters.py
```diff
@@ -0,0 +1,230 @@
...
+        return str(value)
+
+    @staticmethod
+    def _make_values_str(values):
+        if isinstance(values[0], str):
+            text = ", ".join((f"'{x}'" for x in values))
+        elif isinstance(values[0], int) or isinstance(values[0], float):
+            text = ", ".join((f"{x}" for x in values))
+        else:
+            raise Exception(f"Unsupported type: {type(values[0])}")  # TODO: dsg exception
```
lixiliu, line 41: Make this func consistent with `_make_value_str(value)` above? If you have unsupported type here, I'd imagine it's the same above. Better if `_make_values_str(values)` = a func of `_make_value_str(value)`

dsgrid/dimension/dimension_filters.py
```diff
@@ -0,0 +1,230 @@
...
+        title="negate",
+        description="Change the filter to match the negation of the value.",
+        default=False,
+    )
+
+    @validator("operator")
+    def check_operator(cls, operator):
+        return check_operator(operator)
+
+    def apply_filter(self, df, column=None):
```
lixiliu, line 173: Have you tested this method? I feel like it's missing sth (e.g., where do you specify the column/dimension)? Or I just need more documentation.

dsgrid/dimension/dimension_filters.py
```diff
@@ -0,0 +1,230 @@
...
+    for cls in filter_class.__subclasses__():
+        if cls == _DimensionFilterWithWhereClauseModel:
+            # Recurse.
+            subclasses = _get_filter_subclasses(_DimensionFilterWithWhereClauseModel, subclasses)
+        else:
+            subclasses[str(cls.__name__)] = cls
+    return subclasses
+
+
+def make_dimension_filter(values: Dict):
```
lixiliu, line 222: More of a question, what does this do?

dsgrid/project.py
```diff
@@ -125,8 +164,80 @@ def unload_dataset(self, dataset_id):
...
+            return None
+
+        # All dataset columns need to be in the same order.
+        context.consolidate_dataset_metadata()
+        # TODO: Change this when we support long format.
+        pivoted_columns = sorted(context.get_pivoted_columns())
+        dim_columns = context.get_all_dimension_query_names()
+        time_columns = context.get_dimension_query_names(DimensionType.TIME)
+        dim_columns -= time_columns
+        for col in dim_columns:
```
lixiliu, line 195: this could be refactored into context, sth to get cols that not neither pivoted or time

dsgrid/project.py
```diff
@@ -125,8 +164,80 @@ def unload_dataset(self, dataset_id):
...
+            dimension_type = self._config.get_dimension(col).model.dimension_type
+            if dimension_type == context.get_pivoted_dimension_type():
+                dim_columns.remove(col)
+                break
+        dim_columns = sorted(dim_columns)
+        time_columns = sorted(time_columns)
+        expected_columns = time_columns + pivoted_columns + dim_columns
+        for dataset_id, i in zip(context.model.project.dataset_ids, range(len(dfs))):
```
lixiliu, line 203: this could be:
```
for i, dataset_id in enumerate(context.model.project.dataset_ids):
```

dsgrid/project.py
```diff
@@ -125,8 +164,80 @@ def unload_dataset(self, dataset_id):
...
+            remaining = sorted(set(dfs[i].columns).difference(expected_columns))
```
lixiliu, line 204: Can you sort a set? Also NIT: if you make `expected_columns` in the order you want, you can set `final_columns = expected_columns + remaining`

dsgrid/project.py
```diff
@@ -125,8 +164,80 @@ def unload_dataset(self, dataset_id):
...
+            final_columns = time_columns + dim_columns + pivoted_columns + remaining
+            missing = context.get_pivoted_columns(dataset_id=dataset_id).difference(dfs[i].columns)
+            for column in missing:
```
lixiliu, line 207: have you tried concaving dfs with some cols missing? Wouldn't the missing cols just turn into NAN cols?

lixiliu: Hmm, I think online resources say if schemas don't match (for any reason) --> error

dsgrid/project.py
```diff
@@ -125,8 +164,80 @@ def unload_dataset(self, dataset_id):
...
+
+    def _build_filtered_record_ids_by_dimension_type(self, context: QueryContext):
+        record_ids = {}
+        base_dim_to_query_name_mapping = self._config.get_base_dimension_to_query_name_mapping()
+
+        for dim_filter in context.model.project.dimension_filters:
+            dim_type = dim_filter.dimension_type
+            query_name = dim_filter.dimension_query_name
+            df = self._config.get_dimension_records(query_name)
+            df = dim_filter.apply_filter(df).select("id")
```
lixiliu, line 230: I feel like you had a filter model that defaults to id col? If so, should select be based on input for that?

dsgrid/project.py
```diff
@@ -125,8 +164,80 @@ def unload_dataset(self, dataset_id):
...
+            if query_name != base_dim_to_query_name_mapping[dim_type]:
+                mapping_records = self._config.get_base_to_supplemental_mapping_records(query_name)
+                df = (
+                    mapping_records.join(df, on=mapping_records.to_id == df.id)
+                    .selectExpr("from_id AS id")
```
lixiliu, line 235: How did you resolve df.id and renaming "from_id" as if from mapping_records?

dsgrid/project.py
```diff
@@ -125,8 +164,80 @@ def unload_dataset(self, dataset_id):
...
+                    .distinct()
+                )
+            if dim_type in record_ids:
```
lixiliu, line 238: I thought it's allowing supplemental dimensions here too? Doesn't that create more than one instance of a dimension type, potentially?

dsgrid/query/models.py
```diff
@@ -0,0 +1,396 @@
...
+
+class FilteredDatasetModel(DSGBaseModel):
+
+    dataset_id: str = Field(description="Dataset ID")
+    filters: List[Any] = Field(
+        description="Dimension filters to apply to the dataset'",
+    )
+
+
+# TODO: setting alias on a groupBy column doesn't work on a per-dataset aggregation
```
lixiliu, line 25: remove alias option from per-dataset agg model then?

dsgrid/query/models.py
```diff
@@ -0,0 +1,396 @@
...
+        return alias
+
+    def dict(self, *args, **kwargs):
+        data = super().dict(*args, **kwargs)
+        if data["function"] is not None:
+            data["function"] = data["function"].__name__
+        return data
+
+    def get_column_name(self):
+        if self.alias is not None:
```
lixiliu, line 68: none, null interchangeable? I don't get the `getattr`, feels like magic to me

dsgrid/query/models.py
```diff
@@ -0,0 +1,396 @@
...
+            return self.alias
+        if self.function is None:
+            return self.dimension_query_name
+        return f"{self.function.__name__}__{self.dimension_query_name})"
+
+
+class DimensionQueryNamesModel(DSGBaseModel):
+    """Defines columns to include in a SQL aggregation statement. If a value is empty, that
```
lixiliu, line 76: Suggested addition: This defines the list of dimensions the value columns should be aggregated to.

dsgrid/query/peak_load_report.py
```diff
@@ -0,0 +1,56 @@
...
+
+        df = read_dataframe(filename)
+        expr = [F.max(x).alias(x) for x in value_columns]
+        peak_load = df.groupBy(*inputs.group_by_columns).agg(*expr)
+        join_cols = inputs.group_by_columns + value_columns
+        time_columns = context.get_dimension_query_names(DimensionType.TIME)
+        diff = time_columns.difference(df.columns)
+        if diff:
+            raise Exception(f"BUG: expected time column(s) {diff} are not present in table")
+        with_time = peak_load.join(df.select(*time_columns, *join_cols), on=join_cols).sort(
```
lixiliu, line 50: clever

lixiliu (commented):
most of the way -- still need to review query context, submitter, dataset

dsgrid/dataset/dataset_schema_handler_standard.py
daniel-thom: Result from discussion:
- Strip the null rows from load_data_lookup at registration time and insert them into a new file. All future queries will never have to worry about null rows.
- Add a new feature where the dataset contributor can create that secondary file up front and register the pair.

dsgrid/dataset/dataset_schema_handler_standard.py
daniel-thom:
```
pivoted_columns = self.get_pivoted_dimension_columns_mapped_to_project()
assert not self.get_pivoted_dimension_columns_mapped_to_project().difference(ld_df.columns)
```

dsgrid/dataset/pivoted_table.py
daniel-thom: Resolution: defer all work to later. There is a lot of code to write to handle all possible mappings. Also need test data.

dsgrid/dataset/pivoted_table.py
```diff
@@ -0,0 +1,197 @@
...
+        for agg in aggregations:
+            base_query_names = self.project_config.get_base_dimension_query_names()
+            dimension_query_name = None
+            for _, column in agg.iter_dimensions_to_keep():
+                query_name = column.dimension_query_name
+                # No work is required if the user requested the base dimension for the pivoted
+                # dimension.
+                if (
+                    query_name not in base_query_names
+                    and query_name not in context.get_dimension_query_names(pivoted_dim_type)
```
daniel-thom, line 109: It may not be obvious to readers that `context.get_dimension_query_names` refers to already-processed columns.

dsgrid/dataset/pivoted_table.py
daniel-thom: Idea:

Create a new class to store dimension query name information. Store this for every query name in the context. Add accessor methods. Add methods like `should_process_aggregation(query_name)`.
```
{
    "query_name1": {"dimension_type": "base", "format_type": "pivoted"},
     "query_name2": {"dimension_type": "supplemental", "format_type": "stacked"}
}
```

dsgrid/dataset/pivoted_table.py
daniel-thom: This could jumble the column order. No correctness issue, but we may want to preserve the order. Consider this.

dsgrid/dataset/table_format_handler_base.py
daniel-thom: The problem here is that for all stacked dimensions there is one column per dimension, and so we can rename the column to be the query name. Is this true for time? @elainethale @mooneyme Do you have an opinion?

dsgrid/dimension/dimension_filters.py
```diff
@@ -0,0 +1,230 @@
...
+
+    @abc.abstractmethod
+    def apply_filter(self, df):
+        """Apply the filter to a DataFrame"""
+
+    @staticmethod
+    def _make_value_str(value):
+        if isinstance(value, str):
+            return f"'{value}'"
+        return str(value)
```
daniel-thom, line 32: Need to fail if this is not str, int, or float

dsgrid/dimension/dimension_filters.py
daniel-thom: The `between` case needs to cover the case where `value` is a list. Current answer: not handled because we don't support this yet.

dsgrid/project.py
daniel-thom: ok

dsgrid/query/models.py
daniel-thom: We need it. But, I will add code to fail an operation that includes alias for per-dataset agg.

dsgrid/dataset/dataset_schema_handler_standard.py
daniel-thom: Resolving because of #148

dsgrid/project.py
daniel-thom: Line 239 allows multiple instances of one dimension type: base and supplemental.

dsgrid/dataset/pivoted_table.py
daniel-thom: @lixiliu In the best case we can track whether it's a base or supplemental dimension. We are already tracking pivoted columns per dataset and for the project (they could be different across datasets - and the project might want to change to long format). We already have a mechanism to identify base vs supplemental from the project config. Maybe that could be simplified...but not by much.

dsgrid/dataset/long_table.py
```diff
@@ -0,0 +1,14 @@
...
+class LongTable(DatasetTableBase):
+    """Implements behavior for datasets in long format."""
+
+    def __init__(self, df, dimension_records, dimension_mapping_records):
+        super().__init__(df)
+        self._dimension_records = dimension_records
+        self._dimension_mapping_records = dimension_mapping_records
+
+    def process_aggregations(self, aggregations: AggregationModel):
+        assert False
```
elainethale, line 14: Is this inheriting most of its functionality from DatasetTableBase, or is this essentially not implemented yet?

dsgrid/dataset/table_format_handler_base.py
elainethale: I think we decided to leave as-is for now and revisit with TEMPO time PR

dsgrid/query/peak_load_report.py
```diff
@@ -0,0 +1,56 @@
+import logging
```
elainethale, line 1: If this file were named report_peak_load.py, that would start to group reports in a similar spot.

elainethale (approved):
Skimmed through and left a few questions/comments. One more question: 

- Why is dataset/table_format_handler_factor.py empty?

dsgrid/dataset/long_table.py
daniel-thom: Not implemented yet

lixiliu (approved):
Thanks Dan!

dsgrid/dataset/pivoted_table.py
daniel-thom: The comment on lines 105-6 addresses it, I hope.

PR #145, opened 2022-08-30
Ll/update doc

lixiliu:
Completes JIRA Story DSGRID-
Closes GitHub Issue #

## Checklist

- [ ] All tests pass

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -23,13 +23,13 @@ jobs:
           sudo apt install -y pandoc
           python -m pip install --upgrade pip
           python -m pip install '.[admin]'
-          git clone https://${{ secrets.ACCESS_PROJECTS_TOKEN }}@github.com/dsgrid/dsgrid-project-StandardScenarios.git
       - name: build documentation
         run: |
           git submodule init
           git submodule update
           cd docs
           make clean
+          git clone https://${{ secrets.ACCESS_PROJECTS_TOKEN }}@github.com/dsgrid/dsgrid-project-StandardScenarios.git
           make html
       - name: deploy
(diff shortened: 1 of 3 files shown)


mooneyme:
@lixiliu is this ready for review?

mooneyme:
Update from dsgrid meeting today: 

- I will take over this branch to add RST stuff. 
- We prefer to use standard scenarios instead of tests because it includes tomls in the format that users will find helpful as a reference for how they should configure their configs. Things in the test data repo are taken from the format registered by dsgrid and therefore includes additional things that dsgrid does during the process. It also does not include the faster way of registering a dataset in a way that also registers the dimensions. At some point in a more stable future, we may want to find a better way to incorporate these user formatted standard scenario configs in the test data, but this is not a priority right now.
-  I will check to make sure that Lixi's additions of using standard standard scenarios even though it is a private repo will work.

lixiliu:
> Update from dsgrid meeting today:
> 
> * I will take over this branch to add RST stuff.
> * We prefer to use standard scenarios instead of tests because it includes tomls in the format that users will find helpful as a reference for how they should configure their configs. Things in the test data repo are taken from the format registered by dsgrid and therefore includes additional things that dsgrid does during the process. It also does not include the faster way of registering a dataset in a way that also registers the dimensions. At some point in a more stable future, we may want to find a better way to incorporate these user formatted standard scenario configs in the test data, but this is not a priority right now.
> * I will check to make sure that Lixi's additions of using standard standard scenarios even though it is a private repo will work.

I think Dan set up the secret access token in the last commit.

elainethale:
@lixiliu, is this PR obsolete?

lixiliu:
Obsolete PR

PR #147, opened 2022-09-13
rename sector_models to modeled

mooneyme:
Refactor dsgrid code to use `modeled` instead of `sector_models`. 

....
Completes JIRA Story DSGRID-
Closes GitHub Issue #

## Checklist

- [x] All tests pass
- [x] Consider transferring TODOs to JIRA or GitHub

### DIFF
--- a/dsgrid/config/project_config.py
+++ b/dsgrid/config/project_config.py
@@ -182,12 +182,13 @@ class InputDatasetModel(DSGBaseModel):
     )
     # TODO this model_sector must be validated in the dataset_config
     # TODO: this is only necessary for input_model types
-    model_sector: str = Field(  # TODO: should this be data_source instead?
+    model_sector: str = Field(  # TODO: should this be data_source instead? -- still relevant?
         # TODO: need to discuss with team why this is needed. One potential reason is because it is
         # helpful to query at some point which datasets have not yet been registered. Dataset_id may
         # not be all that descriptive, but the data_source is. We may also want the metric_type here too.
         title="model_sector",
         description="Model sector ID, required only if dataset type is ``sector_model``.",
+        # -- still relevant?
         optional=True,
         updateable=False,
     )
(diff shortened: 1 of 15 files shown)


dsgrid/config/project_config.py
```diff
@@ -182,12 +182,13 @@ class InputDatasetModel(DSGBaseModel):
...
+    model_sector: str = Field(  # TODO: should this be data_source instead? -- still relevant?
         # TODO: need to discuss with team why this is needed. One potential reason is because it is
         # helpful to query at some point which datasets have not yet been registered. Dataset_id may
         # not be all that descriptive, but the data_source is. We may also want the metric_type here too.
         title="model_sector",
         description="Model sector ID, required only if dataset type is ``sector_model``.",
+        # -- still relevant?
         optional=True,
         updateable=False,
     )
```
mooneyme, line 194: Is all of this still relevant if we make the switch to "modeled" instead of "sector_model"?

mooneyme (commented):
Open comments below.

dsgrid/config/project_config.py
elainethale: Does this currently have to map to a sector listed in the project base dimensions? If so, that's probably overly strict and we should drop it. But if it's just a loose tag, it might be nice to keep.

dsgrid/config/project_config.py
```diff
@@ -198,6 +198,7 @@ def check_version(cls, version):
 
     @validator("model_sector")
     def check_model_sector(cls, model_sector, values):
+        # TODO: how does change to "modeled" get set here? is this validator still relevant?
         if "dataset_type" in values:
             if not values["dataset_type"] == InputDatasetType.SECTOR_MODEL:
                 raise ValueError("model_sector is only required if dataset_type is 'sector_model'")
```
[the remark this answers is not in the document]
elainethale: I think it could still be reasonable to ask for an optional "sector" tag for "modeled" data sets.

Of course, historical and benchmark data sets can also have sectors, e.g., "electricity," "buildings," "all" ... Maybe it should just be dataset_sector and be an optional metadata tag? Does that make sense? 

Maybe it doesn't make sense here in the project config but makes more sense in the dataset config, because it's metadata and not necessarily something we want to enforce a strict match on between project and dataset?

elainethale (approved):
Let's have a quick conversation about model_sector. Otherwise, this looks good.

daniel-thom (approved):

elainethale:
We think model_sector should be deleted from the InputSectorModel project config and something like it should be added as a metadata tag to the dataset config to be a quick informal indicator of whether a dataset applies to "buildings", "residential", "electricity", "all", etc.

PR #150, opened 2022-09-28
WIP: Add REST API

daniel-thom:


dsgrid/config/dataset_config.py
```diff
@@ -271,7 +271,7 @@ class DatasetConfigModel(DSGBaseModel):
         title="origin_contributors",
         description="List of origin data contributor's first and last names"
         """ e.g., ["Harry Potter", "Ronald Weasley"]""",
-        required=False,
+        default=[],
```
daniel-thom, line 274: These caused schema errors for FastAPI.

dev/README.md
```diff
@@ -527,10 +525,80 @@ $ dsgrid query project run \
...
+This is a bit more complicated because that tool needs to be able to locate the Python script (`dsgrid` in this case)
+and detect that it is a Python script.
+
+1. Find the location of your `dsgrid` command.
+```
+$ which dsgrid
+/Users/dthom/miniconda3/envs/dsgrid/bin/dsgrid
+```
+
+2. Substitute `dsgrid-cli.py` for the usual `dsgrid`. This allows `spark-submit` to detect that it is Python.
```
daniel-thom, line 538: Tell me if this workaround absolutely drives you crazy. If the average non-dsgrid-developer interacts with dsgrid through the web API, it won't be a problem.

tests/conftest.py
```diff
@@ -59,6 +61,14 @@ def _get_latest_commit():
     return commit
 
 
+@pytest.fixture(scope="module")
+def setup_api_server():
+    yield
+    for path in (os.environ["DSGRID_QUERY_OUTPUT_DIR"], os.environ["DSGRID_API_SERVER_STORE_DIR"]):
+        if os.path.exists(path):
+            shutil.rmtree(path)
```
elainethale, line 69: Should there be an option to leave these files for debugging purposes?

elainethale (approved):
Maybe we should ask Nick or Paul if they or someone else would be an appropriate reviewer / would be likely to have suggestions?

dev/README.md
daniel-thom: It depends on how big the data is. Submitting ComStock to StandardScenarios can be done on a laptop with Spark in local mode. Submitting TEMPO requires one real compute node with a Spark cluster. I'm hesitant to write a bunch of documentation with recommendations until we understand the workflow better. I am inclined to say that all jobs should be run with spark-submit because it provides the most straightforward way to customize the Spark parameters.

Consider one workflow that could be possible on either Eagle or AWS:
- We configure an always-on (or mostly-on), small virtual machine that runs a dsgrid web server.
- User sends an API command to perform action that requires a Spark cluster (data submission or query).
- dsgrid provisions the cluster based on some information provided by the user.
- The web server submits the command to the Spark cluster.

dev/README.md
```diff
@@ -527,10 +525,80 @@ $ dsgrid query project run \
...
+```
+$ export DSGRID_LOCAL_REGISTRY=~/.dsgrid-registry
+$ export DSGRID_QUERY_OUTPUT_DIR=api_query_output
+$ export DSGRID_API_SERVER_STORE_DIR=.
+```
+
+Start the API server with this command:
+```
+$ uvicorn dsgrid.api.app:app
+```
```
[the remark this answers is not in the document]
daniel-thom: yes (but haven't tested on Windows) and yes

tests/conftest.py
daniel-thom: I typically set breakpoints that get triggered before this occurs. I'm sure that what you're asking can be done with pytest, but I would have to investigate.

PR #151, opened 2022-10-04
Mm/fix  check fraction sum

mooneyme:
Fix the _check_fraction_sum dimension_mapping validation to check for fraction sums that are ~1 instead of == 1. This resolves the issue around floats data types and precision.

Completes JIRA Story DSGRID-
Closes GitHub Issue #

## Checklist

- [ ] All tests pass

### DIFF
--- a/dsgrid/config/dimension_mapping_base.py
+++ b/dsgrid/config/dimension_mapping_base.py
@@ -168,14 +168,14 @@ class DimensionMappingBaseModel(DSGBaseModel):
         title="to_dimension",
         description="To dimension",
     )
-    from_fraction_tolerance: Optional[float] = Field(
+    from_fraction_tolerance: float = Field(
         title="from_fraction_tolerance",
-        description="Tolerance value to apply to the from_fraction column",
+        description="Tolerance to apply when checking from_fraction column sums",
         default=1e-9,
     )
-    to_fraction_tolerance: Optional[float] = Field(
+    to_fraction_tolerance: float = Field(
         title="to_fraction_tolerance",
-        description="Tolerance value to apply to the to_fraction column",
+        description="Tolerance to apply when checking to_fraction column sums",
         default=1e-9,
     )
     description: str = Field(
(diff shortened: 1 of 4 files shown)


mooneyme:
Meghan TODO:
- tolerance variable should be configurable
- Set tolerance option for any mapping_types which have a from_fraction or to_fraction.  default = 10e-9

mooneyme:
> Meghan TODO:
>
> tolerance variable should be configurable
> Set tolerance option for any mapping_types which have a from_fraction or to_fraction. default = 10e-9

Done.

dsgrid/config/dimension_mapping_base.py
```diff
@@ -168,6 +168,16 @@ class DimensionMappingBaseModel(DSGBaseModel):
         title="to_dimension",
         description="To dimension",
     )
+    from_fraction_tolerance: Optional[float] = Field(
```
daniel-thom, line 171: Let's try to not use `Optional` in any case where there is a default value. `Optional` in this case is equivalent to `Union[None, float]`. In reality, this can only be a float and we don't want to allow `None`. There could be cases where we still want to allow `None`.

daniel-thom (approved):

dsgrid/config/dimension_mapping_base.py
```diff
@@ -168,6 +168,16 @@ class DimensionMappingBaseModel(DSGBaseModel):
...
+        title="from_fraction_tolerance",
+        description="Tolerance value to apply to the from_fraction column",
```
elainethale, line 173: "Tolerance to apply when checking from_fraction column sums"? (And similar edit for to_fraction_tolerance description?)

elainethale (approved):
Minor text edits suggested.

dsgrid/config/dimension_mapping_base.py
mooneyme: Ok so if we have a default value, we don't need optional? 10-4

PR #153, opened 2022-10-06
Ll/project time explosion

lixiliu:
Completes JIRA Story DSGRID-
Closes GitHub Issue #

## Description
Added time conversion from dataset to project:
- convert representative time to project (Tempo), test added
- date_time_dimension_config should be able to handle tz conversion when dataset time dim != project time_dim but need to add test for that
- test added for Comstock, which test the case where dataset time_dim == project time_dim
Added test for tempo time conversion to project

## Checklist

- [ ] All tests pass


dsgrid/config/representative_period_time_dimension_config.py
```diff
@@ -44,10 +48,86 @@ def check_dataset_time_consistency(self, load_data_df):
...
+        # project_time_len = project_time_df.count()
+        # dataset_time_len = self.get_time_dataframe().count()
+        # df_len = df.count()
+        # dfn_len = dfn.count()
+        # expected_len = df_len / dataset_time_len * project_time_len
+        # assert dfn_len == expected_len, f"df with exploded time has len={dfn_len}, but expecting {expected_len}, check join logic!"
+
+        df_check = df.groupBy(ptime_col).count().sort(ptime_col)
+        freq_count = df_check.select("count").distinct().collect()
+        assert len(freq_count) == 1, freq_count
```
daniel-thom, line 125: This check fails on my system if the Spark Session has a Denver time zone. It passes when I set it to UTC. I'm OK if we enforce UTC, but it should be explicit (or maybe we can set it).
```
df_check.filter("count != 212").show()
+-------------------+-----+
|          timestamp|count|
+-------------------+-----+
|2012-03-10 21:00:00|  424|
+-------------------+-----+
```

setup.py
```diff
@@ -85,6 +85,8 @@
         "sqlalchemy",
         "toml",
         "uvicorn",
+        "tzdata",  # time zone stuff
+        "pyarrow",
```
daniel-thom, line 89: Do we really need pyarrow? We don't ever `pandas.read_parquet` in dsgrid. If you want to use it in tests, you can add it the `dev` extras in this file.

tests/test_project_time.py
```diff
@@ -0,0 +1,161 @@
...
+    time_range_ts = time_range.end.tz_localize(None)
+    assert (
+        time_df_ts == time_range_ts
+    ), f"Ending timestamp does not match: {time_df_ts} vs. {time_range_ts}"
+
+
+def check_exploded_tempo_time(project_time_dim, load_data):
+    """
+    - DF.show() (and probably all arithmetics) use spark session time
+    - DF.toPandas() has weird behavior (i.e., 2012-11-04 07:00 gets listed as 08:00, DThom is filing a bug report to spark)
```
daniel-thom, line 98: Needs update based on what we learned today.

tests/test_project_time.py
```diff
@@ -0,0 +1,161 @@
...
+    )
+
+    mismatch = time[time.isna().any(axis=1)]
+    assert n_model == 366 * 24, n_model
+    assert (
+        len(mismatch) == 0
+    ), f"Mismatch:\nn_model={n_model}, n_project={n_project}, n_tempo={n_tempo}\n{mismatch}"
+
+    # delete files
+    shutil.rmtree(project_time_file)
```
daniel-thom, line 160: Need to ensure these files get deleted if an exception occurs before this point. try/finally or a fixture

tests/test_project_time.py
```diff
@@ -0,0 +1,161 @@
...
+    / "simple_standard_scenarios"
+)
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.fixture
+def registry_mgr():
+    return RegistryManager.load(REGISTRY_PATH, offline_mode=True)
+
```
daniel-thom, line 27: I think we need at least two more tests:
1. Verification that the TEMPO annual load data values are correct (not just time).
2. Verification that a query asking for comstock + resstock + tempo works. This can be a simple adjustment to one query in tests/test_query.py.

dsgrid/config/date_time_dimension_config.py
```diff
@@ -50,8 +57,89 @@ def check_dataset_time_consistency(self, load_data_df):
...
+            schema = StructType(
+                [
+                    StructField(time_col, StringType(), False),
+                    StructField("timezone", StringType(), False),
+                ]
+            )
+            raise ValueError("TimeZone = LOCAL or NONE needs fixing")
+
+        model_time = self.list_expected_dataset_timestamps()
+        spark = SparkSession.builder.appName("dgrid").getOrCreate()
```
daniel-thom, line 77: Consider using dsgrid/utills/spark.py. The existing dsgrid _get_active_session could be made public. Consider that SparkSession.getActiveSession will never create a new session but getOrCreate might.

emr/bootstrap-pyspark
```diff
@@ -104,7 +104,7 @@ pyarrow \
 s3fs \
 boto3 \
 pandas \
-pytz \
+tzdata \
```
daniel-thom, line 107: Is this a Windows-only requirement? I could be wrong but I thought I read that Linux would already have the time database.

dsgrid/dataset/dataset_schema_handler_standard.py
```diff
@@ -60,6 +60,7 @@ def get_unique_dimension_rows(self):
     def make_project_dataframe(self):
         # TODO: Can we remove NULLs at registration time?
         lk_df = self._load_data_lookup.filter("id is not NULL")
+        lk_df = self._add_time_zone(lk_df)
```
daniel-thom, line 63: This needs to be done in `make_project_dataframe_from_query` also.

dsgrid/dataset/dataset_schema_handler_standard.py
```diff
@@ -70,7 +71,7 @@ def make_project_dataframe(self):
         )
         # TODO: handle fraction application
         # Currently this requires fraction = 1.0
-        ld_df = ld_df.join(lk_df, on="id").drop("id")
+        ld_df = ld_df.join(lk_df, on="id")
```
daniel-thom, line 74: Keeping `id` will be problematic if/when we take a union of datasets where one is in single-table format. Also, it could be confusing to users about when a dataset has `id` and when it does not.

dsgrid/dataset/dataset_schema_handler_standard.py
daniel-thom, line 63: We need to think about reordering the actions in these functions. Converting time likely has to occur immediate after filtering.

daniel-thom, line 63: Per our informal discussion today, we need something defined at this level that tells us whether we are delivering time as experienced by people in specific time zones or as absolute time. It will change how we perform aggregations.

dsgrid/config/date_time_dimension_config.py
```diff
@@ -50,8 +57,89 @@ def check_dataset_time_consistency(self, load_data_df):
...
+        # model_time2 = model_time[7385:7385+24]
+        # df_time2 = spark.createDataFrame(model_time2, schema=schema)
+        # test = df_time2.withColumn("UTC", F.to_utc_timestamp(F.col("timestamp"), session_tz)).withColumn("project", F.from_utc_timestamp(F.col("UTC"), self.model.timezone.tz_name))
+        # test.write.parquet("/Users/lliu2/Downloads/time_pyspark.parquet")
+        # test.toPandas().head(20)
+        # import pandas as pd; pd.DataFrame(test.collect()).head(20)
+        # breakpoint()
+
+        # convert to dataset timezone
+        df_time = self._convert_time_zone(
```
daniel-thom, line 98: I'm not sure about this function call. At this point, for tempo with dsgrid_conus_2022, `model_time` starts at `datetime.datetime(2012, 1, 1, 0, 0, tzinfo=zoneinfo.ZoneInfo(key='Etc/GMT+5'))`. `df_time` has those exact same absolute points in time. It isn't really converting to EST. Maybe this goes back to the question of absolute time vs local time.

dsgrid/dataset/dataset_schema_handler_standard.py
lixiliu: we should always keep id

dsgrid/dimension/time.py
```diff
@@ -291,3 +351,27 @@ def make_time_range(start, end, frequency, leap_day_adjustment):
...
         return NoOpTimeRange(start, end, frequency, leap_day_adjustment)
     return DatetimeRange(start, end, frequency, leap_day_adjustment)
+
+
+def get_timezone(tz_value: str) -> TimeZone:
+    """convert GEOGRAPHY record time_zone col str to dsgrid TimeZone object"""
+    for tzo in TimeZone:
+        if tz_value == tzo.value:
+            return tzo
+    raise ValueError(f"tz={tz_value} cannot be converted to a TimeZone object")
```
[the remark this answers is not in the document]
lixiliu: yes

dsgrid/utils/dataset.py
```diff
@@ -108,6 +108,16 @@ def map_and_reduce_pivoted_dimension(df, records, pivoted_columns, operation, re
     return df.selectExpr(*nonvalue_cols, *extra_cols, *exprs), sorted(final_columns), dropped
 
 
+@track_timing(timer_stats_collector)
+def add_column_from_records(df, dimension_records, dimension_name, column_to_add):
+    df = df.join(
+        dimension_records.select(F.col("id").alias("record_id"), column_to_add),
+        on=F.col(dimension_name) == F.col("record_id"),
+        how="left",
+    ).drop("record_id")
```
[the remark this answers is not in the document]
lixiliu: Don't think we need to, I believe dim records are already validated during registration so records should contain >= what's in df.

emr/bootstrap-pyspark
lixiliu: I guess so?

dsgrid/dataset/dataset_schema_handler_standard.py
lixiliu: yeah I see your point. "id" are so useful though. We should require it in single-table as well.

tests/test_project_time.py
lixiliu: Need help with 2

emr/bootstrap-pyspark
daniel-thom: I don't think it will be needed for this EMR Linux image.

It is generally recommended that any time zone libraries should attempt to use the system data before using the tzdata package, but some systems (notably Windows) do not deploy zoneinfo binaries of this type, and so tzdata is necessary.
https://tzdata.readthedocs.io/en/latest/#overview

dsgrid/config/annual_time_dimension_config.py
```diff
@@ -49,7 +51,20 @@ def check_dataset_time_consistency(self, load_data_df):
                 f"load_data {time_col}s do not match expected times. mismatch={mismatch}"
             )
 
-    def convert_dataframe(self, df, project_time_dim):
+    def get_time_dataframe(self):
```
daniel-thom, line 54: This method should probably be defined as an abstractmethod in the base class.

dsgrid/config/annual_time_dimension_config.py
```diff
@@ -49,7 +51,20 @@ def check_dataset_time_consistency(self, load_data_df):
...
+        schema = StructType([StructField(time_col, IntegerType(), False)])
+
+        model_time = self.list_expected_dataset_timestamps()
+        df_time = _get_spark_session.createDataFrame(model_time, schema=schema)
+        return df_time
+
+    def get_time_dataframe_in_model_timezone(self):
+        return self.get_time_dataframe()
+
+    def convert_dataframe(self, df=None, project_time_dim=None, df_meta=None):
```
daniel-thom, line 67: Can you add df_meta to the abstractmethod in the base class and add a docstring to explain what it is?

dsgrid/utils/dataset.py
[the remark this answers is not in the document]
daniel-thom: So then why a left join instead of innner join?

tests/test_project_time.py
```diff
@@ -0,0 +1,251 @@
...
+    n_ts = raw_data_df.groupby(groupby_cols)["count"].sum().unique()
+    assert list(n_ts) == [
+        len(model_time)
+    ], f"Mismatch in number of timestamps: {n_ts} vs. {len(model_time)}"
+
+    raw_data_df[enduse_cols] = raw_data_df[enduse_cols].multiply(raw_data_df["count"], axis=0)
+    raw_sum_df = raw_data_df.groupby(groupby_cols)[enduse_cols].sum().sort_index()
+
+    # compare annual sums
+    delta_df = converted_sum_df - raw_sum_df
```
daniel-thom, line 193: Is there any chance of minor differences in these numbers?

dsgrid/config/representative_period_time_dimension_config.py
```diff
@@ -44,10 +48,76 @@ def check_dataset_time_consistency(self, load_data_df):
...
+        session_tz = _get_spark_session().conf.get("spark.sql.session.timeZone")
+        idx = 0
+        for tz_value, tz_name in zip(geo_tz_values, geo_tz_names):
+            local_time_df = project_time_df.withColumn("time_zone", F.lit(tz_value)).withColumn(
+                "local_time",
+                F.from_utc_timestamp(F.to_utc_timestamp(F.col(ptime_col), session_tz), tz_name),
+            )
+            select = [ptime_col, "time_zone"]
+            for col in time_cols:
+                func = col.replace("_", "")
```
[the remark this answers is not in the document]
lixiliu: yes, other func like day of month, etc are still one word in PySpark.

dsgrid/config/date_time_dimension_config.py
```diff
@@ -50,8 +56,59 @@ def check_dataset_time_consistency(self, load_data_df):
...
+
+        if self.model.timezone in [TimeZone.LOCAL, TimeZone.NONE]:
+            # TODO: handle local time zone
+            schema = StructType(
+                [
+                    StructField(time_col, StringType(), False),
+                    StructField("timezone", StringType(), False),
+                ]
+            )
+            raise ValueError("TimeZone = LOCAL or NONE needs fixing")
```
elainethale, line 77: This should probably be a NotImplementedError

dsgrid/config/representative_period_time_dimension_config.py
```diff
@@ -60,6 +130,7 @@ def get_timestamp_load_data_columns(self):
         return self._format_handler.get_timestamp_load_data_columns()
 
     def get_tzinfo(self):
+        # TBD
```
elainethale, line 133: Should this just return TimeZone.LOCAL?

dsgrid/config/representative_period_time_dimension_config.py
```diff
@@ -44,10 +48,76 @@ def check_dataset_time_consistency(self, load_data_df):
...
+            else:
+                time_df = time_df.union(local_time_df)
+            idx += 1
+
+        # join all
+        join_keys = time_cols + ["time_zone"]
+        select = [
+            col if col not in time_cols else F.col(col).cast(IntegerType()) for col in df.columns
+        ]
+        df = df.select(*select).join(time_df, on=join_keys).drop(*join_keys)
```
elainethale, line 119: I can't pretend to understand this code just by reading it. I trust daylight savings time is getting handled somewhere?

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -248,8 +252,20 @@ def _remap_dimension_columns(self, df, pivoted_columns=None):
...
+    def _add_time_zone(self, load_data_df):
+        geography_dim = self._config.get_dimension(DimensionType.GEOGRAPHY)
+        geo_records = geography_dim.get_records_dataframe()
+        geo_name = geography_dim.model.dimension_type.value
+        load_data_df = add_column_from_records(load_data_df, geo_records, geo_name, "time_zone")
+        return load_data_df
+
+    @track_timing(timer_stats_collector)
+    def _convert_time_dimension(self, load_data_df, load_data_lookup=None):
         # This needs to convert the time format as well as time zone (TODO).
```
elainethale, line 264: Remove this todo comment?

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -248,8 +252,20 @@ def _remap_dimension_columns(self, df, pivoted_columns=None):
...
         time_dim = self._config.get_dimension(DimensionType.TIME)
-        time_dim.convert_dataframe(load_data_df, self._project_time_dim)
+        load_data_df = time_dim.convert_dataframe(
+            df=load_data_df,
+            project_time_dim=self._project_time_dim,
+            df_meta=load_data_lookup,
```
elainethale, line 269: What happens with this if the schema is one_table? Why isn't there similar code being added to the dataset_schema_handler_one_table.py file?

dsgrid/dataset/dataset_schema_handler_standard.py
elainethale: I don't like the idea of "id" in the one-table format. It seems highly arbitrary and easy to overwrite.

setup.py
```diff
@@ -16,7 +16,7 @@
 with open(here / "README.md", encoding="utf-8") as f:
     readme = f.read()
 
-dev_requires = ["black", "pre-commit", "devtools", "jupyter", "flake8"]
+dev_requires = ["black", "pre-commit", "devtools", "jupyter", "flake8", "pyarrow"]
```
elainethale, line 19: Why is pyarrow a dev requirement and not a general requirement?

tests/test_project_time.py
```diff
@@ -0,0 +1,251 @@
...
+    check_tempo_load_sum(project_time_dim, tempo, tempo_data, tempo_data_mapped)
+
+    # comstock time conversion
+    comstock_data = comstock._handler._add_time_zone(comstock.load_data_lookup)
+    comstock_data = comstock.load_data.join(comstock_data, on="id")
+    comstock_data = comstock_time_dim.convert_dataframe(
+        df=comstock_data,
+        project_time_dim=project_time_dim,
+    )
+    # TODO: add test for annual_time_dimension_config, res/com with local timezones
```
elainethale, line 97: Would these be in this same test or a different test? I don't really get what needs to be checked here. If we can't write these tests now, maybe document in a git issue?

tests/test_project_time.py
elainethale: Certainly reasonable to use np.isclose or math.isclose here

tests/test_project_time.py
```diff
@@ -0,0 +1,251 @@
...
+        [
+            model_time,
+            project_time.rename(columns={time_col: "project_time"}),
+            tempo_time.rename(columns={time_col: "tempo_time"}),
+        ],
+        axis=1,
+    )
+
+    mismatch = time[time.isna().any(axis=1)]
+    print(time.iloc[7390:7400, :])
```
elainethale, line 247: Why are we printing here?

dsgrid/utils/dataset.py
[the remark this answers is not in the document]
lixiliu: change to inner

tests/test_project_time.py
lixiliu: Link to test query: https://github.com/dsgrid/dsgrid/blob/main/tests/test_queries.py#L986

For #2, import code from test_project_time to get tempo data with fully exploded time. in line 1065

dsgrid/config/annual_time_dimension_config.py
lixiliu: add to base class

dsgrid/config/representative_period_time_dimension_config.py
```diff
@@ -44,10 +48,76 @@ def check_dataset_time_consistency(self, load_data_df):
...
+            return df
+
+        time_cols = self.get_timestamp_load_data_columns()
+        ptime_col = project_time_dim.get_timestamp_load_data_columns()
+        assert len(ptime_col) == 1, ptime_col
+        ptime_col = ptime_col[0]
+
+        # get unique timezones
+        if df_meta is not None:
+            df = df.join(df_meta.select("id", "time_zone"), on="id")
```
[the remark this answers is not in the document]
lixiliu: reconfigured for polymorphism

dsgrid/config/representative_period_time_dimension_config.py
```diff
@@ -44,10 +48,76 @@ def check_dataset_time_consistency(self, load_data_df):
...
+        df_time = _get_spark_session().createDataFrame(model_time, schema=schema)
+
+        return df_time
+
+    def get_time_dataframe_in_model_timezone(self):
+        return self.get_time_dataframe()
+
+    def convert_dataframe(self, df=None, project_time_dim=None, df_meta=None):
+        """convert df time column(s) based on project_time_dim.
+        df_meta is load_data_lookup with time_zone column
```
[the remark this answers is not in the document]
lixiliu: reconfigured for polymorphism, input takes in time_zone_mapping instead of df_meta

dsgrid/config/representative_period_time_dimension_config.py
lixiliu: No, the output of this func needs to be a real tzinfo value

dsgrid/config/representative_period_time_dimension_config.py
lixiliu: Yes, spark does the daylight savings handling internally, provided that the input timestamps are correct and time_zone-aware and that we stick to spark method when converting between time_zones

dsgrid/dataset/dataset_schema_handler_base.py
lixiliu: We don't have a one-table dataset to test yet until AEO growth rates get added to the test dataset, and we do not want to add codes that are not ready to be tested.

setup.py
lixiliu: [10:50 AM] Thom, Daniel
My reasons here:
Minimize the chance for package dependency conflicts. When these happen, they are a nightmare.Minimize our installation size for users.

tests/test_project_time.py
lixiliu: stale

tests/test_project_time.py
lixiliu: Do not need to round here because the load data used in either method came from tempo load_data spark df converted toPandas(), so they should retain the same level of precision.

dsgrid/dimension/time.py
```diff
@@ -294,24 +295,26 @@ def iter_timestamps(self):
...
+        end = (
+            self.end.to_pydatetime().astimezone(ZoneInfo("UTC")) + self.frequency
+        )  # to make end time inclusive
 
         while cur < end:
             if not (
                 self.leap_day_adjustment == LeapDayAdjustmentType.DROP_FEB29
-                and cur.month == 2
-                and cur.day == 29
+                and cur.astimezone(self.tzinfo).month == 2
```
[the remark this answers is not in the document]
lixiliu: no, the point was to keep cur and end in UTC for iter_timestamps (so time delta is not ambiguous), and then convert to config timezone afterward (i.e., in .list_time_range())

dsgrid/config/representative_period_time_dimension_config.py
```diff
@@ -44,10 +48,93 @@ def check_dataset_time_consistency(self, load_data_df):
...
+        df_time = _get_spark_session().createDataFrame(model_time, schema=schema)
+
+        return df_time
+
+    # def build_time_dataframe_with_time_zone(self):
+    #     return self.build_time_dataframe()
+
+    def convert_dataframe(self, df=None, project_time_dim=None, time_zone_mapping=None):
+        # in spark.dayofweek: 1=Sunday, 7=Saturday
+        # in pandas.dt.day_of_week: 0=Monday, 6=Sunday
```
daniel-thom, line 67: Pandas is probably not relevant. This is the dsgrid data format (and Python standard library).

dsgrid/config/representative_period_time_dimension_config.py
```diff
@@ -44,10 +48,93 @@ def check_dataset_time_consistency(self, load_data_df):
...
+            df = df.join(time_zone_mapping, on=key)
+        geo_tz_values = [row.time_zone for row in df.select("time_zone").distinct().collect()]
+        geo_tz_names = [get_timezone(tz).tz_name for tz in geo_tz_values]
+
+        # create time map
+        # temporarily set session time to UTC for timeinfo extraction
+        # Note: timeinfo is extracted from local_time column exactly in DF.show(),
+        # and only UTC seems to convert to local_time correctly for DF.show().
+        # Even though UTC does not always lead to correct time output when DF.toPandas()
+        # the underlying time data is correctly stored when saved to file
```
daniel-thom, line 95: Is this still relevant?

dsgrid/dimension/time.py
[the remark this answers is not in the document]
daniel-thom: I was suggesting that instead of calling `astimezone(self.tzinfo)` 6 times you could call it once at the top of the while loop.

dsgrid/dimension/time.py
```diff
@@ -264,18 +327,26 @@ def list_time_range(self):
             list of datetime
 
         """
-        return list(self.iter_timestamps())
+        utc_list = list(self.iter_timestamps())
+        return [ts.astimezone(self.tzinfo) for ts in utc_list]
```
daniel-thom, line 331: Can you do this in 1 loop iteration instead of 2?

dsgrid/dimension/time.py
```diff
@@ -291,3 +362,11 @@ def make_time_range(start, end, frequency, leap_day_adjustment):
     elif frequency == datetime.timedelta(days=0):
         return NoOpTimeRange(start, end, frequency, leap_day_adjustment)
     return DatetimeRange(start, end, frequency, leap_day_adjustment)
+
+
+def get_timezone(tz_value: str) -> TimeZone:
+    """convert GEOGRAPHY record time_zone col str to dsgrid TimeZone object"""
```
daniel-thom, line 368: Is geography record relevant here? How is this different from `TimeZone(tz_value)`?

tests/test_queries.py
```diff
@@ -1052,7 +1052,7 @@ def accumulate_stats(stats):
         },
         "max": {
             "electricity": max(
-                (com["max"]["electricity"], res["max"]["electricity"], tem["sum"]["electricity"])
+                (com["max"]["electricity"], res["max"]["electricity"], tem["max"]["electricity"])
```
daniel-thom, line 1055: Thanks!

dsgrid/config/representative_period_time_dimension_config.py
lixiliu: I think so, this is intended to explain why we are setting session tz to UTC temporarily for the time info extraction

dsgrid/dimension/time.py
lixiliu: yes! I missed the yield part of the iter_timestamps(). I was always wondering why this works, since I did not see anything returned.

daniel-thom (approved):

dsgrid/config/representative_period_time_dimension_config.py
elainethale: So will it end up always returning None? Or would you want it to return a mapping of ids of some sort to prevailing time zone objects?

PR #155, opened 2022-10-11
Refactor dimension association tables

daniel-thom:
This PR refactor dimension association tables in preparation for registering AEO growth rates.

1. Store association tables by data source.
2. Cache the tables in the spark warehouse so that they don't have to be regenerated at every dataset submission.

I also added a more generic way to register projects and datasets.

### DIFF
--- a/dsgrid/config/project_config.py
+++ b/dsgrid/config/project_config.py
@@ -6,9 +6,18 @@ from typing import Dict, List, Set
 from pydantic import Field
 from pydantic import root_validator, validator
 
+from dsgrid.config.dimension_association_manager import (
+    try_load_dimension_associations,
+    save_dimension_associations,
+)
 from dsgrid.data_models import DSGBaseModel, serialize_model_data
 from dsgrid.dimension.base_models import check_required_dimensions
-from dsgrid.exceptions import DSGInvalidField, DSGInvalidDimension, DSGInvalidParameter
+from dsgrid.exceptions import (
+    DSGInvalidField,
+    DSGInvalidDimension,
+    DSGInvalidParameter,
+    DSGInvalidDimensionAssociation,
+)
 from dsgrid.registry.common import (
     ProjectRegistryStatus,
     DatasetRegistryStatus,
@@ -16,6 +25,7 @@ from dsgrid.registry.common import (
 )
 from dsgrid.registry.dimension_registry_manager import DimensionRegistryManager
 from dsgrid.registry.dimension_mapping_registry_manager import DimensionMappingRegistryManager
+from dsgrid.utils.spark import get_unique_values
 from dsgrid.utils.timing import timer_stats_collector, track_timing
 from dsgrid.utils.utilities import check_uniqueness
 from dsgrid.utils.versioning import handle_version_or_str
@@ -330,6 +340,8 @@ class ProjectDimensionQueryNamesModel(DSGBaseModel):
 class ProjectConfig(ConfigWithDataFilesBase):
     """Provides an interface to a ProjectConfigModel."""
 
+    _DIMENSION_ASSOCIATIONS_TABLE = "dimension_associations"
+
     def __init__(self, model):
         super().__init__(model)
         self._base_dimensions = {}  # DimensionKey to DimensionConfig
@@ -362,7 +374,6 @@ class ProjectConfig(ConfigWithDataFilesBase):
         return config
 
     def load_dimensions_and_mappings(self, dimension_manager, dimension_mapping_manager):
-        self.load_dimension_associations()
         self.load_dimensions(dimension_manager)
         self.load_dimension_mappings(dimension_mapping_manager)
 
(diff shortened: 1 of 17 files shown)


dsgrid/utils/spark.py
```diff
@@ -407,3 +408,102 @@ def _get_spark_session():
...
+    if spark.catalog.tableExists(table_name, dbName=database):
+        return spark.table(table_name)
+    table = try_load_stored_table_from_warehouse(table_name)
+    if table is not None:
+        logger.warning("Table %s was in warehouse but not spark catalog", table_name)
+    return table
+
+
+def is_table_stored(table_name, database="default"):
+    return table_name in list_tables(database=database)
```
lixiliu, line 437: I assume spark_db and spark warehouse are two different things and this check whether it's stored in either (as opposed to `_is_table_stored_in_spark_db`) ?

daniel-thom: Based on the documentation you should be able to reuse tables across sessions, like this
```
$ pyspark
In [1]: spark.createDataFrame([{"a": 1}]).write.saveAsTable("my_table")
In [2]: exit()
$ pyspark
In [1]: spark.sql("select * from my_table")
```
It does work for me generally, but not always within dsgrid and I don't know why. I gave up debugging and just started checking in the warehouse directory manually. Refer to line 471

dsgrid/utils/spark.py
```diff
@@ -407,3 +408,102 @@ def _get_spark_session():
...
+    for table in _get_warehouse_path().iterdir():
+        if table.name == table_name:
+            spark = _get_spark_session()
+            return spark.read.parquet(str(table))
+    return None
+
+
+def _delete_table_if_present_but_not_stored(table_name):
+    if not _is_table_stored_in_spark_db(table_name):
+        for table in list(_get_warehouse_path().iterdir()):
```
lixiliu, line 488: I am confused. `_is_table_stored_in_spark_db` suggests that spark_db refers to `spark.catalog.listTables(dbName=database)`. But here it seems to be a different repo: `list(_get_warehouse_path().iterdir())`

dsgrid/utils/spark.py
```diff
@@ -407,3 +408,102 @@ def _get_spark_session():
...
+    if not _delete_table_if_present_but_not_stored(table_name) and _is_table_stored_in_spark_db(
+        table_name, database=database
+    ):
+        spark.sql(f"DROP TABLE {table_name}")
+        logger.info("Dropped table %s", table_name)
+
+
+# TODO
+# The next functions are workarounds for a problem where a new SparkSession does not load
+# tables saved in a previous session. Need to debug.
```
lixiliu, line 473: This alludes that we may not have a persisted catalog (spark_db, meta_store) within dsgrid

tests/data/standard_scenarios_registration.json
```diff
@@ -0,0 +1,33 @@
...
+      "register_project": true,
+      "datasets": [
+        {
+          "dataset_id": "conus_2022_reference_comstock",
+          "dataset_path": "../dsgrid-data/conus_2022_reference_comstock",
+          "config_file": "../dsgrid-project-StandardScenarios/dsgrid_project/datasets/modeled/comstock/dataset.toml",
+          "dimension_mapping_file": "../dsgrid-project-StandardScenarios/dsgrid_project/datasets/modeled/comstock/dimension_mappings.toml",
+          "register_dataset": true,
+          "submit_to_project": true,
+          "fix_dimension_uuids": false,
```
lixiliu, line 17: what does the fix refer to? Replacing the uuids between config files automatically after registration?

daniel-thom: Suppose a dataset uses a dimension created by the project. The dataset contributor has to put the dimension_id (with UUID) in the file as a dimension reference. If the project gets re-created, that UUID will be wrong. We have code that will update the UUIDs.

dsgrid/config/project_config.py
```diff
@@ -730,43 +732,142 @@ def _iter_datasets_by_status(self, status: DatasetRegistryStatus) -> InputDatase
...
+            logger.info("Start processing dimension=%s table_count=%s", dim, table_count)
             if table is None:
                 table = other
             else:
                 table = table.crossJoin(other)
-                assert table.count() == other.count() * table_count
+                count = table.count()
+                assert count == other.count() * table_count
+                table_count = count
+            other.unpersist()
```
lixiliu, line 811: What about unpersisting `other_dims[dim]`?

dsgrid/config/project_config.py
```diff
@@ -730,43 +732,142 @@ def _iter_datasets_by_status(self, status: DatasetRegistryStatus) -> InputDatase
...
+                if found:
+                    raise DSGInvalidDimensionAssociation(
+                        f"dimension association for {dim_type} has supplemental record ID "
+                        "{supplemental_id} defined in at least two supplemental dimensions"
+                    )
+                found = True
+                new_records = (
+                    table.join(
+                        mapping_records,
+                        on=table[dim_type.value] == mapping_records.to_id,
```
lixiliu, line 855: I think this is a one-to-many join, with one value `table[dim_type.value]` mappable to many values in `mapping_records.to_id`. Did the inner join produce  join as expected? And do you need to prefilter `mapping_records` before the inner join here?

lixiliu: What happens if you do table.left_join(mapping_records)? Could you get the new table directly without doing the union below?

lixiliu (commented):
okay, I'll stop now

PR #157, opened 2022-10-17
Refactor dimension association tables

daniel-thom:
This PR refactors dimension association tables in preparation for registering AEO growth rates.

- Store dimension requirements for each dataset in the project.toml.
- Cache the tables in the spark warehouse so that they don't have to be regenerated at every dataset submission.
- I also added a more generic way to register projects and datasets.


dsgrid/config/project_config.py
```diff
@@ -166,6 +232,11 @@ class InputDatasetModel(DSGBaseModel):
         updateable=False,
         # TODO: add notes about warnings for outdated versions? DSGRID-189.
     )
+    required_dimensions: RequiredDimensionsModel = Field(
```
daniel-thom, line 235: Would it be better to define these at a level higher than dataset so that the requirements can be shared across datasets of the same category?

.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "dsgrid-test-data"]
 	path = dsgrid-test-data
 	url = https://github.com/dsgrid/dsgrid-test-data
-	branch = main
+	branch = dt/dimension-association-refactor
```
lixiliu, line 4: switch back to main when ready to merge

dsgrid/config/project_config.py
```diff
@@ -730,46 +781,204 @@ def _iter_datasets_by_status(self, status: DatasetRegistryStatus) -> InputDatase
...
+            supp_dim_query_names,
+        )
+
+        # Project config construction asserts that there is no intersection of dimensions in
+        # multi and single.
+        existing = set()
+        if pivoted_dimension is not None:
+            existing.add(pivoted_dimension.value)
+        for df in multi_dfs:
+            existing.update(set(df.columns))
```
lixiliu, line 847: Is `existing` getting used later? Did I miss it?

dsgrid/config/project_config.py
```diff
@@ -730,46 +781,204 @@ def _iter_datasets_by_status(self, status: DatasetRegistryStatus) -> InputDatase
...
+                supp_dim_query_names,
+                dim_type,
+                supplemental_id,
+            )
+            record_ids.remove(supplemental_id)
+            record_ids.update(new_base_record_ids)
+
+        return record_ids
+
+    def _map_supplemental_record_to_base_records(
```
lixiliu, line 922: one record is singular the other is plural?

dsgrid/registry/project_registry_manager.py
```diff
@@ -372,20 +372,18 @@ def _register_all_in_one_dimensions(
...
             dt_str = dimension_type.value
-            dt_plural = f"all_{dt_str}s"
-            dt_plural_dash = f"all-{dt_str}s"
-            dt_plural_formal = f"All {dt_str.title()}s"
+            if dt_str.endswith("y"):
+                dt_str = dt_str[:-1]
+                suffix = "ies"
+            else:
+                suffix = "s"
+            dt_plural = f"all_{dt_str}{suffix}"
```
lixiliu, line 384: what if dt_str contains "space", "dash", or "underscore"?

dsgrid/tests/registration_models.py
```diff
@@ -0,0 +1,51 @@
+from pathlib import Path
+
+from pydantic import validator
+
+from dsgrid.data_models import DSGBaseModel
+from dsgrid.utils.files import load_data
+
+
+class DatasetRegistrationModel(DSGBaseModel):
```
lixiliu, line 9: Why is this in the `tests` folder?

lixiliu (commented):
Is there a companion PR in SS for the paradigm shift in the dimension association creation?

dsgrid/config/project_config.py
daniel-thom: line 851

dsgrid/config/project_config.py
daniel-thom: This function replaces one supplemental dimension record ID with its corresponding base dimension record IDs.

dsgrid/registry/project_registry_manager.py
daniel-thom: `dt_str` will be a `DimensionType` value, and so it shouldn't have a space or dash. I suppose a future dimension type could have an underscore. I don't think I follow your concern. Can you elaborate?

dsgrid/tests/registration_models.py
daniel-thom: I expect that this will only be used in test code or by dsgrid developers that are building registries for test purposes.

daniel-thom:
> Is there a companion PR in SS for the paradigm shift in the dimension association creation?

I have a local branch but want to wait until we merge this + the dsgrid-test-data PR. There is one pending design question.

mooneyme:
I added this to a team's chat with dan, but also posting here:

....

I like where this is going!

My main concern here is that with this refactor is centered around messaging of the required dimensions per dataset. At the moment, we are using supplemental dimensions to define required dimensions for a dataset, e.g.:

```
[datasets.required_dimensions]
[datasets.required_dimensions.single_dimensional]
subsector = [ "residential_subsectors",]
metric = [ "residential_end_uses",]
sector = [ "res",]
```

When we declare the dataset dimension requirements in this way, it can be misleading and it may seem like we literally are looking for one subsector called `"residential_subsectors"` from the resstock data. In reality, we want the dataset to have all the residential subsectors that the project's base dimension defines (or at least provide a mapping to the base dimension).

I want it to be clear that we are expecting a subset from project’s base dimension. How can we make that more clear and less confusing for a resstock modeler? Maybe we can have a specific naming convention for these kinds of supplemental dimensions? Maybe `"supplemental.residential_subsectors"` or something like that.

dsgrid/tests/registration_models.py
mooneyme: Can we add a comment at the top of expected use case for this script?

dsgrid/config/project_config.py
daniel-thom, line 235: Something that might be confusing about this implementation is that we use a supplemental dimension record ID to uniquely identify supplemental dimensions while in other places we use a dimension query name. There is a built-in assumption that there is only one record ID for these type of supplemental dimensions. It could be particularly confusing because the record ID is likely to be the same as the dimension query name.

An alternative solution is to have one instance of RequiredDimensionsModel for base dimension records and a second for supplemental dimension query names.

daniel-thom:
> I added this to a team's chat with dan, but also posting here:
> 
> ....
> 
> I like where this is going!
> 
> My main concern here is that with this refactor is centered around messaging of the required dimensions per dataset. At the moment, we are using supplemental dimensions to define required dimensions for a dataset, e.g.:
> 
> ```
> [datasets.required_dimensions]
> [datasets.required_dimensions.single_dimensional]
> subsector = [ "residential_subsectors",]
> metric = [ "residential_end_uses",]
> sector = [ "res",]
> ```
> 
> When we declare the dataset dimension requirements in this way, it can be misleading and it may seem like we literally are looking for one subsector called `"residential_subsectors"` from the resstock data. In reality, we want the dataset to have all the residential subsectors that the project's base dimension defines (or at least provide a mapping to the base dimension).
> 
> I want it to be clear that we are expecting a subset from project’s base dimension. How can we make that more clear and less confusing for a resstock modeler? Maybe we can have a specific naming convention for these kinds of supplemental dimensions? Maybe `"supplemental.residential_subsectors"` or something like that.

I added a note about an alternative solution to the companion dsgrid PR. If we want to get more specific about base vs supplemental, I think we should do it in the data structure definition rather than inventing text formats that need to be decoded.

dsgrid/config/dimension_association_manager.py
```diff
@@ -0,0 +1,93 @@
...
+    pivoted_dimension : DimensionType | None
+
+    Returns
+    -------
+    pyspark.sql.DataFrame | None
+
+    """
+    return try_load_stored_table(
+        _make_dimension_associations_table_name(project_id, dataset_id, pivoted_dimension)
+    )
```
elainethale, line 46: Is it correct that we used to have a project-wide association table, but now we only have dataset-specific association tables? I guess it is a simple matter to concatenate the dataset-specific association tables with the same pivoted dimension to get a project-wide association table?

dsgrid/registry/registry_manager_base.py
```diff
@@ -225,16 +227,20 @@ def _check_update(self, config, config_id, version):
         if version != cur_version:
             raise DSGInvalidParameter(f"version={version} is not current. Current={cur_version}")
 
-    def _update_config(self, config, submitter, update_type, log_message):
-        config_id = config.config_id
-        registry_config = self.get_registry_config(config_id)
-
+    @staticmethod
+    def get_next_version(version: VersionInfo, update_type: VersionUpdateType):
```
elainethale, line 231: I thought we got rid of VersionInfo? Do we just use it for a moment to calculate the new version string?

dsgrid/api/app.py
```diff
@@ -63,7 +63,9 @@
     raise Exception("The environment variable DSGRID_API_SERVER_STORE_DIR must be set.")
 offline_mode = True
 no_prompts = True
-spark = init_spark("dsgrid_api")
+# There could be collisions on the only-allowed SparkSession between the main process and
+# subprocesses that run queries.
+spark = init_spark("dsgrid_api", check_env=False)
```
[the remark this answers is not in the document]
daniel-thom: The default Hive metastore uses Apache Derby which only allows one user. Also, our standalone cluster on Eagle currently only allows one user (that could be improved if we are more particular about allocating cores). So, if the main process on the API server creates a session, receives a request to run a query, starts a child process which will also try to create a session, it will be deadlocked.

dsgrid/config/dimension_association_manager.py
daniel-thom: We used to build a project-wide table but then filter it by a data source before evaluating it, _I think_. I don't think we have a need for a project-wide table, but please correct me if I'm wrong.

dsgrid/registry/registry_manager_base.py
daniel-thom: We no longer use it in the Pydandic serialization/deserialization, but we still use it in the code.

dsgrid/config/project_config.py
```diff
@@ -137,6 +146,63 @@ def handle_dimension_union(cls, values):
...
+class RequiredDimensionRecordsModel(DSGBaseModel):
+
+    # data_source and time are excluded
+    geography: list[str] = []
+    metric: list[str] = []
+    model_year: list[str] = []
+    scenario: list[str] = []
+    sector: list[str] = []
+    subsector: list[str] = []
+    weather_year: list[str] = []
```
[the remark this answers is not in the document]
daniel-thom: any combination of base or supplemental dimension record IDs

dsgrid/config/project_config.py
daniel-thom: Ah, yes, query names do not exist yet, so this idea isn't great. I started off with record IDs because that is what is used for mapping between base and supplemental. As it happens, these supplemental record IDs and query names will be either be identical or very close. We are definitely setting up our users for some confusion.

dsgrid/config/project_config.py
```diff
@@ -166,6 +232,11 @@ class InputDatasetModel(DSGBaseModel):
         updateable=False,
         # TODO: add notes about warnings for outdated versions? DSGRID-189.
     )
+    required_dimensions: RequiredDimensionsModel = Field(
+        title="required_dimenions",
+        description="Defines required record IDs that must exist for each dimension.",
```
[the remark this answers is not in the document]
daniel-thom: No, any combination of base or supplemental dimension _record_ ID.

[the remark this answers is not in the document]
daniel-thom: Here is a specific example:

In this paradigm StandardScenarios would define this supplemental dimension:
```
name = "residential-subsectors"
display_name = "Residential Subsectors"
dimension_query_name = "residential_subsectors"
type = "subsector"
id = "residential-subsectors__656c0d22-c47e-4f95-a96e-1e31c17d8d16"
```
It would have one record:
```
id,name
residential_subsectors,Residential Subsectors
```
It would have these mapping records:
```
from_id,to_id
mobile_home,residential_subsectors
multifamily_2_to_4_units,residential_subsectors
multifamily_5_plus_units,residential_subsectors
single_family_attached,residential_subsectors
single_family_detached,residential_subsectors
```
`name`, `id`, `dimension_query_name`, `display_name`, and `record_id` are all very similar.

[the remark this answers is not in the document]
elainethale: Oh--that is weirder than I thought. I was thinking we could list the supplemental name, id, or query_name so that any base dimension record id that gets mapped to that supplemental dimension would be included.

Being able to reference any supplemental dimension record id also seems to set us up for name conflicts/ambiguities.

Per comments above, maybe we should use different data models for specifying requirements via (a) base dimension record ids versus (b) supplemental dimensions. The latter needs to identify the specific dimension (by name, id, or query name) and then could potentially also let users specify specific record ids.

dsgrid/config/dimension_association_manager.py
elainethale: I don't think we immediately need a project-wide table, and maybe it's never a good idea to construct the whole thing

dsgrid/config/project_config.py
daniel-thom: The default option is here: define one set of requirements per dataset. The downside is that two datasets cannot share the same requirements.

Other options:
1. Define requirements using a field that exists in each dataset config. `data_source`, `sector_description`, some new term. The field must be an enum.
2. Allow the project coordinator to define this free-form. The list of requirements would be at the top level of the project config. Each set of requirements would have a name. Let's suppose one is called `residential`.  Each dataset in `datasets` would have an optional field called `dimension_requirements`. ResStock and NEMS Residential would have `residential.` If the field is not provided then all project dimension records must be present.

elainethale: We discussed and decided to leave under dataset for now, in part because some cases that look like reuse might actually involve some tweaks.

elainethale: The revision distinguishes between base and supplemental dimensions so the latter can be clearly referenced (thus avoiding name conflict issues).

elainethale (approved):

dsgrid/registry/project_registry_manager.py
lixiliu: Not super important, just wondering if we also need to do dt_str.replace() according to the "space", "dash", "underscore" convention in the name.

daniel-thom: The only issue would be the possibility of a future dimension value containing an underscore. I don't think we have any rules about underscores in dimension/dimension_mapping record value IDs or names. Right?

mooneyme (approved):

PR #158, opened 2022-10-21
Fix bugs in conversion of ResStock and ComStock datasets to project

daniel-thom:


dsgrid/project.py
```diff
@@ -161,7 +162,8 @@ def process_query(self, context: QueryContext):
             for df in dfs[1:]:
                 if df.columns != main_df.columns:
                     raise Exception(
-                        f"BUG: dataset columns must match. main={main_df.columns} new={df.columns}"
+                        f"BUG: dataset columns must match. main={main_df.columns} new={df.columns} "
+                        f"diff={set(main_df.columns).symmetric_difference(df.columns)}"
```
lixiliu, line 166: Add `set()` to df.columns?

lixiliu: Interesting, I just realize `set().some_set_funct(list)` is the format used everywhere else too, and it works.

lixiliu (approved):
Makes sense to me, thanks!

PR #159, opened 2022-10-25
Cache mapped datasets during the query process

daniel-thom:
I'm opening this PR because I struggled with queries on Eagle with full ComStock and ResStock datasets. This introduces one main behavioral change: persist each project-mapped dataset to a file during a query. Save the location based on a hash of the dataset portion of the query. This will help us make incremental process. Suppose you're using Eagle and trying to get something done on the debug queue (1 hour limit). Each dataset takes 45 minutes. The query would never finish. With this change we'll finish in two attempts.

The change will be helpful beyond Eagle. We'll be able to avoid rebuilding mapped datasets. The overall queries will also be easier to monitor and debug.

I also added some supporting code, like adding the ability to customize the Spark config for each dataset.

### DIFF
--- a/dsgrid/query/models.py
+++ b/dsgrid/query/models.py
@@ -1,6 +1,5 @@
 import abc
 import enum
-from pathlib import Path
 from typing import Any, List, Optional, Set, Union
 
 import pyspark.sql.functions as F
@@ -11,7 +10,7 @@ from dsgrid.data_models import DSGBaseModel
 
 from dsgrid.dimension.base_models import DimensionType
 from dsgrid.dimension.dimension_filters import make_dimension_filter, DimensionFilterBaseModel
-from dsgrid.utils.files import compute_hash, load_data
+from dsgrid.utils.files import compute_hash
 
 
 class FilteredDatasetModel(DSGBaseModel):
@@ -172,16 +171,22 @@ class DatasetMetadataModel(DSGBaseModel):
     table_format_type: Optional[TableFormatType]
 
 
-class ProjectQueryParamsModel(DSGBaseModel):
-    """Defines how to transform a project into a CompositeDataset"""
+class CacheableQueryBaseModel(DSGBaseModel):
+    def serialize(self):
+        """Return a JSON representation of the model along with a hash that uniquely identifies it."""
+        text = self.json(indent=2)
+        return compute_hash(text.encode()), text
 
-    project_id: str = Field(description="Project ID for query")
-    dataset_ids: List[str] = Field(description="Dataset IDs from which to read")
-    excluded_dataset_ids: List[str] = Field(
-        description="Datasets to exclude from query", default=[]
-    )
-    # TODO: default needs to change
-    include_dsgrid_dataset_components: bool = Field(description="", default=False)
+
+class SparkConfByDataset(DSGBaseModel):
+
+    dataset_id: str
+    conf: dict[str, Any]
+
+
+class ProjectQueryDatasetParamsModel(CacheableQueryBaseModel):
+    """Parameters in a project query that only apply to datasets"""
+
     dimension_filters: List[Any] = Field(
         # Use Any here because we don't want Pydantic to try to discern the types.
         description="Filters to apply to all datasets",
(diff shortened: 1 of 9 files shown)


dsgrid/query/models.py
```diff
@@ -200,10 +205,56 @@ class ProjectQueryParamsModel(DSGBaseModel):
...
+    dataset_ids: List[str] = Field(description="Dataset IDs from which to read")
+    excluded_dataset_ids: List[str] = Field(
+        description="Datasets to exclude from query", default=[]
+    )
+    # TODO: default needs to change
+    include_dsgrid_dataset_components: bool = Field(description="", default=False)
+    dataset_params: ProjectQueryDatasetParamsModel = Field(
+        description="Parameters affecting datasets. Will be used for caching intermediate tables.",
+        default=ProjectQueryDatasetParamsModel(),
+    )
```
elainethale, line 249: Is this additional layer of nesting necessary/helpful? It seems like a lot of what was in ProjectQueryParamsModel is now in ProjectQueryDatasetParamsModel, but you haven't added a whole lot to ProjectQueryParamsModel, and ProjectQueryDatasetParamsModel doesn't appear to be per-dataset information (whereas the spark_conf_per_dataset is, but that's achieved by providing a list of objects that identify the dataset_id).

elainethale (commented):
I'm mostly wondering whether we need ProjectQueryDatasetParamsModel. Otherwise this looks good to me.

dsgrid/query/models.py
daniel-thom: I added ProjectQueryDatasetParamsModel to enable caching. The three parameters in that class uniquely identify a project-mapped dataset. I agree that it doesn't help the organizational structure.

dsgrid/query/models.py
```diff
@@ -200,10 +205,56 @@ class ProjectQueryParamsModel(DSGBaseModel):
...
+        return dimension_filters
+
+    @validator("per_dataset_aggregations")
+    def check_aggregations(cls, aggregations):
+        for agg in aggregations:
+            for dim_type in DimensionType:
+                field = dim_type.value
+                for item in getattr(agg.dimensions, field):
+                    # TODO: we could add support for this, but it does add some ambiguity.
+                    if item.function is not None:
```
lixiliu, line 223: So at the dataset level, function cannot be applied to a dimension?

dsgrid/query/models.py
```diff
@@ -200,10 +205,56 @@ class ProjectQueryParamsModel(DSGBaseModel):
...
+                        raise ValueError(
+                            f"function={item.function} cannot be set in ProjectQueryParamsModel"
```
lixiliu, line 225: I think this is for `ProjectQueryDatasetParamsModel`, same for error msg below. Still confused why we can't have column func here.

dsgrid/query/query_submitter.py
```diff
@@ -32,6 +32,12 @@ def __init__(self, output_dir: Path):
         self._cached_tables_dir().mkdir(exist_ok=True, parents=True)
         self._composite_datasets_dir().mkdir(exist_ok=True, parents=True)
 
+        # TODO: This location will need more consideration.
+        # We might want to store cached datasets in the spark-warehouse and let Spark manage it
+        # for us. However, would we share them on Eagle? What happens on Eagle walltime timeouts
+        # where the tables are left in intermediate states?
```
lixiliu, line 38: I guess we just have to delete tables if they cannot be reused.

dsgrid/utils/spark.py
```diff
@@ -351,19 +357,19 @@ def write_dataframe_and_auto_partition(df, filename, partition_size_mb=128, colu
     total_size = sum((x.stat().st_size for x in filename.glob("*.parquet")))
     desired = math.ceil(total_size / partition_size_bytes)
     actual = df.rdd.getNumPartitions()
-    if actual > desired:
+    if abs(actual - desired) / min(actual, desired) * 100 < rtol_pct:
```
lixiliu, line 360: A bit confused by the min() func here, because it anchors around the smaller value. This is the same as saying, do nothing if partitions falls within 50%-150% of the smaller of actual vs. desired.

dsgrid/utils/spark.py
```diff
@@ -328,6 +331,9 @@ def write_dataframe_and_auto_partition(df, filename, partition_size_mb=128, colu
         Target size in MB for each partition
     columns : None, list
         If not None and repartitioning is needed, partition on these columns.
+    rtol_pct : int
+        Don't repartition or coalesce if the relative difference between target and desired
```
lixiliu, line 335: I think target and desired mean the same thing here?

dsgrid/utils/spark.py
```diff
@@ -400,6 +406,30 @@ def sql_from_sqlalchemy(query):
...
+    orig_settings = {}
+
+    try:
+        for key, val in conf.items():
+            orig_settings[key] = spark.conf.get(key)
+            spark.conf.set(key, val)
+            logger.info("Set %s=%s while running", key, val)
+        yield
+    finally:
+        for key, val in orig_settings.items():
```
lixiliu, line 429: Why is the orig setting being set back here? If try fails, orig_settings will not be complete?

dsgrid/query/models.py
daniel-thom: This is not new. I just moved the code block. We can revisit apart from this PR if necessary.

dsgrid/query/query_submitter.py
daniel-thom: Spark does not handle the partial tables, so we would have to catch the specific error and delete those tables in the warehouse.

dsgrid/utils/spark.py
daniel-thom: Yeah, typo

dsgrid/utils/spark.py
daniel-thom: I chose this because I think we care more about the relative difference than absolute. If anything, perhaps we should anchor on the larger value because it will be more lenient. But I'm open if there are standard ways of expressing this.

dsgrid/utils/spark.py
daniel-thom: Cases:
1. Suppose the first new setting is correct but the second one is incorrect, and an exception is raised. We need to revert the first change.
2. The code block wrapped by this decorator will run when `yield` is called. We need to revert all changes afterwards, regardless of whether it completes successfully or raises an exception.

dsgrid/utils/spark.py
daniel-thom: I agree that you are correct and fixed it.

elainethale (approved):

PR #160, opened 2022-11-02
Fix mapping of pivoted columns

daniel-thom:
This mapping code did not work when one from_id column was mapped to multiple to_id columns.


dsgrid/config/dimension_mapping_base.py
```diff
@@ -103,7 +103,7 @@ class DimensionMappingArchetype(DSGEnum):
         check_fraction_sum_eq1_to_id=True,
     )
     MANY_TO_MANY_MAP_FRACTION_SUM_EQ1_TO_ID = EnumValue(
-        value="many_to_many_map_fraction_sum_eq1",
+        value="many_to_many_map_fraction_sum_to_id_eq1",
```
lixiliu, line 106: yep, nice catch

lixiliu (approved):
Makes sense. This is a bug so obvious now in hindsight

PR #162, opened 2022-11-03
Auto-generate dataset mappings from project supplemental dimensions

daniel-thom:


dsgrid/registry/project_registry_manager.py
```diff
@@ -642,8 +616,160 @@ def _submit_dataset_and_register_mappings(
                     raise DSGValueNotRegistered(f"mapping_id={ref.mapping_id}")
                 references.append(ref)
 
+        references += self._auto_register_reverse_supplemental_mappings(
+            project_config, dataset_config, references, submitter, log_message, context
+        )
         self._submit_dataset(project_config, dataset_config, submitter, log_message, references)
 
+    def _register_mappings_from_file(
```
daniel-thom, line 624: This function has logic that was moved without change.

dsgrid/registry/project_registry_manager.py
```diff
@@ -642,8 +616,160 @@ def _submit_dataset_and_register_mappings(
...
+                # else the dataset may only need to provide a subset of records, and those are
+                # checked in the dimension association table.
+
+        new_mappings = []
+        for from_id, from_version in needs_mapping:
+            from_dim = self._dimension_mgr.get_by_id(from_id, version=from_version)
+            to_dim, to_version = project_config.get_base_dimension_and_version(
+                from_dim.model.dimension_type
+            )
+            # TODO: Our dimension mappings don't store versions of the underlying dimensions.
```
daniel-thom, line 706: We may need to discuss this point.

dsgrid/registry/project_registry_manager.py
```diff
@@ -382,7 +386,7 @@ def _register_all_in_one_dimensions(
             else:
                 suffix = "s"
             dt_plural = f"all_{dt_str}{suffix}"
-            dt_plural_dash = f"all-{dt_str}{suffix}"
+            dt_plural_dash = f"all-{model.project_id}-{dt_str}{suffix}"
```
daniel-thom, line 389: This is an unrelated change. I noticed that we had a potential problem. Every project was going to have all-in-one dimensions with very similar names and IDs. This should make it more clear.

dsgrid/registry/project_registry_manager.py
```diff
@@ -642,8 +616,158 @@ def _submit_dataset_and_register_mappings(
...
+                    from_dimension_type=mapping_config.model.from_dimension.dimension_type,
+                    to_dimension_type=mapping_config.model.to_dimension.dimension_type,
+                    mapping_id=mapping_id,
+                    version=str(self._dimension_mapping_mgr.get_current_version(mapping_id)),
+                )
+            )
+
+        return references
+
+    def _auto_register_reverse_supplemental_mappings(
```
elainethale, line 672: I agree with @lixiliu that this should be done based on a flag rather than completely auto. Not sure if that suggests a name change here or not.

elainethale (approved):

dsgrid/cli/registry.py
```diff
@@ -449,6 +449,16 @@ def register_project(
     "Use it when the mappings are already registered.",
     callback=_path_callback,
 )
+@click.option(
+    "-a",
+    "--autogen-reverse-supplemental-mappings",
```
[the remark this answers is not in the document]
daniel-thom: I don't think we've discussed a convention. The code makes more sense when the parameter is plural. `--autogen-reverse-supplemental-mappings` becomes `autogen_reverse_supplemental_mappings`. That's probably less important than the user interface. I think I could make an argument for either of these:
```
dsgrid registry submit-dataset \
    --autogen-reverse-supplemental-mappings geography \
    --autogen-reverse-supplemental-mappings subsector
```
```
dsgrid registry submit-dataset \
    --autogen-reverse-supplemental-mapping geography \
    --autogen-reverse-supplemental-mapping subsector
```
I tend to prefer the plural because it reminds the user that they can pass the option multiple times.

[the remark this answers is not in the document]
elainethale: I get it now--the option collects them all, so it's collecting mapping*s*. That's fine with me.

dsgrid/registry/dimension_mapping_registry_manager.py
```diff
@@ -546,3 +546,18 @@ def show(
         if return_table:
             return table
         display_table(table)
+
+    def try_get_mapping(self, from_dimension, from_version, to_dimension, to_version):
+        """Return the mapping config and version for from and to dimension, if one exists."""
+        from_id = from_dimension.model.dimension_id
```
lixiliu, line 552: I think from_dimension and to_dimension here are configs? Shouldn't they already contain version info?

dsgrid/registry/project_registry_manager.py
```diff
@@ -642,8 +623,175 @@ def _submit_dataset_and_register_mappings(
...
+            x.dimension_id for x in p_model.dimensions.supplemental_dimension_references
+        }
+        d_dim_from_ids = set()
+        for ref in mapping_references:
+            mapping_config = self._dimension_mapping_mgr.get_by_id(ref.mapping_id)
+            d_dim_from_ids.add(mapping_config.model.from_dimension.dimension_id)
+
+        needs_mapping = []
+        for dim in dataset_config.model.dimension_references:
+            if dim.dimension_id not in p_base_dim_ids and dim.dimension_id not in d_dim_from_ids:
```
lixiliu, line 709: This feels a bit awkward. It might be better instead to check that id is a supplemental dimension and exists as a to_dimension in a mapping:
```
if (dim.dimension_id in p_supp_dim_ids and dim.dimension_id in d_dim_to_ids and dim.dimension_type in autogen_reverse_supplemental_mappings):
```

dsgrid/registry/project_registry_manager.py
```diff
@@ -642,8 +623,175 @@ def _submit_dataset_and_register_mappings(
...
+                if (
+                    dim.dimension_id in p_supp_dim_ids
+                    and dim.dimension_type in autogen_reverse_supplemental_mappings
+                ):
+                    needs_mapping.append((dim.dimension_id, dim.version))
+                # else the dataset may only need to provide a subset of records, and those are
+                # checked in the dimension association table.
+
+        if len(needs_mapping) != len(autogen_reverse_supplemental_mappings):
```
lixiliu, line 718: Is it better to show which dimension can not auto-generate reverse mapping instead?

dsgrid/registry/project_registry_manager.py
```diff
@@ -642,8 +623,175 @@ def _submit_dataset_and_register_mappings(
...
+        for from_id, from_version in needs_mapping:
+            from_dim = self._dimension_mgr.get_by_id(from_id, version=from_version)
+            to_dim, to_version = project_config.get_base_dimension_and_version(
+                from_dim.model.dimension_type
+            )
+            mapping, version = self._dimension_mapping_mgr.try_get_mapping(
+                from_dim, from_version, to_dim, to_version
+            )
+            if mapping is None:
+                p_mapping, _ = self._dimension_mapping_mgr.try_get_mapping(
```
lixiliu, line 734: not following this attempt

dsgrid/registry/project_registry_manager.py
```diff
@@ -642,8 +623,175 @@ def _submit_dataset_and_register_mappings(
...
+                dst = Path(tempfile.gettempdir()) / f"reverse_{p_mapping.config_id}.csv"
+                # Use pandas because spark creates a CSV directory.
+                reverse_records.to_csv(dst, index=False)
+                dimension_type = from_dim.model.dimension_type.value
+                new_mappings.append(
+                    {
+                        "description": f"Maps {dataset_config.config_id} {dimension_type} to project",
+                        "dimension_type": dimension_type,
+                        "file": str(dst),
+                        "mapping_type": "duplication",
```
lixiliu, line 762: Why would this be hard-coded?

dsgrid/registry/dimension_mapping_registry_manager.py
daniel-thom: Dimension configs and models do not contain versions.

dsgrid/registry/project_registry_manager.py
lixiliu, line 718: In cases where one project had successfully registered a reverse mapping, and another dataset could reuse that reverse mapping, in theory, user cannot use the --autogen flag anymore and need to instead reference that mapping, right?

dsgrid/registry/project_registry_manager.py
daniel-thom: I'm not sure I follow the objection. This code
1. Checks that the ID is a project supplemental (which means that there is a reverse mapping to the project base).
2. Checks that the dataset did not already provide a mapping.
3. Checks that the user wants to autogen a mapping.
4. Ignores the case where a missing mapping is OK because of dimension association requirements.

Where would `d_dim_to_ids` come from? `mapping_references`? The dataset hasn't included this dimension in the references.

dsgrid/registry/dimension_mapping_registry_manager.py
```diff
@@ -546,3 +546,18 @@ def show(
         if return_table:
             return table
         display_table(table)
+
+    def try_get_mapping(self, from_dimension, from_version, to_dimension, to_version):
```
lixiliu, line 550: bug

dsgrid/registry/project_registry_manager.py
lixiliu: If default, maybe use:  DimensionMappingType.MANY_TO_MANY_EXPLICIT_MULTIPLIERS

dsgrid/registry/project_registry_manager.py
lixiliu: I follow now

lixiliu (approved):
Per discussed, update try_get_mappings to only loop through project reference mappings and update default mapping type for reversed mappings.

Thanks Dan!

lixiliu (approved):
Thanks, looks good!

PR #163, opened 2022-11-08
Add filtering of one-table format

daniel-thom:


dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -278,8 +305,16 @@ def get_time_zone_mapping(self):
         """
 
     @track_timing(timer_stats_collector)
-    def _convert_time_dimension(self, load_data_df, time_zone_mapping):
+    def _convert_time_dimension(self, load_data_df):
```
lixiliu, line 308: I wonder if we should keep `_convert_time_dimension(self, load_data_df, time_zone_mapping)` and have time_zone_mapping retrieval be a different func instead because, as you pointed out yesterday, the mapping could come from project level geography.

dsgrid/config/project_config.py
```diff
@@ -133,14 +133,22 @@ def check_files(cls, values: dict) -> dict:
...
             "dimension record name",
         )
         return values
 
+    @validator("base_dimensions")
+    def check_time_zone(cls, values: list) -> list:
+        """Validate dimension names are unique across all dimensions."""
+        for dimension in values:
+            if dimension.dimension_type == DimensionType.GEOGRAPHY:
+                check_timezone_in_geography(dimension)
```
elainethale, line 149: Correct--project geography must list a time zone

dsgrid/dataset/dataset_schema_handler_base.py
elainethale: This point makes sense to me. Relatedly, is it always clear how to get_time_zone_mapping from the dataset? 

I wonder if there is or should be a project-specific way to do this that is distinct from the dataset-specific way?

dsgrid/dimension/base_models.py
```diff
@@ -136,33 +136,34 @@ def check_required_dimensions(dimensions, tag):
...
     if dimension.dimension_type != DimensionType.GEOGRAPHY:
         raise DSGInvalidDimension(
             f"Dimension has type {dimension.dimension_type}, "
             "Can only check timezone for Geography."
         )
 
-    tz = set(TimeZone) - TimeZone.NONE
-    record_tz = {rec.time_zone for rec in dimension.records}
+    if not hasattr(dimension.records[0], "time_zone"):
+        raise ValueError("These geography dimension records must include a 'time_zone' column.")
```
elainethale, line 160: nit: It might be helpful to say why these geography dimension records must include a 'time_zone' column. Perhaps the caller can pass a message string?

dsgrid/dimension/base_models.py
```diff
@@ -136,33 +136,34 @@ def check_required_dimensions(dimensions, tag):
...
 
+    tz = set(TimeZone)
+    record_tz = {rec.time_zone for rec in dimension.records}
     diff = record_tz.difference(tz)
     if diff:
         raise DSGInvalidDimension(
-            f"Base Geography dimension {dimension.display} has invalid timezone(s) in "
-            f"records: {dimension.filename}. Use dsgrid TimeZone enum values only."
+            f"Geography dimension {dimension.dimension_id} has invalid timezone(s) in records: "
+            f"{dimension.filename}: {diff}. Use dsgrid TimeZone enum values only."
```
elainethale, line 168: Print the TimeZone enum values here or provide a link to the code/documentation?

elainethale (commented):
Should some tests be added for the new functionality (filtering single-table schema datasets, checking for time_zone in geography)?

dsgrid/utils/dataset.py
```diff
@@ -108,6 +108,48 @@ def map_and_reduce_pivoted_dimension(df, records, pivoted_columns, operation, re
     return df.selectExpr(*nonvalue_cols, *extra_cols, *exprs), sorted(final_columns), dropped
 
 
+def make_time_zone_mapping(load_data_df, geography_dim):
```
daniel-thom, line 111: @lixiliu This function is a modification of older code and is not used. I'm inclined to delete it, and so let me know if you want to keep it.

dsgrid/config/dimensions.py
```diff
@@ -494,6 +502,9 @@ def check_times(cls, ranges, values):
             return ranges
         return _check_time_ranges(ranges, values["str_format"], timedelta(days=365))
 
+    def does_geography_require_time_zone(self):
+        return False
```
lixiliu, line 506: May be unrelated here, what will we ever use TimeZone.NONE for?

dsgrid/config/project_config.py
```diff
@@ -133,14 +133,25 @@ def check_files(cls, values: dict) -> dict:
...
         """Validate dimension names are unique across all dimensions."""
         check_uniqueness(
             [dim.name for dim in values],
             "dimension record name",
         )
         return values
 
+    @validator("base_dimensions")
+    def check_time_zone(cls, values: list) -> list:
+        """Validate dimension names are unique across all dimensions."""
```
lixiliu, line 146: stale doc str

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -260,29 +290,24 @@ def _remap_dimension_columns(self, df, pivoted_columns=None):
...
-        """Get time_zone mapping to map to load_data_df
-        Returns
-        -------
-        pyspark.sql.DataFrame
-            a two-column df containing time_zone and a mapping key column
-        """
+    def _convert_time_before_project_mapping(self):
+        time_dim = self._config.get_dimension(DimensionType.TIME)
+        return (
+            time_dim.model.does_geography_require_time_zone()
```
lixiliu, line 296: NIT: `does_geography_require_time_zone` is a question but `use_project_geography_time_zone` is an affirmative field for TRUE

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -260,29 +290,24 @@ def _remap_dimension_columns(self, df, pivoted_columns=None):
...
+            and not self._config.model.use_project_geography_time_zone
+        )
 
     @track_timing(timer_stats_collector)
-    def _convert_time_dimension(self, load_data_df, time_zone_mapping):
+    def _convert_time_dimension(self, load_data_df, project_config):
         time_dim = self._config.get_dimension(DimensionType.TIME)
-        load_data_df = time_dim.convert_dataframe(
+        if time_dim.model.does_geography_require_time_zone():
```
lixiliu, line 303: This is awkward for me because the checks exist both here and in _convert_time_before_project_mapping(). But I am following here, I think the better solution is to remove _convert_time_before_project_mapping() later.

dsgrid/dataset/dataset_schema_handler_one_table.py
```diff
@@ -102,12 +107,55 @@ def get_unique_dimension_rows(self):
...
+        path = Path(self._config.load_data_path)
+        if path.suffix == ".csv":
+            # write_dataframe_and_auto_partition doesn't support CSV yet
+            overwrite_dataframe_file(path, load_df)
+        else:
+            write_dataframe_and_auto_partition(load_df, path)
+        logger.info("Rewrote simplified %s", self._config.load_data_path)
+
+    def make_project_dataframe(self, project_config):
+        if self._convert_time_before_other_dimensions():
```
lixiliu, line 142: I don't think the if statement is required here.

dsgrid/dataset/dataset_schema_handler_one_table.py
```diff
@@ -102,12 +107,55 @@ def get_unique_dimension_rows(self):
...
+
+        ld_df = self._remap_dimension_columns(
+            ld_df,
+            # Some pivot columns may have been removed.
+            pivoted_columns=set(self._load_data.columns).intersection(
+                self.get_pivoted_dimension_columns()
+            ),
+        )
+
+        if not self._convert_time_before_other_dimensions():
```
lixiliu, line 155: Same here, I don't think the if not statement is required.

dsgrid/utils/dataset.py
lixiliu: sure go for it

lixiliu (approved):
Looks good to me for the most part, some checks may be redundant.

dsgrid/dataset/dataset_schema_handler_base.py
daniel-thom: 1. The checks are conflicting. `_convert_time_before_project_mapping is dependent on `config.model.use_project_geography_time_zone` but this function is not.
2. This function will always call `time_dim.convert_dataframe()`, and we don't want that called twice in one pass..

dsgrid/config/dimensions.py
daniel-thom: I don't know.

dsgrid/config/dimensions.py
```diff
@@ -457,7 +456,7 @@ def check_times(cls, ranges, values):
         return _check_time_ranges(ranges, values["str_format"], values["frequency"])
 
     def is_time_zone_required_in_geography(self):
-        return self.timezone == TimeZone.LOCAL
+        return False
```
lixiliu, line 459: `return self.timezone == TimeZone.NONE`

If we are going down this path, fix `check_timezone()`

lixiliu (approved):

PR #165, opened 2022-11-14
Fix Spark issues when verifying dimension association tables

daniel-thom:
We sometimes encounter issues when submitting a dataset to a project. Too many executors get involved in validating dimension associations and the process either takes too long or encounters memory errors. This PR attempts to fix the issue by restarting the spark session with configuration parameters that limit the executors to 1 - but only if Spark dynamic allocation is enabled.

I also had to change the way we cache dimension and mapping record dataframes so that we don't lose that benefit.

### DIFF
--- a/dsgrid/loggers.py
+++ b/dsgrid/loggers.py
@@ -3,6 +3,7 @@
 import logging
 import logging.config
 import os
+from contextlib import contextmanager
 
 
 # ETH@20210325 - What if you want to set up logging for all loggers, or for all
@@ -98,3 +99,23 @@ def check_log_file_size(filename, limit_mb=10, no_prompts=False):
         val = input(msg)
         if val == "" or val.lower() == "y":
             os.remove(filename)
+
+
+@contextmanager
+def disable_console_logging(name="dsgrid"):
+    logger = logging.getLogger(name)
+    if logger.handlers:
+        console_level = None
+        try:
+            for handler in logger.handlers:
+                if handler.name == "console":
+                    console_level = handler.level
+                    handler.setLevel(logging.FATAL)
+                    break
+            yield
+        finally:
+            for handler in logger.handlers:
+                if handler.name == "console":
+                    assert console_level is not None
+                    handler.setLevel(console_level)
+                    break
--- a/dsgrid/loggers.py
+++ b/dsgrid/loggers.py
@@ -104,18 +104,17 @@ def check_log_file_size(filename, limit_mb=10, no_prompts=False):
 @contextmanager
 def disable_console_logging(name="dsgrid"):
     logger = logging.getLogger(name)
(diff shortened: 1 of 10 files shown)


dsgrid/loggers.py
```diff
@@ -98,3 +99,22 @@ def check_log_file_size(filename, limit_mb=10, no_prompts=False):
...
+        for handler in logger.handlers:
+            if handler.name == "console":
+                console_level = handler.level
+                handler.setLevel(logging.FATAL)
+                break
+        yield
+    finally:
+        for handler in logger.handlers:
+            if handler.name == "console":
+                assert console_level is not None
```
elainethale, line 118: nit: How do you know/expect console_level to not be None?

dsgrid/registry/project_registry_manager.py
```diff
@@ -823,11 +827,22 @@ def _submit_dataset(
...
+                "spark.dynamicAllocation.shuffleTracking.enabled": True,
+                "spark.dynamicAllocation.maxExecutors": 1,
+                "spark.executor.cores": 1,
+            }
+            with restart_spark_with_custom_conf(conf):
+                self._check_dataset_base_to_project_base_mappings(
+                    project_config,
+                    dataset_config,
+                    mapping_references,
+                )
```
elainethale, line 845: What happens when this exits? Does it somehow go back to the previous spark config?

dsgrid/utils/spark.py
```diff
@@ -108,8 +146,8 @@ def read_dataframe(filename, cache=False, require_unique=None, read_with_spark=T
     ----------
     filename : str | Path
         path to file
-    cache : bool
-        If True, cache the DataFrame in memory.
+    table_name : str | None
+        If set, cache the DataFrame in memory. Must be unique.
```
elainethale, line 150: read the cached DataFrame from memory?

dsgrid/registry/project_registry_manager.py
elainethale: I see now that it does. I wasn't reading function names carefully enough ...

elainethale (approved):
All comments are minor, I believe

dsgrid/loggers.py
daniel-thom: Unless something crazy happened, it is the same for loop as the first one and `console_level` should have been set.

dsgrid/utils/spark.py
daniel-thom: This function will always read from the file on disk. The table_name can set to read and then cache. Presumably, the user will use that name rather than calling this function again.

lixiliu (approved):
can't say I understood everything code-wise, but based on our conversation, I get the intention here.

Issue #170, opened 2022-11-29
Illegal env var __DSGRID_SKIP_SAVING_DIMENSION_ASSOCIATIONS__


lixiliu:
Recent feature:
```
    # Saving these tables during tests doesn't add any value and adds complications whenever
    # multiple processes try to access the Hive metastore, particularly when we inject errors.
    # We only need this when we submit datasets to projects and users want to inspect the
    # outcomes. This feature may become opt-in if we encounter similar problems.
    os.environ["__DSGRID_SKIP_SAVING_DIMENSION_ASSOCIATIONS__"] = "1"
```
is causing the following tests_aws-related pytests to fail currently:
```
tests_aws/test_registry/test_aws_registry_workflow_online_mode.py::test_aws_registry_workflow_online_mode FAILED                                    [ 89%]
tests_aws/test_registry/test_data_sync.py::test_data_sync_project_id FAILED                                                                         [ 90%]
tests_aws/test_registry/test_data_sync.py::test_data_sync_dataset_id FAILED                                                                         [ 91%]
tests_aws/test_registry/test_data_sync.py::test_data_sync_project_id_and_dataset_id FAILED                                                          [ 92%]
tests_aws/test_registry/test_data_sync.py::test_data_sync_bad_project_id FAILED                                                                     [ 93%]
tests_aws/test_registry/test_data_sync.py::test_data_sync_bad_dataset_id FAILED                                                                     [ 94%]
tests_aws/test_registry/test_data_sync.py::test_data_sync_project_id_and_bad_dataset_id FAILED                                                      [ 95%]
tests_aws/test_registry/test_data_sync.py::test_data_sync_project_id2 FAILED                                                                        [ 96%]
tests_aws/test_registry/test_data_sync.py::test_data_sync_bad_project_id_with_dataset_lock FAILED                                                   [ 97%]
tests_aws/test_registry/test_push_pull_online_mode.py::test_pull_delete_local PASSED                                                                [ 98%]
tests_aws/test_registry/test_push_pull_online_mode.py::test_pull_exclude PASSED                                                                     [ 99%]
tests_aws/test_registry/test_push_pull_online_mode.py::test_push_exclude PASSED                                                                     [100%]
```
with the following error:
```
E               Exception: Internal environment variables are not allowed to be set in online mode: ['__DSGRID_SKIP_SAVING_DIMENSION_ASSOCIATIONS__']

dsgrid/registry/registry_manager.py:553: Exception
```

PR #171, opened 2022-11-29
Make skip-check environment variables more explicit

daniel-thom:
Closes GitHub Issue #170 

## Description
The code was checking for any internal environment variable being set while in online mode. That was too restrictive. We should only throw errors when those internal environments cause checks to be skipped.


lixiliu (approved):
Looks good, but need to run local pytests first to make sure it's working.

PR #172, opened 2022-12-01
Use pytest tmp_path fixture

daniel-thom:
## Description
I noticed that pytest provides a fixture to manage temporary directories. This PR removes the temporary directories that we were creating in our tests. No functional changes.

### DIFF
--- a/tests/test_filter_registry.py
+++ b/tests/test_filter_registry.py
@@ -1,7 +1,3 @@
-import shutil
-import tempfile
-from pathlib import Path
-
 from dsgrid.config.simple_models import RegistrySimpleModel
 from dsgrid.dimension.base_models import DimensionType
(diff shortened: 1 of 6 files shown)


tests/test_filter_registry.py
```diff
@@ -40,45 +36,42 @@
 }
 
 
-def test_filter_registry():
+def test_filter_registry(tmp_path):
```
lixiliu, line 39: so this works because of the keyword `tmp_path`?

daniel-thom: `tmp_path` is a built-in pytest fixture.
https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html

lixiliu: cool!

lixiliu (approved):
pytest passed for me

PR #173, opened 2022-12-05
Check for case where datasets have invalid columns

daniel-thom:
## Description

This fixes a bug where the code was not checking for the case where a dataset has pivoted
columns that are not defined in the project's dimension and not present
in the dataset dimension mapping record file. The bug resulted mapped dataframes with
invalid columns.

### DIFF
--- a/dsgrid/dataset/dataset_schema_handler_base.py
+++ b/dsgrid/dataset/dataset_schema_handler_base.py
@@ -112,15 +112,15 @@ class DatasetSchemaHandlerBase(abc.ABC):
         return sorted(list(self._config.get_dimension(dim_type).get_unique_ids()))
 
     @track_timing(timer_stats_collector)
-    def get_pivoted_dimension_columns_mapped_to_project(self):
+    def get_pivoted_dimension_columns_mapped_to_project(self) -> set[str]:
         """Get columns for the dimension that is pivoted in load_data and remap them to the
-        project's record names. The returned dict will not include columns that the project does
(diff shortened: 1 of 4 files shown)


dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -130,18 +130,22 @@ def get_pivoted_dimension_columns_mapped_to_project(self):
...
+                from_ids = get_unique_values(records, "from_id")
+                to_ids = get_unique_values(
+                    records.select("to_id").filter("to_id IS NOT NULL"), "to_id"
+                )
                 diff = from_ids.difference(columns)
                 if diff:
                     raise DSGInvalidDataset(
                         f"Dimension_mapping={mapping_config.config_id} has more from_id records than the dataset pivoted {dim_type.value} dimension: {diff}"
                     )
+                diff = from_ids.symmetric_difference(self.get_pivoted_dimension_columns())
```
lixiliu, line 143: this should be `diff = columns.difference(from_ids)`, `(from_ids).difference(columns)` is already being checked above

lixiliu (approved):
Minor comment

dsgrid/dataset/dataset_schema_handler_base.py
daniel-thom: Right. I changed it to be a single check. We don't need both. If needed, I could add a print of the difference going in each direction.

lixiliu (approved):

PR #175, opened 2022-12-06
Remove unused files

daniel-thom:
I don't think we need these files any longer.


elainethale (approved):
You know better than me what isn't used any more. If these aren't used, let's delete.

PR #176, opened 2022-12-08
Add indentation to JSON5 config files

daniel-thom:


dsgrid/config/config_base.py
```diff
@@ -195,7 +196,7 @@ def serialize(self, path, force=False):
                 new_files.append(Path(orig_file).name)
             model_data[field] = new_files
 
-        dump_data(model_data, dst_config_file)
+        dst_config_file.write_text(json5.dumps(model_data, indent=2))
```
elainethale, line 199: I assume you thought about making this a helper in dsgrid.utils.files and decided against? Would we sometimes want to use a different indent level?

elainethale (approved):
I'm a little surprised you didn't write a new helper function to standardize the indent level we use for json5

dsgrid/config/config_base.py
daniel-thom: I did. We have `indent=2` spread throughout the code. In other places we pass the indentation to the Pydantic BaseModel.json() call. We should really be doing that in these places also. I think we aren't because of custom JSON encoders, and there are some TODOs for that.

The only time we don't want indentation is in the API layer where we are sending JSON text over the network. At best we could define a constant for DEFAULT_JSON_INDENTATION.

PR #184, opened 2022-12-19
Put restriction on numpy to fix pyspark

daniel-thom:
All CI runs in PRs will fail until this is merged. Pyspark is using a numpy type that has been deleted.


lixiliu (approved):
Yikes

PR #214, opened 2023-01-18
Remove data_source as a project dimension

daniel-thom:
This PR is the outcome of today's conversation. I don't like it very much. I think a better approach would be to remove `DATA_SOURCE` as a member of `DimensionType`. `data_source` is already a field in `DatasetConfigModel`, which may meet all requirements.

### DIFF
--- a/dsgrid/config/project_config.py
+++ b/dsgrid/config/project_config.py
@@ -12,7 +12,12 @@ from dsgrid.config.dimension_association_manager import (
     save_dimension_associations,
 )
 from dsgrid.data_models import DSGBaseModel, serialize_model_data
-from dsgrid.dimension.base_models import check_required_dimensions, check_timezone_in_geography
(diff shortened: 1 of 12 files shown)


dsgrid/config/project_config.py
```diff
@@ -41,7 +46,6 @@
 from .mapping_tables import MappingTableByNameModel
 from .dimensions import (
     DimensionReferenceModel,
-    DimensionType,
```
lixiliu, line 44: This is simply an error right?

lixiliu (approved):
Looks fine if it's the direction we want to head down

dsgrid/config/project_config.py
daniel-thom: Bad style but not an error.

daniel-thom:
Closing because we will remove data_source from DimensionType.

PR #215, opened 2023-01-19
Automate creation of derived dataset configs

daniel-thom:
In the existing two-table format the load_data_lookup table designates expected-missing dimension records by specifying id=null. This information was getting dropped when the dataset was mapped to the project. This PR adds that information to the mapped dataset by adding a row where the timestamp and all value columns are null.

Note that I also registered comstock_projected and resstock_projected in the test data registry.

There are more behavioral changes in this PR. Refer to the questions and answers below.

**What is a supplemental dimension?**
1. [Existing] It provides a way to change resolution – aggregate/disaggregate, like change county to state.
2. [Existing] It provides a shorthand specifier for a common subset of dimensions, like “all electricity” out of all possible end uses.
3. [New] (2) is split into two types: collapsed and not collapsed. Sometimes the user wants all electricity columns and other times they want those columns aggregated. The convention is to append “Collapsed” to the dimension display name. Nothing about the dimension specifies this attribute.

**How does a user specify that they want all electricity columns to show up in the final table?**
[Old] Specify a supplemental dimension filter with dimension_type=metric and dimension_query_name=electricity.
[New] Specify an aggregration with dimension_type=metric, dimension_query_name=electricity, and operation=anything (doesn’t matter; it’s a no-op _<-- this depends on how the other dimensions are specified. E.g., if one dimension is omitted or different from base resolution, aggregation is happening._). Elaine found the behavior around filtering supplemental dimensions confusing, so it’s possible that we could delete it entirely. If the user wants, “show me Colorado or PCA=13 data only, but keep county resolution,” filtering by supplemental dimension is probably the way to go. In order to delete that feature we would either need to add more dimension record columns or keep adding more and more supplemental dimensions for selection. Please direct any questions to Elaine; these operations are filters to me, not aggregations. 

**Does the value column change when the user performs an aggregation?**
[Old] Yes, the aggregation operation name was appended to the dimension record ID. For example, “all_electricity” became “all_electricity_sum” or “all_electricity_max.”
[New] No. I found that adding the operation to the no-op aggregation case was nonsensical. We can’t tell when the operation will be a no-op. This would be easy to change back if you disagree. Or, we could discuss alternatives.

**Can the user customize the column names of the table produced by a query?**
[Old] No. The column names were always the dimension query names.
[New] Yes. Now the user can specify whether they want dimension query names (default) or dimension types. If the user wants to register this table as a derived dataset then they must specify dimension types (or manually convert). An earlier version of this PR attempted to do the opposite – allow datasets to specify that their column types are dimension query names. That would cause problems and inconsistencies with the new paradigm for supplemental dimensions.

**Can a derived dataset have more than one column per dimension type?**
No. It must follow the same format as a regular dataset with one column for each dimension (perhaps time is a special case).

**Does the code auto-generate everything needed for a derived dataset that has dimensions that don’t match any project supplemental dimension?**
No. It generates a dimension file only. It does not generate required changes for the project config or base-to-supplemental dimension mappings. This is going to be a very tedious process for the user. But, I ran out of time for this milestone. Also, the dsgrid registry code needs improvement to fully support this process. It will need to detect that new supplemental dimensions and base-to-supplemental mappings should not affect existing datasets.


daniel-thom:
I would like to register comstock_projected and resstock_projected in StandardScenarios and then generate a new filtered registry. That will simplify a lot of the existing tests. In order to do that we need the dataset configs properly filled in with all of the dataset metadata and dimension information. Does one of you want to do that? I can share the auto-generated files.

dsgrid/query/derived_dataset.py
```diff
@@ -0,0 +1,178 @@
...
+            is_valid = True  # TODO: how can time be different? This was already mapped to project
+            table_records = None
+        else:
+            is_valid = is_dimension_valid_for_dataset(dim, df)
+            table_records = get_unique_values(df, dim.model.dimension_query_name)
+        if is_valid:
+            dim_ref = project.config.get_dimension_reference(dim.model.dimension_id)
+            dimension_references.append(serialize_model(dim_ref))
+        else:
+            records = dim.get_records_dataframe()
```
daniel-thom, line 78: This is a problematic section that may need changes. Suppose we are generating the dataset config for comstock_projected_2022. The dataframe produced by the query has project-mapped dimensions. The project dimension records for sector and metric do not match the records in the dataframe. The easiest and safest thing to do is generate a new records file for each of these dimensions and register a new dimension (that's what this code does). However, as it happens, this new records file will be identical to what the original comstock_reference dataset has. That obviously won't always be the case because a dataset may use different record IDs than the project.

So, what do we do here for dimensions that are not identical to the project?
1. Always generate new ones, even if duplicate. Rely on the user to change name and description so that they are not misleading.
2. Identify the source datasets from the query, read their dimension records and check for a match. Use the first dimension that matches. This would mean taking name, description, display_name, and query_name regardless of whether they are appropriate.
3. Loop through all dimensions looking for a match in records. Take the first one that matches.

dsgrid/dataset/dataset_schema_handler_standard.py
```diff
@@ -107,6 +116,15 @@ def make_project_dataframe_from_query(self, context: QueryContext, project_confi
         table_handler = PivotedTableHandler(project_config, dataset_id=self.dataset_id)
         return table_handler.convert_columns_to_query_names(ld_df)
 
+    @staticmethod
+    def _add_null_values(ld_df, null_lk_df):
+        if not null_lk_df.rdd.isEmpty():
+            for col in set(ld_df.columns).difference(null_lk_df.columns):
+                null_lk_df = null_lk_df.withColumn(col, F.lit(None))
```
lixiliu, line 123: I see why you need null timestamps now

dsgrid/query/derived_dataset.py
```diff
@@ -0,0 +1,168 @@
...
+
+    Returns
+    -------
+    bool
+        Returns True if the operation is successful.
+    """
+    metadata_file = QuerySubmitterBase.metadata_filename(query_path)
+    query_file = QuerySubmitterBase.query_filename(query_path)
+    table_file = QuerySubmitterBase.table_filename(query_path)
+    if not metadata_file.exists() or not query_file.exists():
```
lixiliu, line 45: no need to check table_file here?

dsgrid/query/derived_dataset.py
```diff
@@ -0,0 +1,168 @@
...
+    # TODO: should there be a warning if the current project version is later?
+    dimensions = []
+    dimension_references = []
+    for dim_type in DimensionType:
+        dims = getattr(metadata.dimensions, dim_type.value)
+        assert len(dims) == 1, dims
+        dim = project.config.get_dimension(next(iter(dims)))
+        df = read_dataframe(table_file)
+        if metadata.pivoted.dimension_type == dim_type:
+            is_valid = bool(metadata.pivoted.columns.difference(df.columns))
```
lixiliu, line 66: If there's a difference here, the metadata.pivoted.columns is considered valid?

dsgrid/query/derived_dataset.py
lixiliu: This is pertinent to the general registration of datasets and whether existing dim records should be reused as much as possible.

This will get complicated more in terms of version control. I wanna say dimensions should be reused as much as possible within a project? So option 2 or 3 (within project)?

dsgrid/dataset/dataset_schema_handler_one_table.py
```diff
@@ -50,9 +48,24 @@ def _check_one_table_data_consistency(self):
...
+        dimension_type_to_column = {}
+        column_to_dimension_type = {}
 
         time_dim = self._config.get_dimension(DimensionType.TIME)
-        time_columns = time_dim.get_timestamp_load_data_columns()
+        if self._config.model.data_schema.column_type == ColumnType.DIMENSION_TYPES:
+            time_columns = time_dim.get_timestamp_load_data_columns()
+            for dimension_type in DimensionType:
+                dimension_type_to_column[dimension_type] = dimension_type.value
+                column_to_dimension_type[dimension_type.value] = dimension_type
```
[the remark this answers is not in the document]
daniel-thom: My solution completes assignment of the dictionaries in one iteration. I know that doesn't really matter because the size is tiny. Is this really harder to read?

dsgrid/dataset/dataset_schema_handler_one_table.py
```diff
@@ -62,8 +75,10 @@ def _check_one_table_data_consistency(self):
             elif col in expected_pivoted_columns:
                 pivoted_dim_found = True
                 pivoted_cols.add(col)
+            elif col not in column_to_dimension_type:
```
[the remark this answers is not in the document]
daniel-thom: It should be correct in either case. Refer to lines 59 and 65.

dsgrid/dataset/dataset_schema_handler_standard.py
daniel-thom: The dataset contributor is the one who sets the schema. We do check this in `check_null_value_in_unique_dimension_rows(dim_table, exclude_columns=None)`

dsgrid/dataset/dataset_schema_handler_one_table.py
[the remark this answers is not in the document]
lixiliu: I get that, but I thought the cols could either be query_name or dimension_types, so shouldn't this check (whether to use column_to_dimension_type or the other dict) depends on if self._config.model.data_schema.column_type == ColumnType.DIMENSION_TYPES?

Or is column_to_dimension_type different depending on column_type?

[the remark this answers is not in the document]
daniel-thom: Yes. Lines 59 and 65

dsgrid/query/derived_dataset.py
```diff
@@ -0,0 +1,168 @@
...
+    """Return True if a derived dataset can be created from a query.
+
+    Returns
+    -------
+    bool
+    """
+    is_valid = True
+    if query.result.supplemental_columns:
+        is_valid = False
+        logger.error("Cannot create a derived dataset from a query with supplemental_columns")
```
lixiliu, line 143: I thought we can? And can register with the help of mappings?

daniel-thom: This is not the same thing as supplemental dimensions. Suppose you use county as the geography dimension but then request that state be added as a column. Now there are two geography dimensions, and so it is invalid.

lixiliu: But this is not checking the number of cols for a given dim type, but just that supplemental_columns exist.

daniel-thom: That inherently means that multiple columns exist. It would be different if the user chose a supplemental dimension in the query.

dsgrid/query/derived_dataset.py
```diff
@@ -0,0 +1,265 @@
...
+from dsgrid.registry.dimension_registry import DimensionRegistry
+from dsgrid.registry.registry_manager import RegistryManager
+from dsgrid.utils.files import dump_data
+from dsgrid.utils.spark import read_dataframe, get_unique_values
+
+
+logger = logging.getLogger(__name__)
+
+
+def create_derived_dataset_confg_from_query(
```
elainethale, line 26: nit: config, not confg

dsgrid/query/derived_dataset.py
```diff
@@ -0,0 +1,265 @@
...
+    dimension_references = []
+    dimension_mapping_references = []
+    base_dim_query_names = project.config.get_base_dimension_query_names()
+    for dim_type in DimensionType:
+        dimension_query_names = getattr(metadata.dimensions, dim_type.value)
+        assert len(dimension_query_names) == 1, dimension_query_names
+        dim_query_name = next(iter(dimension_query_names))
+        dim = project.config.get_dimension(dim_query_name)
+        if dim_type == DimensionType.TIME:
+            is_valid = True  # time should be identical to the project
```
elainethale, line 72: Looks like we're assuming that it is identical--can we actually check?

dsgrid/query/derived_dataset.py
```diff
@@ -0,0 +1,265 @@
...
+        "origin_organization": "",
+        "origin_date": "",
+        "origin_project": "",
+        "origin_version": "",
+        "source": "",
+        "data_classification": data_classification,
+        "use_project_geography_time_zone": True,
+        "dimensions": dimensions,
+        "dimension_references": dimension_references,
+    }
```
elainethale, line 184: Seems like it would be safer to create using the pydantic data model classes?

dsgrid/query/derived_dataset.py
```diff
@@ -0,0 +1,265 @@
...
+
+
+def _get_dimension_mapping_reference(dim_model: DimensionModel, project_config):
+    key, _ = project_config.get_base_to_supplemental_config(dim_model.dimension_query_name)
+    return {
+        "mapping_id": key.id,
+        "from_dimension_type": dim_model.dimension_type.value,
+        "to_dimension_type": dim_model.dimension_type.value,
+        "version": str(key.version),
+    }
```
elainethale, line 265: Again, could this use Pydantic data model directly?

dsgrid/project.py
```diff
@@ -159,11 +160,26 @@ def process_query(self, context: QueryContext, cached_datasets_dir: Path):
...
         pivoted_columns_sorted = sorted(pivoted_columns)
-        dim_columns = context.get_all_dimension_query_names()
-        time_columns = context.get_dimension_query_names(DimensionType.TIME)
+        match context.model.result.column_type:
+            case ColumnType.DIMENSION_QUERY_NAMES:
+                dim_columns = context.get_all_dimension_query_names()
+                time_columns = context.get_dimension_query_names(DimensionType.TIME)
+            case ColumnType.DIMENSION_TYPES:
+                dim_columns = {x.value for x in DimensionType if x != DimensionType.TIME}
+                time_columns = {"timestamp"}
```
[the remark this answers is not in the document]
daniel-thom: I forgot that we can get this from the project time dimension. Fixed.

dsgrid/query/derived_dataset.py
daniel-thom: I added the same check that runs when you register the dataset. Note that it can take some time. If it fails, I'm not planning to guess on what they want. There are several time types and I'm not sure how they would get into this situation intentionally. I can see how they might filter time and then try to register it, which would be incorrect. If you can provide a use case where all of this is expected and I can programmatically determine what the time dimension should be, I can add something.

dsgrid/query/derived_dataset.py
daniel-thom: I did this with dictionaries because it skips validation that would otherwise fail in the dimension section (because of record checks that are invalid in this context). I copied the comment from below here. I don't think it's any less safe. If any required field is missing, registration will fail.

Note that since we now may skip adding a time dimension, more checks will fail if we use the Pydantic models.

dsgrid/config/representative_period_time_dimension_config.py
```diff
@@ -219,13 +222,23 @@ def list_expected_dataset_timestamps(self):
...
-        ]
+        actual_timestamps = []
+        for row in load_data_df.select(*time_columns).distinct().sort(*time_columns).collect():
+            data = row.asDict()
+            num_none = 0
+            for val in data.values():
+                if val is None:
+                    num_none += 1
+            if num_none > 0:
+                if num_none != len(data):
```
lixiliu, line 235: Feels weird to me, I also feel like I've reviewed this part before.

If there's a null, then all cols must be null, so it seems there are two correct possibilities:
- records without any null
- a single record of null
So, to me, this means as long as num_none > 1: it's an error

daniel-thom: What about representative time with one-week-per-month in one-table format and missing data for specific dimension combinations? All 3 time columns must be null.

dsgrid/project.py
```diff
@@ -159,11 +160,33 @@ def process_query(self, context: QueryContext, cached_datasets_dir: Path):
         # TODO #202: Change this when we support long format.
         pivoted_columns = context.get_pivoted_columns()
         pivoted_columns_sorted = sorted(pivoted_columns)
-        dim_columns = context.get_all_dimension_query_names()
-        time_columns = context.get_dimension_query_names(DimensionType.TIME)
+
+        match context.model.result.column_type:
```
lixiliu, line 164: never seen this before, cool

dsgrid/query/derived_dataset.py
```diff
@@ -0,0 +1,283 @@
...
+            if metadata.pivoted.dimension_type == dim_type:
+                unique_data_records = metadata.pivoted.columns
+            else:
+                unique_data_records = _get_unique_data_records(
+                    df, dim.model, query.result.column_type
+                )
+            is_valid = _is_dimension_valid_for_dataset(dim, unique_data_records)
+        if is_valid:
+            dimension_references.append(_get_dimension_reference(dim.model, project.config))
+            if dim.model.dimension_query_name not in base_dim_query_names:
```
lixiliu, line 92: isn't `dimension_query_name` extracted already? It was used to get the dim obj

dsgrid/query/derived_dataset.py
```diff
@@ -0,0 +1,283 @@
...
+        return False
+
+    metadata = DatasetMetadataModel.from_file(metadata_file)
+    project = registry_manager.project_manager.load_project(
+        query.project.project_id, version=query.project.version
+    )
+    new_supplemental_dims_path = dst_path / "new_supplemental_dimensions"
+    df = read_dataframe(table_file)
+    # TODO: should there be a warning if the current project version is later?
+    dimensions = []
```
lixiliu, line 63: This is quite an operation here, would help if there's a sentence describing this section

dsgrid/query/models.py
```diff
@@ -431,6 +439,15 @@ def handle_dimension_filters(cls, dimension_filters):
                 dimension_filters[i] = make_dimension_filter(dimension_filter)
         return dimension_filters
 
+    @validator("column_type")
+    def check_column_type(cls, column_type, values):
+        if column_type == ColumnType.DIMENSION_TYPES and values["supplemental_columns"]:
+            # This would cause duplicate column names.
+            raise ValueError(
+                f"column_type={ColumnType.DIMENSION_TYPES} is incompatible with supplemental_columns"
```
lixiliu, line 447: Basically because all base dimensions would be retained by default, so can't have supplemental_columns. Can't swap a base for a supplemental in this case?

dsgrid/utils/dataset.py
```diff
@@ -86,7 +86,7 @@ def map_and_reduce_pivoted_dimension(df, records, pivoted_columns, operation, re
     final_columns = []
     dropped = set()
     for tid in to_ids:
-        column = f"{tid}_{operation}" if rename else tid
+        column = f"{tid}" if rename else tid
```
lixiliu, line 89: so just tid in either case then

lixiliu (approved):
Minor things, thanks!

dsgrid/config/representative_period_time_dimension_config.py
lixiliu: Got cha, I misinterpreted how data looks like. It's just a single row in dict form, where each item is a column.

dsgrid/query/models.py
daniel-thom: 1. Specify `dimension_query_name = "state"` to get state resolution in the derived dataset: **allowed**
2. Specify `dimension_query_name = "county` as well as `supplemental_columns = ["state"]`: **not allowed** because there will be two geography columns

lixiliu: Thanks, I understand it in the context of column_type == ColumnType.DIMENSION_TYPES

For my understanding, do we allow more than 1 col for a dimension in column_type == ColumnType.DIMENSION_QUERY_NAMES context? I know for derived datasets, we don't allow query_names

If so, we allow the following:
- `dimension_query_name = "state" + supplemental_columns = ["county"]`: this would be invalid to me, do we catch that?
- `dimension_query_name = "state" + supplemental_columns = ["census_division"]`

daniel-thom: We will silently ignore the first case and just give you "state." Is that a bug? Probably. We could check that any supplemental column is a supplemental dimension.

We add the supplemental columns before aggregating to the supplemental dimension. As long as the supplemental columns are bigger than the primary dimension, the second case should be OK. Specifying the primary as `census_division` and `supplemental_columns = ["state"]` is invalid and we don't catch it.

lixiliu: Yeah, adding a no-base dim check for `supplemental_columns` seem simple enough. The rest seems harmless for now.

PR #216, opened 2023-01-23
Use pydantic serialization

### LINKED ISSUE #195: Cleanup code to use custom JSON encoders through pydantic
The code performs some manual serialization in `serialize_model` which shouldn't be necessary.

daniel-thom:
Fixes #195 

Includes a few minor changes
- Patch a previous fix that replaced VersionInfo with str in registry models but did not update some code.
- Use a faster query in the API tests
- Remove some noisy logging


dsgrid/config/mapping_tables.py
```diff
@@ -130,12 +130,21 @@ def add_records(cls, records, values):
...
-            exclude = set()
-        exclude.add("records")
-        data = super().dict(by_alias=by_alias, exclude=exclude)
-        return serialize_model_data(data)
+    def dict(self, *args, **kwargs):
+        exclude = {"records"}
+        if "exclude" in kwargs and kwargs["exclude"] is not None:
+            kwargs["exclude"].union(exclude)
+        else:
+            kwargs["exclude"] = exclude
```
elainethale, line 138: These lines of code are repeated a lot in this PR. Is that okay/should it be refactored?

elainethale (approved):
One minor comment, and obviously the tests should be passing before merging.

elainethale:
I see now that it's not the tests that aren't passing, but rather the code coverage that's dropping slightly. Up to you if you want to fix.

lixiliu (approved):
Nice, the code looks simpler

dsgrid/config/mapping_tables.py
daniel-thom: Fixed

PR #225, opened 2023-02-13
Use Spark to upload datasets to registry

daniel-thom:
Running an `shutil.copytree` on a 1.6 TB directory was not a great idea. This is much better - and should also work on AWS.


lixiliu (approved):
interesting.

Issue #222, opened 2023-02-02
Ensure that the dsgrid EFS period-ending datasets are being mapped correctly to project time


elainethale:
All the data_source = "dsgrid_efs" datasets have 2012 EST, period-ending time. Is this already being handled correctly? If not, we need to implement the mapping.

elainethale:
Most important things to do:
- Throw error if converted dataset timestamps do not match project timestamps
- Provide the method for this particular conversion

PR #228, opened 2023-02-15
Add TimeIntervalType check and shift

lixiliu:
Completes JIRA Story DSGRID-
Closes GitHub Issue # https://github.com/dsgrid/dsgrid/issues/222
dsgrid-test-data PR: https://github.com/dsgrid/dsgrid-test-data/pull/30

Q: What to do when time interval type == instantaneous for project and not the case for dataset and vice versa?
A: Current code behavior would accept (without shifting dataset time) if time_range is the same between dataset and project and one of them has "instantaneous" time interval type, e.g., the following case:

project has 2012-01-01 01:00:00 - 2012-12-31 23:00:00 in `period_beginning` and dataset has the same time range but in `instantaneous` interval

Thoughts from @elainethale :
```
instantaneous should have different units, e.g., MW instead of MWh
```
## Checklist

- [ ] All tests pass


dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -394,4 +395,43 @@ def _convert_time_dimension(self, load_data_df, project_config):
...
+            elif (
+                dtime_interval == TimeIntervalType.PERIOD_ENDING
+                and ptime_interval == TimeIntervalType.PERIOD_BEGINNING
+            ):
+                load_data_df = load_data_df.withColumn(
+                    time_col,
+                    F.col("timestamp")
+                    - F.expr(f"INTERVAL {dtime.get_frequency().seconds} SECONDS"),
+                )
+
```
daniel-thom, line 435: What do you think of this pattern instead? All of the ifs go away
```
match (dtime_interval, ptime_interval):
    case (PERIOD_BEGINNING, PERIOD_ENDING):
        # do stuff
    case (PERIOD_ENDING, PERIOD_BEGINNING):
        # do stuff
```

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -394,4 +395,43 @@ def _convert_time_dimension(self, load_data_df, project_config):
...
+        # Shifiting time here does not override dataset time interval type, would it be confusing?
```
daniel-thom, line 436: Seems fine to me. This is part of the mapping process.

daniel-thom (commented):
Do the tests cover this?

lixiliu:
> Do the tests cover this?

Some do, e.g., make_us_test_registry has project requirement: period-beginning but some of its dataset is period-ending (similar to EFS). I am relying on CI at the moment to see if existing tests pass since I lost access to nrel-aws-dsgrid

After CI passes, I could add some more tests to cover time_interval_type == None or Instantaneous cases

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -394,4 +395,36 @@ def _convert_time_dimension(self, load_data_df, project_config):
 
         if time_dim.model.is_time_zone_required_in_geography():
             load_data_df = load_data_df.drop("time_zone")
+
+        load_data_df = self._convert_dataset_time_to_project_time_interval(
+            load_data_df, project_config
+        )
```
elainethale, line 401: Should this be part of time_dim.convert_dataframe instead of a post-process?

I'm asking in part because reading _convert_dataset_time_to_project_time_interval is confusing. The input load_data_df has already been partially converted to project time, but not completely, and just looking at that function it is not clear that that should be the case.

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -394,4 +395,36 @@ def _convert_time_dimension(self, load_data_df, project_config):
...
+        return load_data_df
+
+    def _convert_dataset_time_to_project_time_interval(self, load_data_df, project_config):
+        """Shift dataset time to match project time based on TimeIntervalType"""
```
elainethale, line 405: If this function stays here the docstring should clarify that load_data_df is the dataset with the time dimension already partially converted to project time. (Assuming I'm reading this function and _convert_time_dimension correctly. Alternatively, maybe this function could be part of TimeDimension / called by time_dim.convert_dataframe?

dsgrid/registry/project_registry_manager.py
```diff
@@ -871,6 +877,37 @@ def _submit_dataset(
...
+                case (TimeIntervalType.PERIOD_BEGINNING, TimeIntervalType.PERIOD_ENDING):
+                    dtime_timestamps = [
+                        DatetimeTimestampType(x.timestamp + dtime.get_frequency())
+                        for x in dtime.list_expected_dataset_timestamps()
+                    ]
+                case (TimeIntervalType.PERIOD_ENDING, TimeIntervalType.PERIOD_BEGINNING):
+                    dtime_timestamps = [
+                        DatetimeTimestampType(x.timestamp - dtime.get_frequency())
+                        for x in dtime.list_expected_dataset_timestamps()
+                    ]
```
elainethale, line 902: This is essentially a repeat of the logic in DatasetSchemaHandlerBase._convert_dataset_time_to_project_time_interval. Should this conversion logic be a utility somewhere with the time functionality?

elainethale (commented):
Logic looks okay, but I have some organizational/readability comments for your consideration. And I agree with @daniel-thom that something should be added to tests.

dsgrid/registry/project_registry_manager.py
lixiliu: Same logic but completely different implementation because one is for list of time objects and the other for spark df. So not the easiest to factor out.

dsgrid/config/time_dimension_base_config.py
```diff
@@ -120,3 +123,71 @@ def list_expected_dataset_timestamps(self):
             List of tuples of columns representing time in the load_data table.
 
         """
+
+    def _convert_time_to_project_time_interval(
+        self, df=None, project_time_dim=None, wrap_time=True
+    ):
+        """Shift time to match project time based on TimeIntervalType
+        If time range spills over into anothr year after time interval alignment,
```
elainethale, line 131: typo: anothr

elainethale (commented):
It looks like maybe the test data submodule commit maybe hasn't been pushed or something and that's causing the tests not to pass? I'd like to approve, since it looks like all the comments have been addressed, but we need the tests passing first.

dsgrid/config/time_dimension_base_config.py
```diff
@@ -110,3 +122,78 @@ def list_expected_dataset_timestamps(self):
...
+                        )
+                        * 24
+                        * 3600
+                    ).cast("timestamp"),
+                )
+                .union(df.filter(~F.col(time_col).isin(diff)))
+            )
+
+        dataset_time = df.agg(F.collect_set(time_col)).collect()[0][0]
+        diff = set(dataset_time).difference(set(project_time))
```
daniel-thom, line 194: Not a complaint - I'm sure I've written identical lines many times. In case it helps readability, this could be
```
if not dataset_time.issubset(project_time):
    assert/raise
```
It's better to capture the diff if you log it, but you aren't doing that.

dsgrid/config/time_dimension_base_config.py
```diff
@@ -110,3 +122,78 @@ def list_expected_dataset_timestamps(self):
...
+
+        if diff:
+            # take most common year from time col
+            main_year = (
+                df.groupBy(F.year(time_col)).count().select(f"year({time_col})").collect()[0][0]
+            )
+
+            # time-wrap by changing the year
+            df = (
```
daniel-thom, line 176: Nice

dsgrid/config/time_dimension_base_config.py
```diff
@@ -110,3 +122,78 @@ def list_expected_dataset_timestamps(self):
...
+
+        # If dataset_time does not match project_time, try time-wrapping
+        dataset_time = df.agg(F.collect_set(time_col)).collect()[0][0]
+        project_time = (
+            project_time_dim.build_time_dataframe().agg(F.collect_set(time_col)).collect()[0][0]
+        )
+
+        diff = set(dataset_time).difference(set(project_time))
+
+        if diff:
```
daniel-thom, line 169: If there is no diff, should this return early? Later you take the diff again.

dsgrid/config/time_dimension_base_config.py
```diff
@@ -110,3 +122,78 @@ def list_expected_dataset_timestamps(self):
...
+                    df = df.withColumn(
+                        time_col,
+                        F.col(time_col)
+                        - F.expr(f"INTERVAL {self.get_frequency().seconds} SECONDS"),
+                    )
+
+        if not wrap_time:
+            return df
+
+        # If dataset_time does not match project_time, try time-wrapping
```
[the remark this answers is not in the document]
lixiliu: If project_time spans more than a year, then main_year is no longer the most common year. Instead, we need to check to see if time diff is to the left or right of project_time and we determine the time_delta from there.
e.g., if project time is two years from 2018-2019, dataset_time has a couple 2017 timestamps, then main_year is 2019 to get the delta of +2 yrs.
If dataset_time instead has some 2020 timestamps that need to be converted, then main_year is 2018 to get a delta of -2 yrs.

I'll change to that

tests/test_derived_datasets.py
```diff
@@ -102,13 +108,15 @@ def test_create_derived_dataset_config(tmp_path):
     orig_df = read_dataframe(REGISTRY_PATH / "data" / dataset_id / "1.0.0" / "table.parquet")
     new_df = read_dataframe(query_output / "table.parquet")
     assert sorted(new_df.columns) == sorted(orig_df.columns)
+
+    # orig_df does not have time-wrapping, so need to load project_time to convert
+    project_time_dim = project.config.get_base_dimension(DimensionType.TIME)
+    orig_df = project_time_dim.convert_dataframe(orig_df, project_time_dim)
```
elainethale, line 114: This seems awkward. After this branch is merged in, if we remade the table.parquet, would it already be in project time? If so, maybe we need to add a TODO to update the original data sometime soon?

tests/test_project_time.py
```diff
@@ -86,6 +87,21 @@ def test_convert_to_project_time(registry_mgr):
...
+    )
+    compare_time_conversion(
+        comstock_time_dim, project_time_dim, wrap_time=False, expect_error=True
+    )
+    compare_time_conversion(
+        resstock_time_dim, project_time_dim, wrap_time=True, expect_error=False
+    )
+    compare_time_conversion(
+        comstock_time_dim, project_time_dim, wrap_time=True, expect_error=False
+    )
```
elainethale, line 101: Are the resstock and comstock time dimensions identical? If so, should we just test one of them?

elainethale (commented):
A couple of minor comments.

But I agree with Dan's comment on the test-data that the changes made there should be reflected somewhere else--either the StandardScenarios repo or the filtering script.

dsgrid/config/time_dimension_base_config.py
```diff
@@ -110,3 +123,82 @@ def list_expected_dataset_timestamps(self):
...
+        df = (
+            df.filter(F.col(time_col).isin(diff))
+            .withColumn(
+                time_col,
+                F.from_unixtime(F.unix_timestamp(time_col) + time_delta).cast("timestamp"),
+            )
+            .union(df.filter(~F.col(time_col).isin(diff)))
+        )
+
+        dataset_time = set(df.agg(F.collect_set(time_col)).collect()[0][0])
```
daniel-thom, line 198: `set(df.select(time_col).distinct().collect())` is more straightforward and is likely a better solution.

elainethale:
@lixiliu, any chance of getting this merged in today?

dsgrid/config/time_dimension_base_config.py
```diff
@@ -110,3 +123,111 @@ def list_expected_dataset_timestamps(self):
...
+
+    @staticmethod
+    def _apply_time_wrap(df, project_time_dim):
+        """If dataset_time does not match project_time, apply time-wrapping"""
+
+        time_col = project_time_dim.get_timestamp_load_data_columns()
+        assert len(time_col) == 1, time_col
+        time_col = time_col[0]
+
+        project_time = set(
```
daniel-thom, line 172: This creates a list and then another list and then a set. You could drop the middle list.
```
{row[0] for row in project_time_dim ...}
```

dsgrid/config/time_dimension_base_config.py
```diff
@@ -110,3 +123,111 @@ def list_expected_dataset_timestamps(self):
...
+            [
+                row[0]
+                for row in df.select(time_col)
+                .filter(f"{time_col} IS NOT NULL")
+                .distinct()
+                .collect()
+            ]
+        )
+        diff = dataset_time.difference(project_time)
+        print(diff)
```
daniel-thom, line 192: stale

dsgrid/config/time_dimension_base_config.py
```diff
@@ -110,3 +123,114 @@ def list_expected_dataset_timestamps(self):
...
+        Shift time_column by time_step in df as needed by comparing from_time_interval to to_time_interval.
+        If new_time_column is None, time_column is shifted in place, else shifted time is added as new_time_column in df
+        """
+        if from_time_interval == to_time_interval:
+            return df
+
+        if new_time_column is None:
+            new_time_column = time_column
+
+        match (from_time_interval, to_time_interval):
```
daniel-thom, line 167: What about INSTANTANEOUS?

lixiliu: We do nothing I think

daniel-thom: This code certainly does nothing in that case. Is that correct? Should we log anything? If this is correct then I suggest adding a comment here indicating that.

daniel-thom: @elainethale The conclusion from Lixi and I is that we should raise an exception in the case of a mismatch with instantaneous. There is a related comment on this subject at the top of the PR. Are you OK with punting on this question?

elainethale: Yes, I am okay with punting. Punting can be a TODO, logging a warning, or raising an exception. Up to you all.

elainethale (approved):
I'm okay with this if Dan is. I see that the code coverage is declining somewhat, maybe because we are temporarily skipping tests? Are you able to check the override button and merge? If not, I can do it.

dsgrid/config/time_dimension_base_config.py
daniel-thom: I added an exception and opened #239

daniel-thom:
>  I see that the code coverage is declining somewhat, maybe because we are temporarily skipping tests?
Yes, that is definitely contributing. It should get fixed when I regenerate the data.

daniel-thom (approved):

PR #230, opened 2023-02-21
Remove null-timestamp rows that are obsolete after aggregations

daniel-thom:
This PR fixes two problems:
- A derived dataset like comstock_projected has rows with `time_est == null` to designate data that is supposed to be there but is expectedly missing. These are obsolete after dropping the sector and subsector dimensions, but weren't being dropped. Now they are.
- The first fix uncovered an existing problem where weren't properly tracking the column names in the table resulting from a query. This is more of a bandaid than a true fix. We are still treating the column names as "dimension_query_names" even though they are really just labels. One example of why this is still a problem is here: https://github.com/dsgrid/dsgrid/blob/466017cb7dc3bfba667e75019bea65ab3699f658/dsgrid/dataset/table_format_handler_base.py#L80 This function runs after all aggregations and allows users to say they want record names instead of IDs. Using actual dimension_query_names here is convenient. If they added `state` as a column, we can very easily look up the `state` supplemental dimension records. If the user says that they want to use an alias, this will break. We could track this with an additional layer of mappings. Let me know what you think. I'm not sure how much of a problem it is just yet. I could fix it now or defer to later. We could also remove the alias functionality and/or the replace-ids-with-names functionality.

### DIFF
--- a/dsgrid/dataset/pivoted_table.py
+++ b/dsgrid/dataset/pivoted_table.py
@@ -8,7 +8,7 @@ from dsgrid.dimension.base_models import DimensionType
 from dsgrid.exceptions import DSGInvalidParameter
 from dsgrid.query.models import AggregationModel
 from dsgrid.query.query_context import QueryContext
-from dsgrid.utils.dataset import map_and_reduce_pivoted_dimension
+from dsgrid.utils.dataset import map_and_reduce_pivoted_dimension, remove_invalid_null_timestamps
 from dsgrid.utils.spark import get_unique_values
 from .table_format_handler_base import TableFormatHandlerBase
 
@@ -73,7 +73,20 @@ class PivotedTableHandler(TableFormatHandlerBase):
         self, df, aggregations: List[AggregationModel], context: QueryContext
     ):
         df = self.process_pivoted_aggregations(df, aggregations, context)
-        return self.process_stacked_aggregations(df, aggregations, context)
+        orig_id = id(df)
+        df = self.process_stacked_aggregations(df, aggregations, context)
+
+        if id(df) != orig_id:
+            # The table could have NULL timestamps that designate expected-missing data.
+            # Those rows could be obsolete after aggregating stacked dimensions.
+            # This is an expensive operation, so only do it if the dataframe changed.
+            pivoted_columns = context.get_pivoted_columns()
+            time_columns = context.get_dimension_query_names(DimensionType.TIME)
+            if time_columns:
+                stacked_columns = set(df.columns) - pivoted_columns.union(time_columns)
+                df = remove_invalid_null_timestamps(df, time_columns, stacked_columns)
+
+        return df
 
     def process_pivoted_aggregations(
         self, df, aggregations: List[AggregationModel], context: QueryContext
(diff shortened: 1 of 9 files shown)


dsgrid/dataset/pivoted_table.py
```diff
@@ -158,39 +171,62 @@ def process_stacked_aggregations(
...
                 "Aggregated dimensions with groupBy %s and agg %s", group_by_cols, agg_expr
             )
 
-        final_query_names = {x: set() for x in DimensionType if x != pivoted_dimension_type}
-        for column in group_by_query_names:
-            dim = self.project_config.get_dimension(column)
-            final_query_names[dim.model.dimension_type].add(column)
-
         for dim_type, columns in final_query_names.items():
+            if dim_type in dropped_dimensions and columns:
```
daniel-thom, line 211: I did some refactoring while making a fix to this function. The main thing is that when the function exits the columns of the dataframe are stored as "dimension query names" in the query context. But they aren't always actual dimension query names.

elainethale:
I don't think we want to remove the alias functionality, because SQL-default names are often not that helpful.

And I don't think we cant to remove the replace ids with names functionality.

If it's not too bad to add the additional layer of mapping needed to support both, I'd like to go ahead and do it now.

daniel-thom:
> 

I addressed the issues. The code now stores a list of column_name/dimension_query_name combinations for each column in a query result table. In most cases they will be the same. In cases where the user asks for an alias they will be different. Depending on filtering, the dimension_query_name may not perfectly represent the records in the column. A perhaps more accurate name is "source_dimension_query_name." That is a bit unwieldy and so I stuck with dimension_query_name. If you all want to make a minor change while I'm out, feel free.

daniel-thom:
I regenerated the comstock portion of the ReEEDS dataset and it matches the previous version except for the NULL timestamps.

daniel-thom:
@elainethale @lixiliu Ready for review.

dsgrid/query/models.py
```diff
@@ -146,17 +154,61 @@ class TableFormatType(enum.Enum):
...
+        if metadata.make_key() not in {x.make_key() for x in container}:
+            container.append(metadata)
+
+    def get_metadata(self, dimension_type: DimensionType):
+        """Return the dimension metadata."""
+        return getattr(self, dimension_type.value)
+
+    def replace_metadata(self, dimension_type: DimensionType, metadata: DimensionMetadataModel):
+        """Replace the dimension metadata."""
+        setattr(self, dimension_type.value, metadata)
```
elainethale, line 195: This doesn't have to be 

```
setattr(self, dimension_type.value, [metadata])
```

?

tests/test_dataset_utils.py
```diff
@@ -192,6 +196,37 @@ def test_is_noop_mapping_false():
...
+    stacked = ["county", "subsector"]
+    time_col = "timestamp"
+    result = remove_invalid_null_timestamps(df, {time_col}, stacked)
+    result = (
+        df.join(
+            df.groupBy(*stacked).agg(F.count_distinct(time_col).alias("count_time")), on=stacked
+        )
+        .filter(f"{time_col} IS NOT NULL or count_time == 0")
+        .drop("count_time")
+    )
```
elainethale, line 224: Why is result overwritten with this line that duplicates the logic in the function `remove_invalid_null_timestamps`?

elainethale (approved):
A couple of minor questions. Thanks for implementing column name tracking!

tests/test_dataset_utils.py
daniel-thom: Oops.

PR #240, opened 2023-04-27
Fix dataset checks

daniel-thom:
Fixes two small bugs in the query and derived-dataset code.

### DIFF
--- a/dsgrid/utils/dataset.py
+++ b/dsgrid/utils/dataset.py
@@ -165,7 +165,9 @@ def check_null_value_in_unique_dimension_rows(dim_table, exclude_columns=None):
 def is_noop_mapping(records: pyspark.sql.DataFrame) -> bool:
     """Return True if the mapping is a no-op."""
     return records.filter(
-        (records.from_id != records.to_id) | (records.from_fraction != 1.0)
+        (records.to_id.isNull() & ~records.from_id.isNull())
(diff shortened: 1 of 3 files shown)


dsgrid/utils/dataset.py
```diff
@@ -165,7 +165,9 @@ def check_null_value_in_unique_dimension_rows(dim_table, exclude_columns=None):
 def is_noop_mapping(records: pyspark.sql.DataFrame) -> bool:
     """Return True if the mapping is a no-op."""
     return records.filter(
-        (records.from_id != records.to_id) | (records.from_fraction != 1.0)
+        (records.to_id.isNull() & ~records.from_id.isNull())
```
elainethale, line 168: Should there also be:

```
(~records.to_id.isNull() & records.from_id.isNull())
```

?

elainethale (approved):
Minor question. Otherwise looks good to me.

dsgrid/utils/dataset.py
elainethale: Good point. Null on the from_id side probably doesn't happen? That would correspond to being explicit about "this thing exists in the project but we aren't mapping anything to it."

daniel-thom: Sounds good. It would be annoying bug to figure out and the extra check won't harm anything.

Issue #243, opened 2023-05-01
Stacked aggregations do not work with ColumnType.DIMENSION_TYPES


daniel-thom:
This function (https://github.com/dsgrid/dsgrid/blob/fee4ddc98ce94e72d82f1243adcfeb981ab9c438/dsgrid/dataset/pivoted_table.py#L198) was written assuming that we would always convert dataset columns from dimension types to dimension query names before getting here. There is code to allow application of SQL functions and aliases. It does not work if the query JSON file says to use dimension types instead of query names (that is what we must do when creating derived datasets).

PR #244, opened 2023-05-12
Fix stacked aggregations in queries

daniel-thom:
The code was not correctly handling stacked aggregations when the user requested final columns to be dimension types instead of dimension query names.

Closes GitHub Issue #243

### DIFF
--- a/dsgrid/config/dimensions.py
+++ b/dsgrid/config/dimensions.py
@@ -271,15 +271,14 @@ class DimensionModel(DimensionBaseModel):
     @validator("file_hash")
     def compute_file_hash(cls, file_hash, values):
         if "filename" not in values:
-            # TODO #194
-            # We are getting here for Time. That shouldn't be happening.
-            # This seems to work, but something is broken.
             return None
         return file_hash or compute_file_hash(values["filename"])
 
     @validator("records", always=True)
     def add_records(cls, records, values):
         """Add records from the file."""
+        if not {"filename", "cls", "name"}.issubset(values):
+            return None
         dim_class = values["cls"]
         if records:
             if isinstance(records[0], dict):
--- a/dsgrid/config/project_config.py
+++ b/dsgrid/config/project_config.py
@@ -774,6 +774,21 @@ class ProjectConfig(ConfigBase):
         # TODO: what about benchmark and historical?
         return False
 
+    def get_load_data_time_columns(self, dimension_query_name):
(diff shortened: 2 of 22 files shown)


dsgrid/config/dimensions.py
```diff
@@ -271,15 +271,14 @@ def check_file(cls, filename):
     @validator("file_hash")
     def compute_file_hash(cls, file_hash, values):
         if "filename" not in values:
-            # TODO #194
```
daniel-thom, line 274: Not reproducible. I think we were getting here because of other errors. This code is correct.

dsgrid/config/dimensions.py
```diff
@@ -271,15 +271,14 @@ def check_file(cls, filename):
...
-            # We are getting here for Time. That shouldn't be happening.
-            # This seems to work, but something is broken.
             return None
         return file_hash or compute_file_hash(values["filename"])
 
     @validator("records", always=True)
     def add_records(cls, records, values):
         """Add records from the file."""
+        if not {"filename", "cls", "name"}.issubset(values):
+            return None
```
elainethale, line 281: Just checking--is it okay/expected that calling this method might result in nothing happening? Seems like yes because this is a validator?

dsgrid/config/project_config.py
```diff
@@ -774,6 +774,21 @@ def has_dataset(self, dataset_id, status=None):
...
+        Parameters
+        ----------
+        dimension_query_name : str
+
+        Returns
+        -------
+        list
+        """
+        dim = self.get_dimension(dimension_query_name)
+        time_columns = dim.get_timestamp_load_data_columns()
```
elainethale, line 789: Should this be `dim.get_time_load_data_columns()`? Or just mimic the name here, `get_load_data_time_columns`?

elainethale (approved):

dsgrid/config/dimensions.py
daniel-thom: We've had a lot of problems like this. Pydantic always runs all validators, which can be problematic if a validator uses the `values` dict. That dict gets populated with values from successfully validated fields. If a validator fails, that field will not be in the dict and so every future validator needs to handle the case. I wish that they had a fail-fast option.

dsgrid/config/project_config.py
daniel-thom: This was named long ago. I can change it if it won't cause merge conflicts for @lixiliu

PR #247, opened 2023-06-09
Disable AWS tests

daniel-thom:
Changes:
- File syncing only applies to datasets now. Fix this in the code. (There is no solution for db syncing yet.)
- Set the default tests directory for pytest. Exclude `tests_aws` since they are non-functional.


elainethale:
@daniel-thom, is this PR obsolete?

daniel-thom:
> @daniel-thom, is this PR obsolete?

Still valid. I forgot about this one.

Issue #218, opened 2023-01-27
Programmatically support unit conversion


elainethale:
Right now unit conversion is supported via dimension mappings. It would be cleaner if dsgrid could register and apply unit conversions as their own thing.

elainethale:
Consensus:
- Use pint
- Project defines unit library (could be defaults from Pint)
- Datasets must define new units/unit conversions if they use different strings than the project unit library (e.g., see https://pint.readthedocs.io/en/stable/advanced/defining.html). This is similar to mapping between dataset and project labels for dimension elements
- Metric dimensions have unit columns (I think this is already true in practice, but maybe not strictly enforced--we should check)

elainethale:
Closed with #248

PR #248, opened 2023-06-22
Convert units when mapping a dataset to project

daniel-thom:
Closes GitHub Issue #218 

This PR adds support for unit conversion. It will be in draft state until we rebuild the StandardScenarios registry with the desired unit definitions for both base and supplemental dimensions.

### DIFF
--- a/dev/README.md
+++ b/dev/README.md
@@ -96,10 +96,16 @@ If you don't copy the files, you can specify the config file with `arangodb --co
 
 #### Local
 Run the ArangoDB Docker container by following instructions at
-https://www.arangodb.com/download-major/docker/. For example:
+https://www.arangodb.com/download-major/docker/. Note the details about data persistence.
+
+For example:
 
 ```
-docker run --name=arango-container -p 8529:8529 -e ARANGO_ROOT_PASSWORD=openSesame arangodb/arangodb:3.10.4
+$ docker create --name arangodb-persist arangodb true
+```
+
+```
+$ docker run --name=arango-container --volumes-from arangodb-persist -p 8529:8529 -e ARANGO_ROOT_PASSWORD=openSesame arangodb/arangodb:3.10.4
 ```
 
 Once the docker container is running, Arango commands will need to be preceded with `docker exec`. For example, using the container name from above:
--- a/dsgrid/config/dataset_config.py
+++ b/dsgrid/config/dataset_config.py
@@ -305,11 +305,10 @@ class DatasetConfigModel(DSGBaseModel):
         description="List of data tags",
         default=[],
     )
-    enable_metric_unit_conversion: bool = Field(
+    enable_unit_conversion: bool = Field(
         default=True,
-        description="If true, dsgrid will convert metric values to project units based on the "
-        "contents of the 'unit' column of the metric dimension records. If the dataset "
-        "defines a dimension mapping for unit conversion then this value should be false.",
+        description="If the dataset uses its dimension mapping for the metric dimension to also "
+        "perform unit conversion, then this value should be false.",
     )
     # This field must be listed before dimensions.
     use_project_geography_time_zone: bool = Field(
--- a/dsgrid/config/dataset_config.py
+++ b/dsgrid/config/dataset_config.py
@@ -305,6 +305,12 @@ class DatasetConfigModel(DSGBaseModel):
         description="List of data tags",
         default=[],
     )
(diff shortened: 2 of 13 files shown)


dev/README.md
```diff
@@ -96,10 +96,16 @@ If you don't copy the files, you can specify the config file with `arangodb --co
...
+For example:
+
+```
+$ docker create --name arangodb-persist arangodb true
+```
 
 ```
-docker run --name=arango-container -p 8529:8529 -e ARANGO_ROOT_PASSWORD=openSesame arangodb/arangodb:3.10.4
+$ docker run --name=arango-container --volumes-from arangodb-persist -p 8529:8529 -e ARANGO_ROOT_PASSWORD=openSesame arangodb/arangodb:3.10.4
 ```
```
elainethale, line 109: Nice--so these changes mean that you don't have to reload data into the database each time you spin it up?

dsgrid/config/dataset_config.py
```diff
@@ -305,6 +305,12 @@ class DatasetConfigModel(DSGBaseModel):
         description="List of data tags",
         default=[],
     )
+    enable_metric_unit_conversion: bool = Field(
```
elainethale, line 308: Just reading this makes me think "metric system" or converting everything into metric units instead of ip units. Would it be clearer to say "metric dimension"? Or just "enable_unit_conversion" (since metric dimensions are the only ones with units anyway)?

dsgrid/config/dataset_config.py
```diff
@@ -305,6 +305,12 @@ class DatasetConfigModel(DSGBaseModel):
...
+        default=True,
+        description="If true, dsgrid will convert metric values to project units based on the "
+        "contents of the 'unit' column of the metric dimension records. If the dataset "
+        "defines a dimension mapping for unit conversion then this value should be false.",
```
elainethale, line 312: Suggested edit: "If the dataset uses its dimension mapping for the metric dimension to also perform unit conversion, then this value should be false."

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -11,6 +11,7 @@
 from dsgrid.exceptions import DSGInvalidDataset, DSGInvalidQuery
 from dsgrid.dimension.time import TimeDimensionType
 from dsgrid.query.query_context import QueryContext
+from dsgrid.units.energy import convert_units
```
elainethale, line 14: How does this change when we add other sorts of metrics (e.g., capacity, population/counts)?

dsgrid/dataset/pivoted_table.py
```diff
@@ -227,6 +240,8 @@ def process_stacked_aggregations(
             if not columns:
                 continue
 
+            # TODO: This does not handle the scenario where the metric dimension is stacked and
+            # needs unit conversion.
```
elainethale, line 244: Is there somewhere we should be raising a NotImplementedError?

dsgrid/units/energy.py
```diff
@@ -0,0 +1,178 @@
...
+
+KWH = "kWh"
+MWH = "MWh"
+GWH = "GWh"
+TWH = "TWh"
+THERM = "therm"
+MBTU = "MBtu"
+KWH_PER_THERM = 29.3001111
+KWH_PER_MBTU = 0.293071
+THERMS_PER_MBTU = 0.01
```
elainethale, line 19: @lixiliu, what version of "therm" should we be using? And what reference should we (maybe) provide here in a comment? Per https://en.wikipedia.org/wiki/Therm, this appears to be the "US" version of Therm with BTU defined by water at 59 F (https://en.wikipedia.org/wiki/British_thermal_unit).

(I also appreciate the latter link's quick note that BTU definitions vary by up to 0.5%.)

dev/README.md
daniel-thom: Correct

dsgrid/config/dataset_config.py
daniel-thom: Good point. I like`enable_unit_conversion`.

dsgrid/dataset/dataset_schema_handler_base.py
daniel-thom: I would probably use 
```
import dsgrid.units.energy as energy
energy.convert_units()
```
I can make the change now.

elainethale (approved):

PR #260, opened 2023-10-11
Update HPC instructions for Kestrel

daniel-thom:
Updates:
1. We no longer have access to Eagle, and so I need to change the instructions for running on Kestrel. Note that `Singularity` has been rebranded as `Apptainer`.
2. ArangoDB stopped providing binaries for Mac, so I removed that content.
3. Added some error checks for a condition exposed by the DECARB project registration.


elainethale (approved):
LGTM

Issue #249, opened 2023-07-11
Supplemental dimensions always need a mapping


elainethale:
Currently in the project config, a list of 'supplemental_dimensions' is followed by a list of `base_to_supplemental` `dimension_mappings`. However, each supplemental dimension basically corresponds to exactly one of the base_to_supplemental mappings, and it would be easier for project coordinators if each member of the pair was next to each other.

Upon discussion, the team agreed that the cleanest fix for this from a user perspective would be for the supplemental dimension model to have a mapping model, e.g.,

```json5
supplemental_dimensions: [
    {
            class: 'CensusDivision',
            description: 'US Census Divisions',
            file: 'dimensions/supplemental/census_divisions.csv',
            module: 'dsgrid.dimension.standard',
            name: 'US Census Divisions',
            display_name: 'Census Division',
            type: 'geography',
            mapping: {
                description: 'Maps US Counties 2020 L48 to CensusDivision',
                file: 'dimension_mappings/base_to_supplemental/lookup_county_to_censusdivision.csv',
                mapping_type: 'many_to_one_aggregation',
                from_dimension: { // Not needed because implied to be project base geography dimension?
                    type: 'geography',
                    name: 'US Counties 2020 L48',
                },
                to_dimension: { // Not needed because implied to be this dimension?
                    name: 'US Census Divisions',
                    type: 'geography',
                },
            }
    },
    ...
]
```

PR #261, opened 2023-10-13
Move base-to-supplemental mappings in ProjectConfig

daniel-thom:
This PR makes it easier for project coordinators to define mappings for supplemental dimensions. It also adds the dsgrid-project-StandardScenarios repository as a submodule for better testing.

Closes GitHub Issue #249

### DIFF
--- a/tests/cli/test_registry.py
+++ b/tests/cli/test_registry.py
@@ -15,6 +15,10 @@ from dsgrid.tests.common import (
     replace_dimension_names_with_current_ids,
 )
 
+STANDARD_SCENARIOS_PROJECT_REPO = (
+    Path(__file__).parent.parent.parent / "dsgrid-project-StandardScenarios"
+)
+
 
 def test_register_dimensions_and_mappings(tmp_registry_db):
     src_dir, tmpdir, db_name = tmp_registry_db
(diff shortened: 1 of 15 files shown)


tests/cli/test_registry.py
```diff
@@ -172,3 +176,29 @@ def test_list_project_dimension_query_names(cached_registry):
...
+            "--offline",
+            "registry",
+            "projects",
+            "register",
+            str(project_config),
+            "--log-message",
+            "log",
+        ],
+    )
+    assert result.exit_code == 0
```
elainethale, line 204: I agree that this test isn't problematic per se. How awkward would it be later to replace the standard scenarios submodule with something else?

elainethale (approved):
Let's chat about the implications of having dsgrid-project-StandardScenarios as a submodule. Otherwise this looks good to me.

tests/cli/test_registry.py
daniel-thom: The dilemma is this: What is the best way to get full test coverage of our code? Do we create fake projects that have everything that our real projects have? Or do we just clone those projects in the tests? Making a git submodule is the easiest way to keep the “test” data in sync with the code. It could be a simple clone as long as we track the branch name. We should probably try to register every project that is open source. A secondary benefit is that it provides extra checking on the project repository. We have introduced breakages many times.

elainethale:
After discussing, we're going to add all project repositories as submodules so we can keep them in sync. It will still be important to make filtered databases and add them to the test-data repository in order to test mapping and queries.

Issue #258, opened 2023-10-10
Add optional scaling factor to load_data_lookup 


elainethale:
This will support dGen only specifying one year of timeseries data.

PR #263, opened 2023-10-17
Add optional scaling factor to load_data_lookup

daniel-thom:
Closes GitHub Issue #258

### DIFF
--- a/dsgrid/dataset/dataset_schema_handler_standard.py
+++ b/dsgrid/dataset/dataset_schema_handler_standard.py
@@ -3,6 +3,7 @@ from pathlib import Path
 
 import pyspark.sql.functions as F
 
+from dsgrid.common import SCALING_FACTOR_COLUMN
 from dsgrid.config.dataset_config import DatasetConfig
 from dsgrid.config.simple_models import DimensionSimpleModel
 from dsgrid.dataset.dataset_schema_handler_base import DatasetSchemaHandlerBase
@@ -11,7 +12,7 @@ from dsgrid.dimension.base_models import DimensionType
 from dsgrid.exceptions import DSGInvalidDataset
 from dsgrid.query.models import TableFormatType, ColumnType
 from dsgrid.query.query_context import QueryContext
-from dsgrid.utils.dataset import check_null_value_in_unique_dimension_rows
+from dsgrid.utils.dataset import check_null_value_in_unique_dimension_rows, apply_scaling_factor
 from dsgrid.utils.spark import (
     # create_dataframe_from_pandas,
     read_dataframe,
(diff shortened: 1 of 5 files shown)


dsgrid/dataset/dataset_schema_handler_standard.py
```diff
@@ -66,12 +67,13 @@ def make_project_dataframe(self, project_config):
...
             ld_df = self._convert_time_dimension(ld_df, project_config)
 
-        lk_df = self._remap_dimension_columns(lk_df)
         null_lk_df = self._remap_dimension_columns(null_lk_df)
         ld_df = self._remap_dimension_columns(ld_df)
         pivoted_columns = set(ld_df.columns).intersection(
             self.get_pivoted_dimension_columns_mapped_to_project()
         )
+        if SCALING_FACTOR_COLUMN in ld_df.columns:
+            ld_df = apply_scaling_factor(ld_df, pivoted_columns)
```
elainethale, line 76: This seems like an important line to cover with a test, but codecov says it's not covered yet.

dsgrid/utils/dataset.py
```diff
@@ -139,6 +141,21 @@ def add_column_from_records(df, dimension_records, dimension_name, column_to_add
...
+) -> DataFrame:
+    """Apply the scaling factor to all value columns and then drop the scaling factor column."""
+    for column in value_columns:
+        df = df.withColumn(
+            column,
+            F.when(
+                F.col(scaling_factor_column).isNotNull(),
+                F.col(column) * F.col(scaling_factor_column),
+            ).otherwise(F.col(column)),
+        )
```
elainethale, line 155: When does the scaling factor column from load_data_lookup get joined to the load_data table? I don't see that happening in dataset_schema_handler_standard.py.

elainethale (changes requested):
I'd like to see a test using this feature with a 2-table dataset before merging. 

I think the current logic is fine (NULL == 1.0).

dsgrid/utils/dataset.py
daniel-thom: Lines 63 and 102 of dataset_schema_handler_standard.py

dsgrid/dataset/dataset_schema_handler_standard.py
daniel-thom: I agree. I wrote a unit test for the scaling factor logic but not an integration test with a real dataset. Sorry, I meant to call that out. It takes some time to invent datasets and check things like this. For resstock and comstock we perform independent verification of all the logic that goes into mapping a dataset to a project. The open question is whether we defer this integration test until we have a real dataset or spend time inventing one here.

PR #264, opened 2023-10-17
Allow specification of the database username and password in the CLI

daniel-thom:
This PR allows you to override the username and password that are in your runtime config file. Previously, you could only change the URL and database name on the CLI.


dsgrid/cli/config.py
```diff
@@ -62,7 +62,7 @@ def config():
 def create(
     timings,
     database_name,
-    database_url,
```
elainethale, line 65: Shouldn't this function also add user name and password options?

elainethale (changes requested):
Users should at least be able to specify their user name, and maybe also password when they dsgrid config create.

dsgrid/cli/config.py
daniel-thom: Yeah, good point. Fixed.

elainethale (approved):

PR #266, opened 2023-10-26
Update package infrastructure

daniel-thom:
1. Update the Python package to use a pyproject.toml with hatchling (instead of setup.py). This follows the latest recommendations from https://packaging.python.org/en/latest/tutorials/packaging-projects/#creating-pyproject-toml.
2. Update the Pydantic to version 2. This is a massive change. They completely changed their API. The tool `bump-pydantic` made some changes automatically. I had to make many other changes manually. Several of the updates are really good for us. For example, they have a feature to `discriminate` between serialized subclasses based on a field. Secondly, they have new serialization functions that simplify our logic. Overall, the changes should maintain identical behavior.
3. Update Spark to v3.5.


dsgrid/config/dataset_config.py
```diff
@@ -190,222 +198,325 @@ class GrowthRateType(DSGEnum):
...
+            },
+        ),
+    ]
+    dataset_qualifier_metadata: Annotated[
+        Optional[Union[QuantityModel, GrowthRateModel]],
+        Field(
+            default=QuantityModel(dataset_qualifier_type="quantity"),
+            title="dataset_qualifier_metadata",
+            description="Additional metadata to include related to the dataset_qualifier",
+            discriminator="dataset_qualifier_type",
```
daniel-thom, line 266: This is an example of using a discriminator. I had to slightly reorganize the fields to make it work.

dsgrid/config/dimensions.py
```diff
@@ -169,142 +238,145 @@ def check_name(cls, name):
...
-
-    @staticmethod
-    def _handle_kwargs(**kwargs):
-        exclude = {"cls", "file", "filename"}
-        if "exclude" in kwargs and kwargs["exclude"] is not None:
-            kwargs["exclude"].union(exclude)
-        else:
-            kwargs["exclude"] = exclude
-        return kwargs
+    @field_serializer("cls", "filename")
```
daniel-thom, line 377: This is an example of how we can customize serialization of a field or exclude it.

pyproject.toml
```diff
@@ -1,3 +1,106 @@
+[build-system]
```
daniel-thom, line 1: Note the change here.

elainethale (approved):
Thanks for keeping us up-to-date, @daniel-thom!

PR #272, opened 2023-12-26
Add support for unpivoted tables

daniel-thom:
This PR allows dataset contributors to store their tables in unpivoted format. It also changes the default behavior of queries to convert the final table to unpivoted format. Users have an option to specify pivoted instead.

The PR also includes some minor changes to documentation and error checks for dataset submission.


docs/source/tutorials/create_and_submit_dataset.rst
```diff
@@ -48,7 +48,10 @@ instructions on how to assign values in this file.
       ],
     }
 
-2. Choose a data format as described in :ref:`dataset-formats`. This dataset uses the two-table
+2. Choose a data format as described in :ref:`dataset-formats`. Please note all requirements listed
```
daniel-thom, line 51: @ashreeta I added you as a reviewer to check the dataset submission additions. Please feel free to suggest additional content that would help dataset contributors. Also note the file dataset_formats.rst above.

dsgrid/dataset/dataset_schema_handler_one_table.py
```diff
@@ -50,17 +52,34 @@ def _check_one_table_data_consistency(self):
...
+                self._config.get_table_format_type() == TableFormatType.UNPIVOTED
+                and col == VALUE_COLUMN
+            ):
+                continue
+            elif expected_pivoted_columns is not None and col in expected_pivoted_columns:
                 pivoted_dim_found = True
                 pivoted_cols.add(col)
             else:
-                dimension_types.add(DimensionType.from_column(col))
+                dim_type = DimensionType.from_column(col)
```
daniel-thom, line 78: @ashreeta Note this new check. You may have other suggested checks.

elainethale (commented):
As discussed during our meeting yesterday, it would be helpful to add a few more tests to cover more dataset formats in more situations (e.g., queries).

dsgrid/config/dataset_config.py
```diff
@@ -99,23 +101,11 @@ class InputDatasetType(DSGEnum):
     BENCHMARK = "benchmark"
 
 
-class DataSchemaType(DSGEnum):
+class DataSchemaType(str, Enum):
```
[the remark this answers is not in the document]
daniel-thom: Yes

docs/source/reference/dataset_formats.rst
```diff
@@ -14,13 +14,10 @@ Requirements
...
-3. Tables must be pivoted with the records of one dimension. This is usually much more efficient
-   than a "long-narrow" format with a single "metric value" column. Support could be added if
-   necessary.
-4. Values of dimension columns must be strings. This includes ``model_year`` and ``weather_year``.
-5. Each dimension column name except time must match dsgrid dimension types (geography, sector,
+3. Values of dimension columns must be strings. This includes ``model_year`` and ``weather_year``.
+4. Each dimension column name except time must match dsgrid dimension types (geography, sector,
    subsector, etc.).
-6. The values in each dimension column must match the dataset's dimension records.
+5. The values in each dimension column must match the dataset's dimension records.
```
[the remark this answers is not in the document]
daniel-thom: I added some additional text. Let me know what you think.

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -261,9 +256,65 @@ def _convert_units(self, df, project_metric_records, pivoted_columns):
                 break
 
         dataset_records = self._config.get_dimension(DimensionType.METRIC).get_records_dataframe()
-        df = energy.convert_units(
-            df, pivoted_columns, dataset_records, mapping_records, project_metric_records
+        match self._config.get_table_format_type():
+            case TableFormatType.PIVOTED:
+                df = energy.convert_units_pivoted(
+                    df, value_columns, dataset_records, mapping_records, project_metric_records
+                )
```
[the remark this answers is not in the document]
daniel-thom: This piece of code does not care and has always supported any dimension being pivoted. This is only mapping dataset dimensions to project dimensions. There has been another case that only supports metric. That is where the query asks to aggregate the pivoted dimension, such as with `end_uses_by_fuel_type`. In that example for StandardScenarios we are converting all energy types to MWh and so there is a second unit conversion that must occur. Up until now that code did not support other dimension types as the pivoted type. I tried to address that limitation in this PR (refer to pivoted_table.py). However, I did not remove the exception because it isn’t tested.

dsgrid/dataset/pivoted_table.py
```diff
@@ -30,91 +22,13 @@ class PivotedTableHandler(TableFormatHandlerBase):
     def __init__(self, project_config, dataset_id=None):
         super().__init__(project_config, dataset_id=dataset_id)
```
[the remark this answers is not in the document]
daniel-thom: There are now two classes, PivotedTableHandler and UnpivotedTableHandler, with the same parent class. That parent class has the “deleted” methods.

dsgrid/query/report_peak_load.py
```diff
@@ -34,23 +34,27 @@ def generate(
     ):
         if context.get_table_format_type() == TableFormatType.PIVOTED:
             value_columns = list(context.get_pivoted_columns())
+            group_by_columns = inputs.group_by_columns
         else:
-            # TODO #202: this is TBD
-            # value_columns = ["value"]
-            raise NotImplementedError(f"Unsupported {context.get_table_format_type()=}")
+            value_columns = ["value"]
```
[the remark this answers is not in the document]
daniel-thom: Thanks, fixed in the next push.

tests/simple_standard_scenarios_datasets.py
```diff
@@ -312,5 +319,66 @@ def perform_op_by_electricity(stats, table, name, operation):
     stats[name]["count"] = table.count()
 
 
+def make_unpivoted_datasets():
```
daniel-thom, line 322: This is a substantial behavioral change. After filtering the SS datasets the code will now unpivot the two comstock tables. That gives us two-table and one-table pivoted datasets that we can query in the tests with no changes to that test code.

dsgrid/project.py
```diff
@@ -226,8 +226,37 @@ def process_query(self, context: QueryContext, cached_datasets_dir: Path):
...
+        expected_columns.append(VALUE_COLUMN)
+
+        datasets = {}
+        for dataset_id, path in filenames.items():
+            df = read_dataframe(path)
+            unexpected = sorted(set(df.columns).difference(expected_columns))
+            if unexpected:
+                raise Exception(f"Unexpected columns are present in {dataset_id=} {unexpected=}")
+            datasets[dataset_id] = DatasetExpressionHandler(
+                df.select(*expected_columns), time_columns + dim_columns, [VALUE_COLUMN]
```
lixiliu, line 249: I guess `df.select(*expected_columns)` here makes sure all expected_columns are present in the df. Nothing to do

lixiliu (commented):
80% through

dsgrid/dataset/dataset_schema_handler_standard.py
```diff
@@ -292,16 +291,16 @@ def filter_data(self, dimensions: list[DimensionSimpleModel]):
...
-
-        pivoted_dimension_type = self.get_pivoted_dimension_type()
         pivoted_columns_to_remove = set()
-        for dim in dimensions:
-            column = dim.dimension_type.value
-            if column in lookup_columns:
-                lookup = lookup.filter(lookup[column].isin(dim.record_ids))
-            elif dim.dimension_type == pivoted_dimension_type:
-                pivoted_columns_to_remove = pivoted_columns.difference(dim.record_ids)
+        pivoted_dimension_type = self._config.get_pivoted_dimension_type()
```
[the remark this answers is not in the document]
daniel-thom: They are the same. The other is a shortcut. Perhaps I didn’t change all cases.

dsgrid/project.py
daniel-thom: It sets the column order.

dsgrid/query/query_context.py
```diff
@@ -95,37 +108,72 @@ def get_all_dimension_query_names(self):
...
+                    ColumnType.DIMENSION_QUERY_NAMES | ColumnType.DIMENSION_TYPES,
+                    _,
+                    True,
+                ):
+                    column_names = list(pivoted_columns)
+                    assert pivoted_columns.issubset(
+                        project_config.get_dimension(name).get_unique_ids()
+                    )
+                case (
+                    TableFormatType.PIVOTED | TableFormatType.UNPIVOTED,
```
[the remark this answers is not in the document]
daniel-thom: It protects against the case where a new enum value gets added but is not covered here.

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -245,7 +233,14 @@ def _check_dataset_time_consistency_by_time_array(self, time_dim, time_cols, loa
                 f"unique time array lengths = {len(distinct_ta_counts)}"
             )
 
-    def _convert_units(self, df, project_metric_records, pivoted_columns):
+    def _check_load_data_unpivoted_columns(self, df):
+        logger.info("Check load data unpivoted columns.")
+        if self._config.model.data_schema.table_format.value_column not in df.columns:
```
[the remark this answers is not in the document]
daniel-thom: The check happens at the call site, like this:
```
        match self._config.get_table_format_type():
            case TableFormatType.PIVOTED:
                self._check_load_data_pivoted_columns()
            case TableFormatType.UNPIVOTED:
                self._check_load_data_unpivoted_value_column(self._load_data)
```
Does that address your concern?

docs/source/tutorials/create_and_submit_dataset.rst
```diff
@@ -270,6 +276,12 @@ Records file snippet::
    column. If the dataset is missing data for specific dimension combinations, include a row for
    each combination and set ``id`` to ``null``.
 
+.. warning:: All dimension columns must be strings, including columns that look like numbers, such
```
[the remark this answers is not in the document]
daniel-thom: GitHub may have hidden some required context. Does the full content change your opinion?
```
5. Create ``load_data_lookup.parquet``. The ``id`` column should match the values in
   ``load_data.parquet`` so that a single table can be produced by joining the two tables on that
   column. If the dataset is missing data for specific dimension combinations, include a row for
   each combination and set ``id`` to ``null``.

.. warning:: All dimension columns must be strings, including columns that look like numbers, such
   as ``model_year``.

.. warning:: If your dataset uses FIPS county codes, be sure to not inadvertently drop leading
   zeros.
```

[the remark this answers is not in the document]
lixiliu: Not exactly, the initial issue I had was that it took me a moment to remember that there are two or three types of columns: dimension columns, value columns, columns that are neither (id, scale factor). We don't talk about value columns much, I just thought if we do this, it would be more clear:
```
warning:: All dimension columns except time, including columns that look like numbers, such as ``model_year``, are considered attributes and therefore they must be strings. Conversely, only value columns are numeric as they measure a quantity.
```

dsgrid/query/report_peak_load.py
```diff
@@ -34,23 +35,27 @@ def generate(
     ):
         if context.get_table_format_type() == TableFormatType.PIVOTED:
             value_columns = list(context.get_pivoted_columns())
+            group_by_columns = inputs.group_by_columns
         else:
-            # TODO #202: this is TBD
-            # value_columns = ["value"]
-            raise NotImplementedError(f"Unsupported {context.get_table_format_type()=}")
+            value_columns = [VALUE_COLUMN]
```
lixiliu, line 40: Why not make this else as a check against UNPIVOTED instead? That way, there will never be more than one metric column...

dsgrid/units/energy.py
```diff
@@ -56,11 +59,13 @@ def to_gwh(unit_col, value_col):
     """Convert a column to gWh."""
     return (
         F.when(F.col(unit_col) == KWH, (F.col(value_col) / 1_000_000))
+        .when(F.col(unit_col) == "", F.col(value_col))
```
lixiliu, line 62: redundant

dsgrid/units/energy.py
```diff
@@ -146,6 +287,45 @@ def convert_units(df, columns, from_records, from_to_records, to_unit_records) -
...
+        If None, mapping is not required and from_records contain the units.
+    to_unit_records : DataFrame
+        Metric dimension records for the target IDs
+    """
+    unit_col = "unit"  # must match EnergyEndUse.unit
+    tmp1 = from_records.select("id", unit_col).withColumnRenamed(unit_col, "from_unit")
+    if from_to_records is None:
+        unit_df = tmp1.select("id", "from_unit")
+    else:
+        tmp2 = from_to_records.select("from_id", "to_id")
```
lixiliu, line 314: This assumes from_id has the same unit as to_id, is that always true?

dsgrid/units/energy.py
```diff
@@ -146,6 +287,45 @@ def convert_units(df, columns, from_records, from_to_records, to_unit_records) -
...
+        unit_df = (
+            tmp1.join(tmp2, on=tmp1["id"] == tmp2["from_id"])
+            .selectExpr("to_id AS id", "from_unit")
```
lixiliu, line 317: This assumes from_id has the same unit as to_id, is that always true?

dsgrid/units/energy.py
```diff
@@ -100,23 +107,157 @@ def to_mbtu(unit_col, value_col):
...
 
-def convert_units(df, columns, from_records, from_to_records, to_unit_records) -> DataFrame:
+def from_any_to_any(from_unit_col, to_unit_col, value_col):
+    """Convert a column based on from/to columns."""
+    return (
+        F.when(F.col(from_unit_col) == F.col(to_unit_col), F.col(value_col))
+        .when(F.col(from_unit_col) == "", F.col(value_col))
+        # From KWh
+        .when(
+            (F.col(from_unit_col) == KWH) & (F.col(to_unit_col) == MWH), F.col(value_col) / 1_000
```
lixiliu, line 122: I recommend making more constants to use here to convert between SI prefixes (K, M, G, T) and keep everything in muplication as much as you can (e.g., create THERM_PER_KWH in addition to KWH_PER_THERM and just do that division there), because we'll have to do this many times over when other units are introduced.

elainethale (approved):
Looks good to me. A few last minor comments.

dsgrid/units/energy.py
```diff
@@ -30,93 +71,237 @@ def to_kwh(unit_col, value_col):
...
         .when(F.col(unit_col) == MBTU, F.col(value_col))
+        .when(F.col(unit_col) == "", F.col(value_col))
         .otherwise(None)
     )
 
 
-def convert_units(df, columns, from_records, from_to_records, to_unit_records) -> DataFrame:
+def from_any_to_any(from_unit_col, to_unit_col, value_col):
+    """Convert a column based on from/to columns."""
+    return (
```
lixiliu, line 156: Can we daisy-chain the to_XXX() we already created here or do a match case?
```
 if to_unit_col == MWH:
   return to_mbtu(from_unit_col, value_col)
 ```

lixiliu (approved):
I only reviewed the `units/energy.py` this time, and it looks good! Thanks for refactoring. Minor changes recommended.

dsgrid/units/energy.py
daniel-thom: Nice suggestion. I am very surprised that this worked.

lixiliu: Cool! Wasn't sure that daisy-training would work either.

PR #273, opened 2024-01-05
Add registry path checks

daniel-thom:
This fixes an issue found today where the database was created from a personal computer instead of the HPC and resulted in a confusing error message.

### DIFF
--- a/dsgrid/registry/registry_manager_base.py
+++ b/dsgrid/registry/registry_manager_base.py
@@ -31,7 +31,7 @@ class RegistryManagerBase(abc.ABC):
 
         if not path.exists():
             logger.warning(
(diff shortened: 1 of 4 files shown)


dsgrid/registry/registry_manager_base.py
```diff
@@ -29,6 +29,13 @@ def __init__(self, path, params: RegistryManagerParams):
         self._params = params
         self._db = None
 
+        if not path.exists():
+            logger.warning(
+                "The registry path=%s does not exist. You will able to inspect the registry "
```
elainethale, line 34: nit: "registry data path" instead of "registry path"?

elainethale (approved):
Small typo (perhaps) to fix. Otherwise LGTM

PR #277, opened 2024-01-22
Fix detection of Optional type with a None value

daniel-thom:
Fixes a bug that was causing failures in the registration of dGen, which was mapping some model years to null.

### DIFF
--- a/dsgrid/utils/spark.py
+++ b/dsgrid/utils/spark.py
@@ -8,7 +8,7 @@ import os
 import shutil
 from contextlib import contextmanager
 from pathlib import Path
-from typing import AnyStr, Union, get_origin, get_args
+from typing import AnyStr, Union, get_origin, get_args, Type
 
 import pandas as pd
 import pyspark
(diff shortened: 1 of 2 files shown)


dsgrid/utils/spark.py
```diff
@@ -342,6 +342,33 @@ def models_to_dataframe(models, table_name=None):
...
+        if not types:
+            msg = f"Unhandled Union type: {python_type=} {args=}"
+            raise NotImplementedError(msg)
+        elif len(types) > 1:
+            msg = f"Unhandled Union type: {types=}"
+            raise NotImplementedError(msg)
+        else:
+            python_type = types[0]
+
+    return python_type
```
elainethale, line 369: Are you trying to find the type that this field is when it's not None, so you can set the type of the Spark dataframe column correctly?

tests/test_spark_utils.py
```diff
@@ -18,3 +21,8 @@ def test_try_read_dataframe_valid(tmp_path):
     df = try_read_dataframe(filename)
     assert isinstance(df, DataFrame)
     assert df.collect()[0].a == 1
+
+
+def test_get_type_from_union():
+    assert get_type_from_union(Optional[str]) is str
+    assert get_type_from_union(Optional[DayType]) is str
```
elainethale, line 28: So Optional[X] is the same as Union[X,None]?

dsgrid/utils/spark.py
daniel-thom: yes

tests/test_spark_utils.py
daniel-thom: yes

ashreeta (approved):

PR #278, opened 2024-01-22
Detect inconsistent subset dimensions

daniel-thom:
This fixes an oversight in the registration of a project’s subset dimensions. We were allowing
1. subset dimension record IDs that are not project base dimension record IDs (clearly a bug)
2. base dimension record IDs to be left out of a subset dimension record file. This is disadvantageous in cases where a record ID is not part of any subset, such as `unspecified`. However, I think it is safer to make the project coordinator list these than provide the chance that a record ID is left out accidentally.

Also, this fixes a bug where we were not allowing specification of dataset dimension requirements like this (needed by DECARB buildings):
```
              multi_dimensional: [
                {
                  sector: {
                    base: [
                      "com",
                    ],
                  },
                  metric: {
                    subset: [
                      {
                          name: 'End Uses by Sector',
                          selectors: ['commercial_end_uses'],
                      },
                    ]
                  }
                },
                {
                  sector: {
                    base: [
                      "res",
                    ],
                  },
                  metric: {
                    subset: [
                      {
                        name: 'End Uses by Sector',
                        selectors: ['residential_end_uses'],
                      },
                    ]
                  }
                },
              ],
```


elainethale (changes requested):
The change looks good to me, but can we add a test or two?

daniel-thom:
> The change looks good to me, but can we add a test or two?

The positive case will be covered when we merge the corresponding DECARB project.json5 file. There is a test in `tests/cli/test_registry.py` that registers the project and runs this code. I have observed that the negative tests work because the existing files were improperly defined. I could cover everything if we invested time in creating a test dataset that follows the format of DECARB buildings.

elainethale:
> > The change looks good to me, but can we add a test or two?
> 
> The positive case will be covered when we merge the corresponding DECARB project.json5 file. There is a test in `tests/cli/test_registry.py` that registers the project and runs this code. I have observed that the negative tests work because the existing files were improperly defined. I could cover everything if we invested time in creating a test dataset that follows the format of DECARB buildings.

Hopefully it will be worth doing that soon. Should we start a list of TODOs to remind us of the things we want to test once that's in place?

Issue #257, opened 2023-08-29
Handle exceptions cleanly in all CLI commands


daniel-thom:
In most dsgrid CLI commands we let exceptions go back to the user. An example is that if the user tries to register a project with an invalid dimension, they will receive a large stacktrace in the console. We can handle these in a better way by catching them and reporting simple error messages. Perhaps stacktraces can get written to log files only.

PR #279, opened 2024-01-22
Suppress exceptions in CLI commands

daniel-thom:
Suppress tracebacks in CLI commands for dsgrid exceptions so that it is easier for users to debug problems. (#257)
    
This commit changes the behavior of dsgrid exceptions that get raised in CLI commands. Instead of letting the entire traceback get dumped to the console, the code only displays summary information. This will make it easier for users to figure out what failed.

There is an option to preserve the current behavior and reraise exceptions.


elainethale (approved):
LGTM. Is going through and making sure we're raising the right types of exceptions a next step?

daniel-thom:
> LGTM. Is going through and making sure we're raising the right types of exceptions a next step?

I don’t know that we are raising incorrect exceptions. It would be good to check. It would be especially good to ensure that we are providing clear corrective actions.

ashreeta (approved):

PR #280, opened 2024-01-23
Pre-check dataset_id in register_and_submit_dataset

daniel-thom:
The code was not checking for proper status of the dataset in the project before registering the dataset. This was causing a lot of wasted time.

### DIFF
--- a/dsgrid/registry/project_registry_manager.py
+++ b/dsgrid/registry/project_registry_manager.py
@@ -662,6 +662,14 @@ class ProjectRegistryManager(RegistryManagerBase):
         if not self.has_id(project_id):
             msg = f"{project_id=}"
             raise DSGValueNotRegistered(msg)
+
+        dataset_config = DatasetConfig.load_from_user_path(dataset_config_file, dataset_path)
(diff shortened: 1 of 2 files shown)


dsgrid/registry/project_registry_manager.py
```diff
@@ -764,12 +772,8 @@ def _submit_dataset_and_register_mappings(
     ):
         logger.info("Submit dataset=%s to project=%s.", dataset_id, project_config.config_id)
         self._check_if_not_registered(project_config.config_id)
+        self._raise_if_not_unregistered(project_config, dataset_id)
```
elainethale, line 775: Is this redundant with the line above?

elainethale (approved):
One minor question

dsgrid/registry/project_registry_manager.py
daniel-thom: Line 774 checks if the project_id is registered. Line 775 checks if the dataset is present in the project config and has an unregistered status.

elainethale: I see that now--apologies for missing it the first time!

Issue #274, opened 2024-01-09
Roll hours to support annual hourly (etc.) data that's for the same year, but covering a different time zone (e.g., covers UTC instead of EST)


elainethale:

ashreeta:
@lixiliu I used the code below in pyspark to manually change the column year from 2017 to 2018. I then has to convert from the string type to a timestamp.

from pyspark.sql.functions import col, when, lit, substring, concat

# Assuming your DataFrame is named df
df = df.withColumn(
    "timestamp_est",
    when(col("timestamp_est").startswith("2017"), 
         concat(lit("2018"), substring(col("timestamp_est"), 5, 16))
    ).otherwise(col("timestamp_est"))
)

PR #282, opened 2024-01-24
Ll/time wrap

lixiliu:
Completes JIRA Story DSGRID-
Closes GitHub Issue #274 

## Description
Add "wrap_time_allowed" as a project InputDatasetModel parameter. Via this parameter, project can tell dsgrid whether time-wrapping is allowed when mapping dataset time to project time due to difference in time zone.

Note: when aligning time_interval_type, time-wrapping is automatically applied if the time range spills out to a different year after shifting for the interval type.

dsgrid-test-data PR: https://github.com/dsgrid/dsgrid-test-data/pull/39

If dGen dataset is being registered with CY2018 in UTC, we'll need a Decarb PR as well to add wrap_time_allowed=True to project.json5's dGen input dataset config.

@ashreeta - I think this PR is ready to be tested with dGen's UTC dataset submission to project.

### Graphical explanation of time-wrapping
dsgrid_time_wrapping

## Checklist

- [x] All tests pass

### DIFF
--- a/dsgrid/config/dimensions.py
+++ b/dsgrid/config/dimensions.py
@@ -546,6 +546,14 @@ class DateTimeDimensionModel(TimeDimensionBaseModel):
             },
         ),
     ]
+    wrap_time_allowed: Annotated[
+        Optional[bool],
+        Field(
+            title="wrap_time_allowed",
+            description="Whether to allow dataset time to be wrapped to project time if different",
+            default=False,
+        ),
+    ]
 
     @model_validator(mode="after")
     def check_time_type_and_class_consistency(self):
--- a/dsgrid/config/dimensions.py
+++ b/dsgrid/config/dimensions.py
@@ -546,14 +546,6 @@ class DateTimeDimensionModel(TimeDimensionBaseModel):
             },
         ),
     ]
-    wrap_time_allowed: Annotated[
-        Optional[bool],
-        Field(
-            title="wrap_time_allowed",
-            description="Whether to allow dataset time to be wrapped to project time if different",
-            default=False,
-        ),
-    ]
 
     @model_validator(mode="after")
     def check_time_type_and_class_consistency(self):
--- a/dsgrid/config/time_dimension_base_config.py
+++ b/dsgrid/config/time_dimension_base_config.py
@@ -1,5 +1,6 @@
 import abc
 from datetime import datetime, timedelta
+from typing import Optional
 
 import pandas as pd
 import pyspark.sql.functions as F
@@ -14,7 +15,7 @@ class TimeDimensionBaseConfig(DimensionBaseConfigWithoutFiles, abc.ABC):
     """Base class for all time dimension configs"""
 
     @abc.abstractmethod
-    def check_dataset_time_consistency(self, load_data_df, time_columns: list):
+    def check_dataset_time_consistency(self, load_data_df, time_columns: list[str]):
         """Check consistency of the load data with the time dimension.
 
         Parameters
@@ -30,7 +31,7 @@ class TimeDimensionBaseConfig(DimensionBaseConfigWithoutFiles, abc.ABC):
         """
 
     @abc.abstractmethod
-    def build_time_dataframe(self, model_years: list = None):
+    def build_time_dataframe(self, model_years: Optional[list[int]] = None):
         """Build time dimension as specified in config in a spark dataframe.
 
         Parameters
@@ -61,8 +62,8 @@ class TimeDimensionBaseConfig(DimensionBaseConfigWithoutFiles, abc.ABC):
         self,
         df,
         project_time_dim,
-        model_years: list = None,
-        value_columns: set = None,
+        model_years: Optional[list[int]] = None,
+        value_columns: Optional[set[str]] = None,
         wrap_time_allowed: bool = False,
     ):
         """Convert input df to use project's time format and time zone.
@@ -138,7 +139,7 @@ class TimeDimensionBaseConfig(DimensionBaseConfigWithoutFiles, abc.ABC):
         return df.withColumnRenamed(time_cols[0], self.model.dimension_query_name)
 
     @abc.abstractmethod
-    def get_time_ranges(self, model_years: list = None):
+    def get_time_ranges(self, model_years: Optional[list[int]] = None):
         """Return time ranges with timezone applied.
 
         Parameters
@@ -175,7 +176,7 @@ class TimeDimensionBaseConfig(DimensionBaseConfigWithoutFiles, abc.ABC):
         """
 
     @abc.abstractmethod
-    def list_expected_dataset_timestamps(self, model_years: list = None):
+    def list_expected_dataset_timestamps(self, model_years: Optional[list[int]] = None):
         """Return a list of the timestamps expected in the load_data table.
 
         Parameters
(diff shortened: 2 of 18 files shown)


tests/test_project_time.py
```diff
@@ -78,38 +78,43 @@ def test_convert_to_project_time(registry_mgr):
...
-        comstock_time_dim, project_time_dim, wrap_time=False, expect_error=True
-    )
-    compare_time_conversion(
-        resstock_time_dim, project_time_dim, wrap_time=True, expect_error=False
-    )
-    compare_time_conversion(
-        comstock_time_dim, project_time_dim, wrap_time=True, expect_error=False
-    )
+
+    # [2.1] test time-wrap when dataset and project have different time interval type
```
daniel-thom, line 82: It looks like you could fairly easily split these into different tests because the setup is simple. Then the comments become test names. The main advantage is that you don’t have to worry about resetting state. Not sure if it makes sense, but you could also parameterize things like time zone and wrap_time.

dsgrid/config/dimensions.py
```diff
@@ -546,6 +546,14 @@ class DateTimeDimensionModel(TimeDimensionBaseModel):
             },
         ),
     ]
+    wrap_time_allowed: Annotated[
+        Optional[bool],
```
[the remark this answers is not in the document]
elainethale: I would also add that Optional[bool] is exceptionally confusing? (How is None different from False?)

dsgrid/config/dimensions.py
```diff
@@ -546,6 +546,14 @@ class DateTimeDimensionModel(TimeDimensionBaseModel):
             },
         ),
     ]
+    wrap_time_allowed: Annotated[
```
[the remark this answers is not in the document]
elainethale: I agree. Probably the project coordinator should indicate whether they want to wrap time for particular datasets.

dsgrid/config/time_dimension_base_config.py
```diff
@@ -181,10 +181,12 @@ def list_expected_dataset_timestamps(self, model_years=None):
...
         """
 
-    def _convert_time_to_project_time_interval(self, df, project_time_dim=None, wrap_time=True):
+    def _convert_time_to_project_time_interval(self, df, project_time_dim=None):
         """Shift time to match project time based on TimeIntervalType
-        If time range spills over into another year after time interval alignment,
+        If wrap_time_allowed in project_time_dim:
+        - If time range spills over into another year after time interval alignment,
         the time range will be wrapped around so it's bounded within the year
+        - Time will also wrapped if dataset has a different time zone than project
```
elainethale, line 189: Does this mean that the exact same function can be used to handle period-ending/period-beginning and different base time zones? (Neat!)

dsgrid/config/dimensions.py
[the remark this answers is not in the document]
elainethale: But looking at more of the code below, I don't think the project coordinator should have to flip any switches to wrap time just for period-beginning/period-ending. Maybe there should be a project config that specifies a maximum amount of time-wrapping (in number of timestamps and/or hours)?

tests/test_project_time.py
```diff
@@ -78,38 +78,43 @@ def test_convert_to_project_time(registry_mgr):
...
+    # expect no error with original dataset and project time configs
+    compare_time_conversion(resstock_time_dim, project_time_dim, expect_error=False)
+    compare_time_conversion(comstock_time_dim, project_time_dim, expect_error=False)
     compare_time_conversion(
         tempo_time_dim, project_time_dim, df=tempo_data_mapped_time, expect_error=False
     )
+    project_time_dim.model.wrap_time_allowed = False
+    compare_time_conversion(resstock_time_dim, project_time_dim, expect_error=True)
+    compare_time_conversion(comstock_time_dim, project_time_dim, expect_error=True)
```
elainethale, line 91: nit: Add optional kwarg that provides a text snippet that should be in the expected error?

dsgrid/config/dimensions.py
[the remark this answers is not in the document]
daniel-thom: At risk of being overly pedantic, None and False are different. None means “not set.” Some functions may want to know if a boolean variable is True, false, or not set.

Secondly, while I would never do this, I’ve seen it done around here:
```
if a == False
```
Thirdly, there might be some case where you want to compare two values.
```
a = False
b = None
a != b
```

dsgrid/config/dimensions.py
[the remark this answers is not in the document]
lixiliu: Good suggestion to move this parameter to InputDatasetModel. I think it makes sense to allow time wrap automatically if it's only needed to align time interval type and have "wrap_time_allowed" to handle the particular case where the time zones are misaligned? It might be too confusing for the coordinator to limit the max amount of time-wrapping, at least I can't think of a concrete use case for that.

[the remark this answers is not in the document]
elainethale: That sounds good to me.

ashreeta (approved):
Hi I had a look at all the changes but fail to understand where exactly the timestamps are being converted? The only change seems to be the addition of this time_wrap into the project config? Can you direct me to which fictions exactly handle time-wrapping?

tests/test_project_time.py
lixiliu: The error msg is here: https://github.com/dsgrid/dsgrid/blob/ae76c0814d59652ce213f7a563bbff51238268f8/tests/test_project_time.py#L182-L184

lixiliu:
> Hi I had a look at all the changes but fail to understand where exactly the timestamps are being converted? The only change seems to be the addition of this time_wrap into the project config? Can you direct me to which functions handle time-wrapping?

Hi @ashreeta - the dataset_schema_handler uses a class function of the time dimension to [convert the dataset's time col according to project_time_dim](https://github.com/dsgrid/dsgrid/blob/9f418463ac59f0dd59ffd147649b4ad798a2670f/dsgrid/dataset/dataset_schema_handler_base.py#L421-L426). Each time_dimension_config has its own convert_dataframe() function. For the [datetime_dimension_config](https://github.com/dsgrid/dsgrid/blob/9f418463ac59f0dd59ffd147649b4ad798a2670f/dsgrid/config/date_time_dimension_config.py#L93-L95), `self.convert_dataframe()` is informed by `self._convert_time_to_project_time_interval()`, which is defined in the [time_dimension_base_config here](https://github.com/dsgrid/dsgrid/blob/9f418463ac59f0dd59ffd147649b4ad798a2670f/dsgrid/config/time_dimension_base_config.py#L184-L206). The time-wrapping happens here.

dsgrid/config/time_dimension_base_config.py
```diff
@@ -296,7 +359,13 @@ def _apply_time_wrap(df, project_time_dim):
 
         return df
 
-    def _build_time_ranges(self, time_ranges, str_format, model_years=None, tz=None):
+    def _build_time_ranges(
+        self,
+        time_ranges: TimeRangeModel,
+        str_format: str,
+        model_years: list = None,
```
daniel-thom, line 366: ditto

tests/test_project_time.py
```diff
@@ -21,52 +21,64 @@
...
+def test_no_unexpected_timezone():
+    for tzo in TimeZone:
+        assert (
+            tzo.is_standard() + tzo.is_prevailing() == 1
+        ), f"{tzo} can either be prevailing or standard"
+
+
+def test_build_time_dataframe(project, resstock, comstock):
+    project_time_dim = project.config.get_base_dimension(DimensionType.TIME)
+    # alteratively, project_time_dim = resstock._handler._project_time_dim
```
daniel-thom, line 68: We shouldn’t store that attribute in the handler. It can retrieve the time dim on demand.

daniel-thom (commented):
Looks good, several nits

dsgrid/config/time_dimension_base_config.py
```diff
@@ -258,10 +295,34 @@ def _apply_time_wrap(df, project_time_dim):
...
+        if difference == "left":
+            return dataset_time.difference(project_time)
+
+        if difference == "right":
+            return project_time.difference(dataset_time)
+
+        if difference == "symmetric":
+            return dataset_time.symmetric_difference(project_time)
+
+        raise ValueError(f"Unsupport function input {difference=}")
```
elainethale, line 308: nit: "Unsupported"

elainethale (approved):

tests/test_project_time.py
daniel-thom: I don’t care about tests accessing private methods. That’s often the best way to unit-test the code. I am suggesting that the dataset handler doesn’t need to store the project time dimension as an attribute. It feels odd keep that reference. The handler can retrieve that dimension config object from the project_config whenever it needs it. It’s a simple change but doesn’t have to be done here.

daniel-thom: I think it’s best to remove this comment. It is currently identical to the line above it.

daniel-thom (approved):

PR #283, opened 2024-01-30
Fix generation and handling of dimension assocations

daniel-thom:


tests/test_project.py
```diff
@@ -38,8 +38,8 @@ def test_project_load(cached_registry):
     assert len(records) == 1
     assert records[0].id == "all_subsectors"
 
-    table = project.config.load_dimension_associations(DATASET_ID, DimensionType.METRIC)
-    assert table.count() > 0
+    # table = project.config.load_dimension_associations(DATASET_ID, DimensionType.METRIC)
```
daniel-thom, line 41: need to remove this

dsgrid/dataset/dataset_schema_handler_one_table.py
```diff
@@ -110,21 +115,15 @@ def _check_one_table_data_consistency(self):
...
         time_cols = set(self._get_time_dimension_columns())
         pivoted_cols = set(self._config.get_pivoted_dimension_columns())
         exclude = time_cols.union(pivoted_cols)
         if self._config.get_table_format_type() == TableFormatType.UNPIVOTED:
             exclude.add(VALUE_COLUMN)
         dim_cols = [x for x in self._load_data.columns if x not in exclude]
         df = self._load_data.select(*dim_cols).distinct()
-
-        dim_table = self._remap_dimension_columns(df, True).distinct()
-        check_null_value_in_unique_dimension_rows(dim_table, exclude_columns=exclude)
```
daniel-thom, line 125: @lixiliu In this file we were previously not explicitly checking for null values during dataset registration (now we are), but were checking for them during dataset submission (after mapping to project). I removed the latter check because, when we map stacked dimensions, we first filter out all cases of `to_id == NULL`. So then it should not be possible that we have null values, right? Please confirm. If you want to keep the check for safety, I can do so.

dsgrid/utils/scratch_dir_context.py
```diff
@@ -0,0 +1,49 @@
...
+        self._paths.append(path)
+
+    def list_tracked_paths(self) -> list[Path]:
+        """Return a list of paths being tracked."""
+        return self._paths[:]
+
+    def get_temp_filename(self) -> Path:
+        """Return a temporary filename based in the scratch directory."""
+        with NamedTemporaryFile(dir=self._scratch_dir) as f:
+            return Path(f.name)
```
elainethale, line 32: How does this get closed/deleted? (Does it?)

dsgrid/registry/project_registry_manager.py
```diff
@@ -1074,74 +1063,67 @@ def _check_dataset_base_to_project_base_mappings(
...
+        config: ProjectConfig,
+        dataset_id: str,
+        context: ScratchDirContext,
+    ):
+        logger.info("Make dimension association table for %s", dataset_id)
+        path = context.get_temp_filename()
+        df = config.make_dimension_association_table(dataset_id, context)
+        df = write_dataframe_and_auto_partition(df, path)
+        context.add_tracked_path(path)
+        logger.info("Wrote dimension associations for dataset %s", dataset_id)
```
elainethale, line 1125: nit: Tell the user where the dimension associations were written to?

elainethale (approved):
LGTM. Minor comments. I can't easily review for "correctness," other than to verify that tests are all passing on GitHub.

dsgrid/cli/common.py
```diff
@@ -73,6 +73,18 @@ def handle_dsgrid_exception(ctx, func, *args, **kwargs):
...
 
+def handle_scratch_dir(*args):
+    """Handle the user input for scratch_dir. If a path is passed, ensure it exists."""
+    val = args[2]
+    if val is None:
+        return val
+    path = Path(val)
+    if not path.exists:
+        msg = f"scratch-dir={path} does not exist"
+        raise ValueError(msg)
```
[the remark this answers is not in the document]
daniel-thom: This is actually the base directory for the scratch directory. The CLI help is
```
    help="Temporary directory for dsgrid. Must be accessible on all compute nodes. Defaults to "
     "the current directory.”,
```
At runtime the code will create `./__dsgrid_scratch__` and then delete it when done.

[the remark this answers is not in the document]
daniel-thom: Perhaps I should include the words `base directory` in the help

dsgrid/registry/project_registry_manager.py
daniel-thom: I would say no. We are going to delete it when finished. Do we need an option to keep these around? If that is needed then perhaps we need specific CLI commands to create project and dataset dimension association files.

dsgrid/utils/scratch_dir_context.py
```diff
@@ -0,0 +1,49 @@
...
+        """Return the scratch directory."""
+        return self._scratch_dir
+
+    def add_tracked_path(self, path: Path) -> None:
+        """Add tracking of a path in the scratch directory."""
+        self._paths.append(path)
+
+    def list_tracked_paths(self) -> list[Path]:
+        """Return a list of paths being tracked."""
+        return self._paths[:]
```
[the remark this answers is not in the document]
daniel-thom: That is a loaded question. :) I clearly prefer what I wrote. It says, “return a full slice (not a copy) of the list.” It is equivalent to calling `list`. Is calling `list` more clear to others?

dsgrid/utils/scratch_dir_context.py
daniel-thom: It is deleted as soon as this code block exits. It only serves to generate a unique name in the given path. The expectation is that callers are going to use the name. It would not work if we were doing multiprocessing, as there could be collisions (we are obviously not going to do that). I should note that.

dsgrid/cli/common.py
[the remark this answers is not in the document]
elainethale: Maybe include words like, "parent directory for a temporary directory named __dsgrid_scratch__"?

dsgrid/utils/scratch_dir_context.py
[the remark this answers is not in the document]
elainethale: I kind of understood what [:] should do, but I still looked it up because it was unfamiliar to me ... Not sure how @lixiliu or @ashreeta feel about it.

daniel-thom:
I just completed full resubmission of all StandardScenarios datasets and derived datasets with this branch. I also created a registry with DECBARB and submitted the buildings dataset. The last few commits were to fix bugs found during that process. We now have much higher confidence in our handling of the various dataset formats.

dsgrid/dataset/dataset_schema_handler_one_table.py
lixiliu: Yeah, if we already check to make sure that no null value exists for each dimension in the dataset (other than id) during registration, I would not expect nulls to be there in the dimension combinations.

dsgrid/dataset/dataset_schema_handler_standard.py
```diff
@@ -49,13 +49,16 @@ def check_consistency(self):
...
 
-    @track_timing(timer_stats_collector)
-    def get_unique_dimension_rows(self):
-        """Get distinct combinations of remapped dimensions, including id.
-        Check each col in combination for null value."""
-        dim_table = self._remap_dimension_columns(self._load_data_lookup, False).distinct()
-        check_null_value_in_unique_dimension_rows(dim_table)
-        return dim_table
+    def make_dimension_association_table(self) -> DataFrame:
+        df = self._remap_dimension_columns(self._load_data_lookup, False).distinct()
```
lixiliu, line 53: Do we need to filter out id and scaling factor here?

dsgrid/query/query_context.py
```diff
@@ -68,7 +69,18 @@ def _consolidate_pivoted_metadata(self):
                     "Please set the output format to 'unpivoted'."
                 )
 
-    def get_pivoted_columns(self, dataset_id=None):
+    def get_value_columns(self) -> set[str]:
+        """Return the value columns in the final dataset."""
+        match self.get_table_format_type():
+            case TableFormatType.PIVOTED:
+                return self.get_pivoted_columns()
```
lixiliu, line 76: Looks like get_pivoted_columns() uses get_dimension_column_names(), which uses _dataset_metadata(). I didn't trace this all the way down but can I assume the _dataset_metadata() stays up to date with what's available from the dataset, like if the dataset has been filtered?

dsgrid/registry/project_registry_manager.py
```diff
@@ -1007,22 +1008,11 @@ def _submit_dataset(
         if os.environ.get("__DSGRID_SKIP_CHECK_DATASET_TO_PROJECT_MAPPING__") is not None:
             logger.warning("Skip dataset-to-project mapping checks")
         else:
-            # This operation can be very problematic if there are many executors and runs much
-            # faster with a single executor on a single core.
-            # The dynamic allocation settings will only take effect if the worker was started
-            # with spark.shuffle.service.enabled=true
```
lixiliu, line 1013: not a problem anymore?

dsgrid/dataset/dataset_schema_handler_standard.py
daniel-thom: Those will get ignored later, but we should do that to minimize the size. I’ll add a function to do this for both handlers.

dsgrid/registry/project_registry_manager.py
daniel-thom: nope

dsgrid/query/query_context.py
daniel-thom: yes

dsgrid/registry/project_registry_manager.py
```diff
@@ -1074,74 +1063,67 @@ def _check_dataset_base_to_project_base_mappings(
...
+
+    def _handle_dimension_association_errors(
+        self, diff, mapped_dataset_table, dataset_config, project_id
+    ):
         dataset_id = dataset_config.config_id
-        assoc_table = project_config.load_dimension_associations(
-            dataset_id, pivoted_dimension=pivoted_dimension, try_load_cache=False
+        out_file = f"{dataset_id}__{project_id}__missing_dimension_record_combinations.csv"
+        diff.cache()
+        diff.write.options(header=True).mode("overwrite").csv(out_file)
```
lixiliu, line 1085: Nice to export the entire missing association

dsgrid/utils/scratch_dir_context.py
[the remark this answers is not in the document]
lixiliu: Same, not familiar with the [:] notation but the output type hint says it's a list, so that's clear. I don't have a preference here.

lixiliu (approved):
Looks good to me!

PR #286, opened 2024-02-12
Fix registration of projects with duplicate dimensions

daniel-thom:
The code was not handling registration of DECARB after registering StandardScenarios because of some duplicate dimensions.

### DIFF
--- a/dsgrid/registry/dimension_mapping_registry_manager.py
+++ b/dsgrid/registry/dimension_mapping_registry_manager.py
@@ -69,19 +69,27 @@ class DimensionMappingRegistryManager(RegistryManagerBase):
         self._dimension_mgr = val
 
     def _replace_duplicates(self, config: DimensionMappingsConfig):
+        def make_key(model):
+            return (
+                model.from_dimension.dimension_id,
+                model.to_dimension.dimension_id,
+                model.file_hash,
+            )
+
         hashes = {}
         for model in self.db.iter_models(all_versions=True):
-            hashes[model.file_hash] = model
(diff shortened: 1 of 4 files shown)


dsgrid/registry/dimension_registry_manager.py
```diff
@@ -237,9 +248,8 @@ def _register(self, config, submitter, log_message, context):
             registration.version,
         )
 
-        context.add_ids(RegistryType.DIMENSION, dimension_ids, self)
-        dimension_ids.extend(existing_ids)
```
daniel-thom, line 241: This was the bug. The function was written to not care about order but the project registry manager expected them to be in order.

dsgrid/registry/dimension_mapping_registry_manager.py
```diff
@@ -69,19 +69,27 @@ def dimension_manager(self, val: DimensionRegistryManager):
...
+                model.file_hash,
+            )
+
         hashes = {}
         for model in self.db.iter_models(all_versions=True):
-            hashes[model.file_hash] = model
+            key = make_key(model)
+            if key in hashes:
+                msg = f"Bug: the same file_hash exists in multiple mappings: {model.mapping_id} {key}"
+                raise Exception(msg)
```
elainethale, line 84: What is the user supposed to do if they hit this exception?

elainethale: And why is this a plain Exception? (When do we use plain Exceptions and when do we use dsgrid-named exceptions?)

PR #288, opened 2024-02-16
Add support for OneWeekdayDayAndOneWeekendDayPerMonthByHourType

daniel-thom:
This allows us to register the TEMPO dataset for DECARB.

The tests use a dataset in https://github.com/dsgrid/dsgrid-test-data/pull/40

### DIFF
--- a/dsgrid/config/representative_period_time_dimension_config.py
+++ b/dsgrid/config/representative_period_time_dimension_config.py
@@ -2,9 +2,10 @@ import abc
 import calendar
 import logging
 from datetime import datetime, timedelta
+from typing import Type
+
 from pyspark.sql.types import StructType, StructField, IntegerType
 import pyspark.sql.functions as F
-
 import pandas as pd
 
 from dsgrid.dimension.time import (
@@ -14,7 +15,10 @@ from dsgrid.dimension.time import (
     LeapDayAdjustmentType,
 )
 from dsgrid.exceptions import DSGInvalidDataset
-from dsgrid.time.types import OneWeekPerMonthByHourType
+from dsgrid.time.types import (
+    OneWeekPerMonthByHourType,
+    OneWeekdayDayAndOneWeekendDayPerMonthByHourType,
+)
 from dsgrid.utils.timing import track_timing, timer_stats_collector
 from dsgrid.utils.spark import get_spark_session
 from .dimensions import RepresentativePeriodTimeDimensionModel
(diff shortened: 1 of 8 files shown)


dsgrid/config/representative_period_time_dimension_config.py
```diff
@@ -196,6 +211,38 @@ def check_dataset_time_consistency(self, expected_timestamps, load_data_df, time
         DSGInvalidDataset
 
         """
+        logger.info("Check %s dataset time consistency.", self.__class__.__name__)
+        actual_timestamps = []
+        for row in load_data_df.select(*time_columns).distinct().sort(*time_columns).collect():
```
daniel-thom, line 216: Logic is moved, unchanged. Also, this could be pure spark and faster, but it likely doesn’t matter because the number of rows is small.

dsgrid/time/types.py
```diff
@@ -24,10 +25,30 @@ class Season(DSGEnum):
 
 # The types below represent the timestamps that exist as columns in all datasets.
 
-DatetimeTimestampType = namedtuple("DatetimeTimestampType", ["timestamp"])
 
-AnnualTimestampType = namedtuple("AnnualTimestampType", ["time_year"])
+class DatetimeTimestampType(NamedTuple):
```
daniel-thom, line 29: The change from namedtuple to NamedTuple allows type annotations to work on the fields.

dsgrid/config/representative_period_time_dimension_config.py
```diff
@@ -128,13 +138,14 @@ def convert_dataframe(
                     expr = f"{func}(local_time) AS {col}"
                     if col == "day_of_week":
                         expr = f"mod(dayofweek(local_time)+7-2, 7) AS {col}"
+                    elif col == "is_weekday":
```
daniel-thom, line 141: This is the only real change related to mapping of the times.

lixiliu: Did you check out the comment at the top of the func? I think you need this:
```
f"(mod(dayofweek(local_time)+7-2, 7) < 5) AS {col}"
```

# in spark.dayofweek: 1=Sunday, 7=Saturday
# dsgrid uses python standard library (same for pandas), which has day_of_week: 0=Monday, 6=Sunday
# the mapping is: python.dt.day_of_week = [(i+7-2)%7 for i in spark.dayofweek]

dsgrid/time/types.py
```diff
@@ -24,10 +25,30 @@ class Season(DSGEnum):
...
+    """Single column with only year."""
+
+    time_year: int
+
+
+class OneWeekPerMonthByHourType(NamedTuple):
+    """Columns of representative time with one week per month."""
+
+    month: int
+    day_of_week: int
```
elainethale, line 45: Per Lixi's comment above, should we clarify the interpretation of this integer?

dsgrid/config/representative_period_time_dimension_config.py
daniel-thom: This is totally my bad. I read through all of this and the Spark documentation. I intended to use this function (https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.weekday.html), but I guess I copy/pasted the line above and forgot to swap the functions. We should use `weekday` everywhere because it lines up with Python.

tests/test_representative_time.py
```diff
@@ -0,0 +1,130 @@
...
+    project_time_config = make_date_time_config()
+    mapped_df = config.convert_dataframe(df, project_time_config)
+    timestamps = mapped_df.select("timestamp").distinct().sort("timestamp").collect()
+    zi = ZoneInfo("EST")
+    est_timestamps = [x.timestamp.astimezone(zi) for x in timestamps]
+    start = datetime(year=2018, month=1, day=1, tzinfo=zi)
+    resolution = timedelta(hours=1)
+    expected_timestamps = [start + i * resolution for i in range(8760)]
+    assert est_timestamps == expected_timestamps
+    assert mapped_df.filter(f"{VALUE_COLUMN} IS NULL").rdd.isEmpty()
```
elainethale, line 130: In this test, could we check that the peak time is happening at the expected timestamp?

elainethale (approved):
One suggested additional testing check we could perform. Otherwise LGTM.

elainethale (approved):

PR #289, opened 2024-03-05
Dt/fix skew

daniel-thom:
This PR addresses an issue we experienced mapping the DECARB buildings dataset to project. The disaggregation of region to county caused a major issue where one Spark executor thread got stuck, seemingly indefinitely. A message like this was repeated continually. It appears to be caused by data skew, though the imbalance didn't seem too severe. The fact that this dataset uses the unpivoted format likely made the problem much worse.

I used a variation of what online sources call a "salting technique” to unblock the Spark workers. I basically just repartitioned and serialized the dataset immediately after mapping the geography dimension. The logic repartitions on the first value column, which is hopefully sufficiently random to always result in balanced partitions.

### DIFF
--- a/dsgrid/dataset/dataset_schema_handler_one_table.py
+++ b/dsgrid/dataset/dataset_schema_handler_one_table.py
@@ -180,6 +180,7 @@ class OneTableDatasetSchemaHandler(DatasetSchemaHandlerBase):
             # There is currently no case that needs model years or value columns.
             ld_df = self._convert_time_dimension(ld_df, project_config)
 
+        # TODO: This might need to handle data skew in the future.
         ld_df = self._remap_dimension_columns(ld_df, True)
         value_columns = set(ld_df.columns).intersection(self.get_value_columns_mapped_to_project())
         ld_df = self._apply_fraction(ld_df, value_columns)
@@ -212,7 +213,11 @@ class OneTableDatasetSchemaHandler(DatasetSchemaHandlerBase):
             ld_df = self._convert_time_dimension(ld_df, project_config)
 
         ld_df = self._remap_dimension_columns(
-            ld_df, True, filtered_records=context.get_record_ids()
+            ld_df,
+            True,
+            filtered_records=context.get_record_ids(),
+            handle_data_skew=True,
+            scratch_dir_context=context.scratch_dir_context,
         )
         value_columns = set(ld_df.columns).intersection(self.get_value_columns_mapped_to_project())
         ld_df = self._apply_fraction(ld_df, value_columns)
@@ -220,6 +225,12 @@ class OneTableDatasetSchemaHandler(DatasetSchemaHandlerBase):
             DimensionType.METRIC
         ).get_records_dataframe()
         ld_df = self._convert_units(ld_df, project_metric_records, value_columns)
+        # print("start write")
+        # breakpoint()
+        # ld_df.write.mode("overwrite").parquet("tmp.parquet")
+        # spark = get_spark_session()
+        # ld_df = spark.read.load("tmp.parquet")
+        # print("done with read back")
         if not convert_time_before_project_mapping:
             m_year_df = context.try_get_record_ids_by_dimension_type(DimensionType.MODEL_YEAR)
             if m_year_df is None:
--- a/dsgrid/query/query_submitter.py
+++ b/dsgrid/query/query_submitter.py
@@ -2,12 +2,15 @@ import abc
 import logging
 import shutil
 from pathlib import Path
+from typing import Optional
 from zipfile import ZipFile
 
+from dsgrid.utils.scratch_dir_context import ScratchDirContext
 from dsgrid.dataset.table_format_handler_factory import make_table_format_handler
 from dsgrid.dimension.base_models import DimensionCategory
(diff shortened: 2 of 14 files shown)


dsgrid/utils/dataset.py
```diff
@@ -208,3 +209,56 @@ def remove_invalid_null_timestamps(df, time_columns, stacked_columns):
...
+    # Using a variation of what online sources call a "salting technique" solves the issue.
+    # Apply the technique to mappings that will cause an explosion of rows.
+    # Note that this probably isn't needed in all cases and we may need to adjust in the
+    # future.
+    # The case with buildings was particularly severe because it is an unpivoted dataset.
+
+    # Note: log messages below are checked in the tests.
+    if mapping_type in {
+        DimensionMappingType.ONE_TO_MANY_DISAGGREGATION,
+        DimensionMappingType.ONE_TO_MANY_ASSIGNMENT,
```
daniel-thom, line 237: Based on team meeting: Only do this for ONE_TO_MANY_DISAGGREGATION. And add a column of random floats.

dsgrid/dataset/dataset_schema_handler_one_table.py
```diff
@@ -212,14 +213,24 @@ def make_project_dataframe_from_query(self, context: QueryContext, project_confi
...
         project_metric_records = project_config.get_base_dimension(
             DimensionType.METRIC
         ).get_records_dataframe()
         ld_df = self._convert_units(ld_df, project_metric_records, value_columns)
+        # print("start write")
+        # breakpoint()
+        # ld_df.write.mode("overwrite").parquet("tmp.parquet")
+        # spark = get_spark_session()
+        # ld_df = spark.read.load("tmp.parquet")
+        # print("done with read back")
```
elainethale, line 233: Delete before merge? Or convert prints to logger.debug?

dsgrid/query/query_submitter.py
```diff
@@ -309,11 +314,17 @@ def submit(
             Raised if the model defines a project version
         DSGInvalidQuery
             Raised if the query is invalid
-
         """
-        return self._run_query(
-            model, load_cached_table, persist_intermediate_table, zip_file=zip_file, force=force
-        )[0]
+        scratch_dir = scratch_dir or DsgridRuntimeConfig.load().get_scratch_dir()
```
elainethale, line 318: Do we have documentation for where the user should specify their scratch directory?

dsgrid/registry/project_registry_manager.py
```diff
@@ -1126,7 +1126,6 @@ def _make_dimension_associations(
         df = config.make_dimension_association_table(dataset_id, context)
         path = context.get_temp_filename(suffix=".parquet")
         df = write_dataframe_and_auto_partition(df, path)
-        context.add_tracked_path(path)
```
elainethale, line 1129: This seems like a non sequitur

dsgrid/utils/dataset.py
```diff
@@ -208,3 +209,56 @@ def remove_invalid_null_timestamps(df, time_columns, stacked_columns):
...
+
+@track_timing(timer_stats_collector)
+def repartition_if_needed_by_mapping(
+    df: DataFrame,
+    value_columns: list[str],
+    mapping_type: DimensionMappingType,
+    dimension_type: DimensionType,
+    scratch_dir_context: ScratchDirContext,
+) -> DataFrame:
+    """Repartition the dataframe if the mapping will cause data skew."""
```
elainethale, line 222: nit: "might cause data skew."

dsgrid/utils/spark.py
```diff
@@ -632,7 +632,6 @@ def open_file(index):
         if not f_out.closed:
             f_out.close()
 
-    context.add_tracked_path(csv_dir)
```
elainethale, line 635: Same as above--is this deletion intended?

tests/test_dataset_utils.py
```diff
@@ -271,6 +277,52 @@ def test_apply_scaling_factor():
...
+    try:
+        with caplog.at_level(logging.INFO):
+            df = repartition_if_needed_by_mapping(
+                df,
+                value_columns,
+                DimensionMappingType.ONE_TO_MANY_DISAGGREGATION,
+                DimensionType.GEOGRAPHY,
+                context,
+            )
+            assert "DSGRID_SKIP_MAPPING_SKEW_REPARTITION is true" in caplog.text
```
elainethale, line 307: Also assert that "Completed repartition" is not in caplog.text?

dsgrid/query/query_submitter.py
daniel-thom:
```
dsgrid —help
…
  -s, --scratch-dir TEXT    Base directory for dsgrid temporary directories.
                            Must be accessible on all compute nodes. Defaults
                            to the current directory.
```

dsgrid/registry/project_registry_manager.py
daniel-thom: Kinda. `context.get_temp_filename` previously did not add tracking of the returned filename. I found it very easy to forget to add the path afterwards. So, I combined the steps into one and removed this call.

elainethale (approved):

PR #291, opened 2024-03-25
Fix handling of single/multi dimension requirements

daniel-thom:
Fixes two issues:
1. We were not properly checking for redundancy in the single and multi-dimensional requirements in the project config.
2. We did not handle the case where one multi-dimensional requirement has a partial overlap in dimensions with another requirement. We can assume that any missing dimension should use the base dimension records.

Requires https://github.com/dsgrid/dsgrid-project-DECARB/pull/22

### DIFF
--- a/dsgrid/config/project_config.py
+++ b/dsgrid/config/project_config.py
@@ -462,12 +462,14 @@ class RequiredDimensionsModel(DSGBaseModel):
                         )
                         raise ValueError(msg)
                     dims.append(field)
+
             if len(dims) < 2:
                 msg = (
                     "A multi_dimensional dimension requirement must contain at least two "
                     f"dimensions: {item}"
                 )
                 raise ValueError(msg)
+
             dim_combo = tuple(sorted(dims))
             if dim_combo not in dim_combos:
                 for other in dim_combos:
--- a/dsgrid/config/project_config.py
+++ b/dsgrid/config/project_config.py
@@ -395,6 +395,10 @@ class RequiredDimensionRecordsByTypeModel(DSGBaseModel):
     subset: list[RequiredSubsetDimensionRecordsModel] = []
     supplemental: list[RequiredSupplementalDimensionRecordsModel] = []
 
+    def defines_dimension_requirement(self) -> bool:
+        """Returns True if the model defines a dimension requirement."""
+        return bool(self.base) or bool(self.subset) or bool(self.supplemental)
+
 
 class RequiredDimensionRecordsModel(DSGBaseModel):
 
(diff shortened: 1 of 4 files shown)


dsgrid/config/project_config.py
```diff
@@ -1318,6 +1332,36 @@ def supplemental_dimensions(self) -> dict:
         return self._supplemental_dimensions
 
 
+def _get_needed_base_dimensions(
```
daniel-thom, line 1335: I considered another solution to this problem. We could require that the user enter a consistent set of dimensions. 

For example, consider this [project config](https://github.com/dsgrid/dsgrid-project-DECARB/blob/ahcyip-patch-3/project/project.json5#L122). That config assumes that the first two multi-dim requirements will use all base dimension records. Rather than infer this fact with the code here, we could require a specification with a keyword like `__all__` as shown below:

```
subsector: {base: ['bev_compact', 'bev_midsize', …]},
metric: {base: ['electricity_ev_ldv_home_l1', 'electricity_ev_ldv_home_l2’, …],
geography: {base: [“__all__”]}
```

A solution like that would be more explicit and remove the need for code like this. This code does solve the example above in a straightforward way. However, there could be cases where several multi-dim requirements have different partial dimension overlaps and the result may not be intuitive.

elainethale: I do think it would be best for it to be clear what dimension_types the multi_dimensional requirements are defined over. I think your proposal for `geography: {base: ["__all__"]}` works. Would it also work to define, e.g., a `(subsector, metric, geography)` tuple somewhere and then assume that any that aren't listed are equivalent to `base: ["__all__"]`?

daniel-thom: As of now there isn’t a great place to put that tuple. I would prefer the `”__all__”` idea such that the rule is this:

All requirement objects in the multi-dim array with overlapping dimensions must be exactly overlapping. The code is very simple and the config file is not bad. The `__all__` tag isn’t great, but isn’t terrible either.

I suppose that I might prefer what I implemented today because it’s straightforward for the simple cases. However, like I said, I can imagine it getting confusing. Are you saying that you prefer the explicit approach with `__all__`? It would be simple to switch over.

elainethale: I think I prefer the explicit approach? My intuition isn't fully educated, as I couldn't really read through all your code today, but maybe that's a point in favor of just being explicit ...

daniel-thom: I implemented the `__all__` approach in the code and DECARB data.

elainethale (approved):

PR #292, opened 2024-04-08
Add support for base-to-supplemental dimension mappings with fractional values

daniel-thom:
Closes GitHub Issue #199


tests/test_queries.py
```diff
@@ -279,36 +280,62 @@ def test_query_cli_create_validate(tmp_path):
...
 
 
-def test_query_cli_run(tmp_path):
+def test_query_cli_run(tmp_path, cached_registry):
+    conn: DatabaseConnection = cached_registry
+    shutdown_project()
     output_dir = tmp_path / "queries"
-    project = get_project(
-        QueryTestElectricityValues.get_database_name(), QueryTestElectricityValues.get_project_id()
+    baseline_query = TEST_PROJECT_PATH / "test_efs" / "queries" / "baseline.json5"
```
daniel-thom, line 287: This needs a second test where the query uses pivoted format instead.

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -185,6 +185,7 @@ def make_project_dataframe_from_query(self, context, project_config):
     def _check_aggregations(self, context):
         if self._config.get_table_format_type() == TableFormatType.PIVOTED:
             pivoted_type = self._config.get_pivoted_dimension_type()
+            assert pivoted_type is not None
```
daniel-thom, line 188: These two changes fix complaints from pyright.

dsgrid/dataset/table_format_handler_base.py
```diff
@@ -109,6 +111,11 @@ def add_columns(
                 dataset_id=self.dataset_id,
             )
 
+        if "fraction" in df.columns:
```
elainethale, line 114: Why is it "fraction" here, whereas it was "from_fraction" above?

elainethale: Nevermind. I see now that `map_and_reduce_stacked_dimension` makes this change.

elainethale, line 114: I am surprised to see this at the same level as `for column in column_models:`. Am I reading this code (including the changes in query/models.py) correctly to say that multiple supplemental dimensions could be in the column_model list, e.g., for multiple dimension types? Or does one call to add_columns always process a single supplemental dimension, such that column_models is a list of supplemental dimension records?

dsgrid/dataset/pivoted_table.py
```diff
@@ -170,7 +170,7 @@ def process_stacked_aggregations(
             if not columns:
                 continue
 
-            df = self.add_columns(df, columns, context, True)
+            df = self.add_columns(df, columns, context, aggregation_columns=pivoted_columns)
```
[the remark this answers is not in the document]
daniel-thom: No. It is just a slight change in protocol. Before, you passed whether an aggregation was allowed. Now, if aggregation is allowed, you pass a list of value columns. The presence of value columns means that aggregation is allowed.

dsgrid/dataset/table_format_handler_base.py
```diff
@@ -65,22 +68,19 @@ def add_columns(
...
-                raise DSGInvalidParameter(
-                    f"Mapping dimension query name {query_name} produced from_fractions other than 1.0: {from_fractions}"
-                )
-            records = records.drop("from_fraction")
+                from_fractions = get_unique_values(records, "from_fraction")
+                if len(from_fractions) != 1 and float(next(iter(from_fractions))) != 1.0:
+                    raise DSGInvalidParameter(
+                        f"Mapping dimension query name {query_name} produced from_fractions other than 1.0: {from_fractions}"
+                    )
+
```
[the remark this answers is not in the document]
daniel-thom: Correct. However, I've been thinking about this case and have mostly concluded that it should be removed. We support the notion of `supplemental_columns` in the query result data model. It means that you can add a column, such as `state` to the table without having to fill out the aggregation section (which is tedious). That is duplicate to specifying `geography: ["county", "state"]` in the aggregation section.

1. Having two ways of doing the same thing is confusing to users.
2. The current implementation has a bug with ordering if you specify both an aggregation, such as county to state and a supplemental column, such as census_region. The best solution is to make the user specify all the columns they want in the aggregation section so that we can apply it in one groupBy expression.

Do you see any reason not to remove the notion of supplemental columns?

dsgrid/dataset/table_format_handler_base.py
daniel-thom: `column_models` is essentially a list of dimension query names in one aggregation. There can be multiple per dimension type (`county, state`) and multiple dimensions. There could also be multiple supplemental dimensions of the same type (`state, census_region`).

dsgrid/dataset/table_format_handler_base.py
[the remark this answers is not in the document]
elainethale: Yes--I think we wanted people to be able to see the county column with the state column also available, because counties are easier to understand if the state information is handy.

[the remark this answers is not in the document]
elainethale: I just realized that this feature could be really helpful for the crazy TEMPO subsectors--if we made a TEMPO-specific subsector class then we could have the mashed up ids AND each element broken out in a separate column.

dsgrid/dataset/table_format_handler_base.py
elainethale: Okay--so one aggregation means that the fractions can be continually collected: fraction = 1.0 * fraction_1 * fraction_2 * ... and then all applied at the end. That seems reasonable to me (although I can't fully prove it to myself without spending a bunch more time).

daniel-thom: Yes. This is the same methodology used for mapping dataset dimensions to the project.

dsgrid/dataset/table_format_handler_base.py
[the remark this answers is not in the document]
daniel-thom: Your reference to a new subsector class makes me think that we are missing each other. The `supplemental_columns` feature uses dimension mappings, not columns in the dimension record file.

Consider this example.
- User specifies that they want an aggregation of county to state and inserts `geography: ["state"]`.
- User wants the census region in the result table and so adds `supplemental_columns: ["census_region"]`

The supplemental columns have to be added before the base dimension is dropped. Then the code has to track these new columns during the aggregation process (there is currently a bug which can be fixed.)

My thought yesterday that this is all too complex for the users and the code and that we should give them one place to specify what columns they want in the table. The downside is that it is more complex to type in the JSON file. I think that's why we added the feature in the first place.

Let's go back to the case of just adding the state column (no aggregation).

Current recommendation:
```
supplemental_columns: ["state"]
```

Proposed:
```
aggregations: [
  {
    aggregation_function: "sum",
    dimensions: {
       geography: ["county", "state"]
       # every other dimension type + query name
    }
  }
]
```
There is a CLI command that auto-generates the aggregation section with base dimension query names, but it produces the more verbose form:
```
  "geography": [
    {
      "dimension_query_name": "county",
      "function": null,
      "alias": null
    }
  ],
```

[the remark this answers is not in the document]
daniel-thom: The `supplemental_columns` feature is removed.

PR #293, opened 2024-05-07
Optimize queries

daniel-thom:
1. Queries on unpivoted datasets were running unit conversion when it wasn't necessary, causing a slow-down.
2. Persist the query before evaluating the function remove_invalid_null_timestamps. This removes redundant processing and caused a 2x speed-up in the TEMPO dataset.
3. Improve metric dimension checks. We can validate that an aggregation query includes the metric dimension up front.
4. Fix aggregation query when the target dataset is pivoted and the source has expected-missing data.


elainethale (approved):
LGTM

PR #295, opened 2024-05-17
Disallow aggregations of pivoted tables

daniel-thom:
This commit removes all use of pivoted tables during the aggregation phase of dsgrid queries. Users can still use a pivoted format for their datasets and for the final result of queries. However, dsgrid will store all dataset in unpivoted format.


dsgrid/config/dataset_config.py
```diff
@@ -677,7 +678,9 @@ def get_value_columns(self) -> list[str]:
             case TableFormatType.PIVOTED:
                 return self.get_pivoted_dimension_columns()
             case TableFormatType.UNPIVOTED:
-                return [self.model.data_schema.table_format.value_column]
+                # TODO DT: is this acceptable
+                return [VALUE_COLUMN]
+                # return [self.model.data_schema.table_format.value_column]
```
daniel-thom, line 683: We were rarely using the user customization. I'm suggesting here that we should only ever use `"value"`. It will be easier to understand and code against. Users are already forced to encode column names with dsgrid dimension types, and so this doesn't seem like an inconvenience to me.

tests/test_project.py
```diff
@@ -135,51 +126,51 @@ def test_dimension_map_and_reduce_in_dataset(cached_registry):
     assert len(fraction) == 1
     assert fraction[0] == (0.9 * 1.3)
 
+    # TODO DT: Do we need this?
```
daniel-thom, line 129: @lixiliu Can I delete this code block? I think it was testing pivoted column "map and reduce" which we no longer perform.

dsgrid/config/dataset_config.py
lixiliu: I could see users wanting to name the column in whatever unit the value col represents, esp if a table is only going to have one type of value and one unit, but if we only want to support "value", it would be fine with me.

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -432,6 +308,19 @@ def _iter_dataset_record_ids(self, context: QueryContext):
...
+    def _list_dimension_columns(df: DataFrame) -> list[str]:
+        columns = DimensionType.get_allowed_dimension_column_names()
+        return [x for x in df.columns if x in columns]
+
+    def _list_dimensions(self, df: DataFrame) -> list[DimensionType]:
+        dims = [DimensionType(x) for x in DatasetSchemaHandlerBase._list_dimension_columns(df)]
+        if self._config.get_table_format_type() == TableFormatType.PIVOTED:
+            pivoted_type = self._config.get_pivoted_dimension_type()
+            assert pivoted_type is not None
+            dims.append(pivoted_type)
```
lixiliu, line 321: Is this right? _list_dimensions() is used to retain relevant columns and make association table. If pivoted on metric, metric would not be a column name right?

lixiliu: Do we have a test for this section?

lixiliu (commented):
It's a lot of code changes, I'd appreciate a walk through to make sure I haven't glossed over anything that I shouldn't.

dsgrid/query/query_submitter.py
```diff
@@ -406,3 +394,9 @@ def _save_composite_dataset(self, context: QueryContext, df, repartition):
         filename = output_dir / "table.parquet"
         self._save_result(context, df, filename, repartition)
         self.metadata_filename(output_dir).write_text(context.metadata.model_dump_json(indent=2))
+
+
+def _pivot_table(df: DataFrame, context: QueryContext):
+    pivoted_column = context.convert_to_pivoted()
+    ids = [x for x in df.columns if x not in {pivoted_column, VALUE_COLUMN}]
+    return df.groupBy(*ids).pivot(pivoted_column).sum(VALUE_COLUMN)
```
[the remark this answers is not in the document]
daniel-thom: Discussed and agreed to leave it the way it is because we aren't tracking the values across filters.

lixiliu (approved):

PR #298, opened 2024-06-10
Simplify time dimension interfaces

daniel-thom:
The time conversion code was incorrectly using the model year dimension. Removed model year from the `TimeDimensionBaseConfig` interfaces. Mapping the time dimension now requires case-by-case conditionals to account for requirements that span dimension types.

Added persistence of the intermediate query before making time adjustments in order to avoid repeated evaluations of the query.

### DIFF
--- a/dsgrid/config/annual_time_dimension_config.py
+++ b/dsgrid/config/annual_time_dimension_config.py
@@ -4,17 +4,15 @@ from datetime import timedelta
 import pandas as pd
 import pyspark.sql.functions as F
 from pyspark.sql import DataFrame
-from pyspark.sql.types import StructType, StructField, IntegerType
+from pyspark.sql.types import StructType, StructField, IntegerType, StringType
 
 from dsgrid.config.date_time_dimension_config import DateTimeDimensionConfig
-from dsgrid.dimension.time import (
-    MeasurementType,
-    AnnualTimeRange,
-    TimeDimensionType,
-)
+from dsgrid.dimension.base_models import DimensionType
+from dsgrid.dimension.time import AnnualTimeRange
 from dsgrid.exceptions import DSGInvalidDataset
 from dsgrid.time.types import AnnualTimestampType
 from dsgrid.utils.timing import timer_stats_collector, track_timing
+from dsgrid.utils.scratch_dir_context import ScratchDirContext
 from dsgrid.utils.spark import get_spark_session, custom_spark_conf
 from .dimensions import AnnualTimeDimensionModel
 from .time_dimension_base_config import TimeDimensionBaseConfig
@@ -75,61 +73,12 @@ class AnnualTimeDimensionConfig(TimeDimensionBaseConfig):
         df,
         project_time_dim,
         value_columns: set[str],
+        scratch_dir_context: ScratchDirContext,
         wrap_time_allowed=False,
         time_based_data_adjustment=None,
     ):
-        assert value_columns is not None
-        match self.model.measurement_type:
-            case MeasurementType.MEASURED:
-                df = self.map_annual_time_measured_to_datetime(df, project_time_dim)
-            case MeasurementType.TOTAL:
-                df = self.map_annual_total_to_datetime(df, project_time_dim, value_columns)
-            case _:
-                raise NotImplementedError(f"Unhandled: {self.model.measurement_type}")
-
-        return df
-
-    def map_annual_time_measured_to_datetime(
-        self, annual_df: DataFrame, dt_dim: DateTimeDimensionConfig
-    ):
-        """Map a dataframe with MeasuredType.MEASURED to DateTime."""
-        # Does this category even make sense? Why would someone report
-        # something other than total for annual?
-        dt_time_col = dt_dim.get_load_data_time_columns()[0]
-        df = dt_dim.build_time_dataframe()
-        with custom_spark_conf(
-            {"spark.sql.session.timeZone": dt_dim.model.datetime_format.timezone.tz_name}
-        ):
-            years = df.withColumn("year", F.year(dt_time_col)).select("year").distinct().collect()
-            if len(years) != 1:
-                raise Exception("Bug: project time has more than one year: {years=}")
-        return df.crossJoin(annual_df)
-
-    def map_annual_total_to_datetime(
-        self, annual_df: DataFrame, dt_dim: DateTimeDimensionConfig, value_columns: set[str]
-    ):
-        """Map a dataframe with MeasuredType.TOTAL to DateTime."""
-        assert (
-            dt_dim.model.time_type == TimeDimensionType.DATETIME
-        ), "dt_dim must be datetime type."
-        frequency = dt_dim.model.frequency
-        df = dt_dim.build_time_dataframe()
-        dt_time_col = dt_dim.get_load_data_time_columns()[0]
-        with custom_spark_conf(
-            {"spark.sql.session.timeZone": dt_dim.model.datetime_format.timezone.tz_name}
-        ):
-            years = df.withColumn("year", F.year(dt_time_col)).select("year").distinct().collect()
-            if len(years) != 1:
-                raise Exception("Bug: project time has more than one year: {years=}")
-            year = years[0].year
-            if self.model.include_leap_day and year % 4 == 0:
-                measured_duration = timedelta(days=366)
-            else:
-                measured_duration = timedelta(days=365)
-        df2 = df.crossJoin(annual_df)
-        for column in value_columns:
-            df2 = df2.withColumn(column, F.col(column) / (measured_duration / frequency))
-        return df2
+        # Currently, only map_annual_historical_time_to_date_time is supported.
+        raise NotImplementedError("AnnualTimeDimensionConfig.convert_dataframe")
 
     def get_frequency(self):
         return timedelta(days=365)
@@ -167,3 +116,39 @@ class AnnualTimeDimensionConfig(TimeDimensionBaseConfig):
             start, end = (int(time_range.start), int(time_range.end))
             timestamps += [AnnualTimestampType(x) for x in range(start, end + 1)]
         return timestamps
+
+
+def map_annual_historical_time_to_date_time(
+    df: DataFrame,
+    annual_dim: AnnualTimeDimensionConfig,
+    dt_dim: DateTimeDimensionConfig,
+    value_columns: set[str],
+) -> DataFrame:
+    """Map a DataFrame with an annual/historical time dimension to a DateTime time dimension."""
+    annual_col = annual_dim.get_load_data_time_columns()[0]
+    dt_df = dt_dim.build_time_dataframe()
+    myear_column = DimensionType.MODEL_YEAR.value
+    dt_col = dt_dim.get_load_data_time_columns()[0]
+
+    # Note that MeasurementType.TOTAL has already been verified.
+    with custom_spark_conf(
+        {"spark.sql.session.timeZone": dt_dim.model.datetime_format.timezone.tz_name}
+    ):
+        years = dt_df.withColumn("year", F.year(dt_col)).select("year").distinct().collect()
+        if len(years) != 1:
+            msg = "DateTime dimension has more than one year: {years=}"
+            raise NotImplementedError(msg)
+        if annual_dim.model.include_leap_day and years[0].year % 4 == 0:
+            measured_duration = timedelta(days=366)
+        else:
+            measured_duration = timedelta(days=365)
+
+    df2 = (
+        df.crossJoin(dt_df)
+        .withColumn(myear_column, F.col(annual_col).cast(StringType()))
+        .drop(annual_col)
+    )
+    frequency = dt_dim.model.frequency
+    for column in value_columns:
+        df2 = df2.withColumn(column, F.col(column) / (measured_duration / frequency))
+    return df2
--- a/dsgrid/config/annual_time_dimension_config.py
+++ b/dsgrid/config/annual_time_dimension_config.py
@@ -3,8 +3,10 @@ from datetime import timedelta
 
 import pandas as pd
 import pyspark.sql.functions as F
+from pyspark.sql import DataFrame
 from pyspark.sql.types import StructType, StructField, IntegerType
 
+from dsgrid.config.date_time_dimension_config import DateTimeDimensionConfig
 from dsgrid.dimension.time import (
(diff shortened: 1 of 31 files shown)


dsgrid/config/annual_time_dimension_config.py
```diff
@@ -70,73 +71,70 @@ def convert_dataframe(
...
+        scratch_dir_context: ScratchDirContext,
         wrap_time_allowed=False,
         data_adjustment=None,
     ):
-        assert model_years is not None
         assert value_columns is not None
         match self.model.measurement_type:
             case MeasurementType.MEASURED:
-                df = self.map_annual_time_measured_to_datetime(df, project_time_dim, model_years)
+                df = self.map_annual_time_measured_to_datetime(df, project_time_dim)
```
daniel-thom, line 82: I'm not sure that it makes sense for annual time to have a measurement type other than total. To have some other type, wouldn't we need to have frequency in the config?

dsgrid/config/annual_time_dimension_config.py
```diff
@@ -70,73 +71,70 @@ def convert_dataframe(
...
             dt_dim.model.time_type == TimeDimensionType.DATETIME
         ), "dt_dim must be datetime type."
         frequency = dt_dim.model.frequency
-        time_col = self.get_load_data_time_columns()[0]
-        dfs = []
-        for model_year in model_years:
-            if self.model.include_leap_day and model_year % 4 == 0:
+        df = dt_dim.build_time_dataframe()
+        dt_time_col = dt_dim.get_load_data_time_columns()[0]
+        with custom_spark_conf(
```
daniel-thom, line 114: I'm not sure that this logic is correct. Do we need to handle the case of a project time range that spans one year?

dsgrid/config/time_dimension_base_config.py
```diff
@@ -229,19 +211,16 @@ def _convert_time_to_project_time_interval(
         has a different time zone than project. wrap_time_allowed is specified by
         the project's InputDatasetModel.
         """
-        if project_time_dim is None:
-            return df
-
-        df = self._align_time_interval_type(df, project_time_dim)
+        # Persist the table because the code below will need to evaluate the query multiple times.
+        df = persist_intermediate_table(df, context)
+        df, handled_time_diff = self._align_time_interval_type(df, project_time_dim)
```
daniel-thom, line 216: @lixiliu Note this change. You may have been making it in your branch. I need it working here.

dsgrid/config/annual_time_dimension_config.py
```diff
@@ -70,73 +72,22 @@ def convert_dataframe(
...
-                df = df.join(annual_df, on=F.year(df[tmp_col]) == annual_df[time_col]).drop(
-                    time_col, tmp_col
-                )
-            for column in value_columns:
-                df = df.withColumn(column, F.col(column) / (measured_duration / frequency))
-            if not df.rdd.isEmpty():
-                dfs.append(df)
-
-        return union(dfs)
+        # Currently, only map_annual_historical_time_to_date_time is supported.
```
daniel-thom, line 80: This change removes polymorphism for this `convert_dataframe` interface. It is difficult, if not impossible, to make one consistent interface because mapping time dimensions from dataset to project may require other dimensions.

dsgrid/config/annual_time_dimension_config.py
```diff
@@ -70,73 +72,22 @@ def convert_dataframe(
...
-        value_columns=None,
+        value_columns: set[str],
+        scratch_dir_context: ScratchDirContext,
         wrap_time_allowed=False,
         data_adjustment=None,
     ):
-        assert model_years is not None
-        assert value_columns is not None
-        match self.model.measurement_type:
-            case MeasurementType.MEASURED:
```
daniel-thom, line 81: I am removing support for measured historical time. What we were doing is not correct because there is no definition of the interval being measured. The measured/min/max/mean types need to specify the interval for the measurement.

dsgrid/config/annual_time_dimension_config.py
```diff
@@ -161,15 +112,45 @@ def get_tzinfo(self):
...
+    dt_df = dt_dim.build_time_dataframe()
+    myear_column = DimensionType.MODEL_YEAR.value
+    dt_col = dt_dim.get_load_data_time_columns()[0]
+
+    # Note that MeasurementType.TOTAL has already been verified.
+    with custom_spark_conf(
+        {"spark.sql.session.timeZone": dt_dim.model.datetime_format.timezone.tz_name}
+    ):
+        years = dt_df.withColumn("year", F.year(dt_col)).select("year").distinct().collect()
+        if len(years) != 1:
```
daniel-thom, line 140: Is this acceptable? I'm not sure what to do if the project time spans multiple years.

dsgrid/config/dimensions.py
```diff
@@ -696,6 +696,16 @@ def check_times(
             return ranges
         return _check_time_ranges(ranges, info.data["str_format"], timedelta(days=365))
 
+    @field_validator("measurement_type")
+    @classmethod
+    def check_measurement_type(cls, measurement_type: MeasurementType) -> MeasurementType:
+        # This restriction exists because any other measurement type would require a frequency,
+        # and that isn't part of the model definition.
+        if measurement_type != MeasurementType.TOTAL:
```
daniel-thom, line 704: Isn't this a problem for all time dimensions?

dsgrid/config/time_dimension_base_config.py
```diff
@@ -229,19 +219,18 @@ def _convert_time_to_project_time_interval(
...
+        df = persist_intermediate_table(df, context)
+        df, handled_time_diff = self._align_time_interval_type(df, project_time_dim, wrap_time)
 
-        df = self._align_time_interval_type(df, project_time_dim)
-
-        if wrap_time:
-            diff = self._time_difference(df, project_time_dim, difference="left")
-            if diff:
-                df = self._apply_time_wrap(df, project_time_dim, diff)
+        if wrap_time and not handled_time_diff:
```
daniel-thom, line 226: @lixiliu Let me know if I should remove these changes in favor of yours.

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -562,13 +563,26 @@ def _convert_time_dimension(
                 geography_dim = self._config.get_dimension(DimensionType.GEOGRAPHY)
             load_data_df = add_time_zone(load_data_df, geography_dim)
 
-        if model_years is not None:
-            model_years = sorted(int(x) for x in model_years)
+        if isinstance(time_dim, AnnualTimeDimensionConfig):
+            project_time_dim = project_config.get_base_dimension(DimensionType.TIME)
+            if not isinstance(project_time_dim, DateTimeDimensionConfig):
+                msg = f"annual time can only be mapped to DateTime: {project_time_dim.model.time_type}"
```
elainethale, line 569:
```suggestion
                msg = f"Annual time can only be mapped to DateTime, but project time is of type: {project_time_dim.model.time_type}"
```

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -562,13 +563,26 @@ def _convert_time_dimension(
...
+                raise NotImplementedError(msg)
+            if self._config.model.dataset_type == InputDatasetType.HISTORICAL:
+                return map_annual_historical_time_to_date_time(
+                    load_data_df,
+                    time_dim,
+                    project_time_dim,
+                    value_columns,
+                )
+            msg = f"Cannot map AnnualTime / {self._config.model.dataset_type}"
```
elainethale, line 578:
```suggestion
            msg = f"Cannot map AnnualTime for a dataset of type {self._config.model.dataset_type}"
```

dsgrid/config/dimensions.py
elainethale: Maybe right now, when we only really support energy units, but we need to expand to also support other units, e.g., power, for which other measurement types (instantaneous, mean) would make sense.

dsgrid/config/time_dimension_base_config.py
```diff
@@ -254,26 +243,34 @@ def _align_time_interval_type(self, df, project_time_dim):
         time_col = time_col[0]
 
         if dtime_interval == ptime_interval:
-            return df
+            return df, False
```
elainethale, line 246: Since above you just call `self._handle_time_difference(df, project_time_dim, wrap_time)` in this case, could you do that here instead and avoid having to pass a tuple of (dataframe, bool)?

dsgrid/config/time_dimension_base_config.py
```diff
@@ -254,26 +243,34 @@ def _align_time_interval_type(self, df, project_time_dim):
...
         )
+        return self._handle_time_difference(df, project_time_dim, wrap_time), True
+
+    def _handle_time_difference(
+        self, df: DataFrame, project_time_dim: "TimeDimensionBaseConfig", wrap_time: bool
+    ) -> DataFrame:
         diff = self._time_difference(df, project_time_dim, difference="left")
         if diff:
+            if not wrap_time:
+                logger.warning("Wrapping time even though the config does not permit it.")
```
elainethale, line 259: Can we provide any more information in this warning message so the user can get a sense for what's happening?

dsgrid/config/annual_time_dimension_config.py
elainethale: Annual time tells you the interval--one year. But looking at the options, I see that "measured" is essentially synonymous with "instantaneous", i.e., more like power than energy units, so I'm fine with removing support for now.

dsgrid/config/annual_time_dimension_config.py
elainethale: Yes, acceptable.

dsgrid/config/annual_time_dimension_config.py
```diff
@@ -161,15 +112,45 @@ def get_tzinfo(self):
...
+            msg = "DateTime dimension has more than one year: {years=}"
+            raise NotImplementedError(msg)
+        if annual_dim.model.include_leap_day and years[0].year % 4 == 0:
```
elainethale, line 143: nit: no leap day in 2100. 

```
(years[0].year % 4 == 0) & ((years[0].year % 100 != 0) or (years[0].year % 400 == 0))
```

elainethale (commented):
I like the interface clean-up!

dsgrid/project.py
```diff
@@ -155,10 +159,29 @@ def unload_dataset(self, dataset_id):
...
+        self,
+        dataset_id: str,
+        output_dir: Path,
+        overwrite=False,
+        scratch_dir: Optional[Path] = None,
+    ) -> DataFrame:
+        """Transform a dataset by mapping its dimensions to the project's dimensions and write it
+        to the filesystem.
+        """
+        output_file = output_dir / "table.parquet"
```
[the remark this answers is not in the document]
daniel-thom: This is an unused function. It provides the ability to map a dataset to project dimensions without a query. It could be supplanted with an auto-generated query. Anyway, in order to work, we have to persist the dataframe to files in a permanent location because we may be relying on temporary files (the scratch directory gets deleted when the function exits.

I made a mistake in the return. We should either return the path alone or the dataframe and the path.

dsgrid/utils/spark.py
```diff
@@ -547,6 +561,17 @@ def write_dataframe(df: DataFrame, filename: str | Path) -> None:
...
 
+def persist_intermediate_table(df: DataFrame, context: ScratchDirContext) -> DataFrame:
+    """Persist a table to the scratch directory. This can be helpful to avoid multiple
+    evaluations of the same query.
+    """
+    # Note: This does not use the Spark warehouse because we are not properly configuring or
+    # managing it across sessions. And, we are already using the scratch dir for our own files.
+    path = context.get_temp_filename(suffix=".parquet")
+    write_dataframe(df, path)
+    return read_dataframe(path)
```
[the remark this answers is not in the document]
elainethale: On the first point, don't you have to be able to access the file path where you persisted the intermediate result in order to make progress later?

dsgrid/config/time_dimension_base_config.py
```diff
@@ -210,19 +211,20 @@ def _convert_time_to_project_time(
...
+        # df = persist_intermediate_table(df, context)
+
         time_col = project_time_dim.get_load_data_time_columns()
         assert len(time_col) == 1, time_col
         time_col = time_col[0]
         time_map = (
             df.select(time_col).distinct().select(time_col, F.col(time_col).alias("orig_ts"))
-        )
+        ).cache()
+        time_map.count()
```
lixiliu, line 227: What does this do?

daniel-thom: Calling df.cache() by itself does nothing. If you call df.count(), it will materialize the current table.

lixiliu: Maybe we materialize just before time_map_diff?

daniel-thom: `df` is a very large table. The query represented by `time_map` reduces that table to 8760/8784 rows with one column. We should materialize it as soon as possible. Otherwise, that query will be evaluated in each of `_align_time_interval_type`, `filter_to_project_timestamps`, `time_difference`, and `apply_time_wrap`. Or, am I missing your point?

tests/test_annual_time.py
```diff
@@ -91,50 +89,41 @@ def date_time_dimension():
...
-        df.withColumn("year", F.year(df.timestamp)).select("year").distinct().collect()
+    expected_by_year = {
+        x.time_year: x.electricity_sales / (366 * 24) for x in annual_dataframe.collect()
+    }
+    _check_values(annual_dataframe, date_time_dimension, df, expected_by_year)
+
+
+def _check_values(annual_dataframe, date_time_dimension, df, expected_by_year):
+    num_rows = annual_dataframe.count()
+    num_timestamps = 24 * 7
```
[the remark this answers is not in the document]
daniel-thom: It is the expectation of this test that the code will produce this number of timestamps.

dsgrid/utils/spark.py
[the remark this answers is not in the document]
daniel-thom: I retract that point. We can never use this feature to resume.

lixiliu (approved):
We should add GitHub issues on things we are punting, like cleaning up time_utils or the "max", "avg",... options for metric quantity

Issue #304, opened 2024-07-02
Add CLI command to add new supplemental dimensions to a project


daniel-thom:
Pre-conditions:
- A project is registered and has datasets submitted to it.
- User has one or more new supplemental dimensions defined in a JSON5 file + CSV files where the format is equal to the unregistered dimensions and supplemental dimensions in project config files.

Post-conditions:
- The supplemental dimensions and mappings are registered and added to the project config.
- The project's version has its patch number increased (1.0.0 -> 1.0.1).
- All datasets that were successfully submitted to the project are unaffected. They do not need to be resubmitted.

PR #308, opened 2024-07-03
Add CLI command to add supplemental dimensions

daniel-thom:
Closes GitHub Issue #304

### DIFF
--- a/dsgrid/cli/registry.py
+++ b/dsgrid/cli/registry.py
@@ -696,12 +696,12 @@ def update_project(
     "--log-message",
     required=True,
     type=str,
-    help="Reason for addition",
+    help="Please specify the reason for this addition.",
 )
 def register_supplemental_dimensions(
     ctx, registry_manager, project_id, filename: Path, log_message
 ):
-    """Register new supplemental dimensions to a project."""
+    """Register new supplemental dimensions with a project."""
     submitter = getpass.getuser()
     project_mgr = registry_manager.project_manager
     res = handle_dsgrid_exception(
--- a/dsgrid/cli/registry.py
+++ b/dsgrid/cli/registry.py
@@ -686,6 +686,36 @@ def update_project(
         return 1
 
 
+@click.command()
+@click.pass_obj
+@click.pass_context
+@click.argument("project_id")
+@click.argument("filename", callback=_path_callback)
+@click.option(
+    "-l",
+    "--log-message",
+    required=True,
+    type=str,
+    help="Reason for addition",
+)
+def register_supplemental_dimensions(
+    ctx, registry_manager, project_id, filename: Path, log_message
+):
+    """Register new supplemental dimensions to a project."""
+    submitter = getpass.getuser()
+    project_mgr = registry_manager.project_manager
+    res = handle_dsgrid_exception(
+        ctx,
+        project_mgr.register_supplemental_dimensions,
+        project_id,
+        filename,
+        submitter,
+        log_message,
+    )
+    if res[1] != 0:
+        return 1
+
+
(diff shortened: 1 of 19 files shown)


dsgrid/cli/registry.py
```diff
@@ -686,6 +686,36 @@ def update_project(
...
+@click.pass_obj
+@click.pass_context
+@click.argument("project_id")
+@click.argument("filename", callback=_path_callback)
+@click.option(
+    "-l",
+    "--log-message",
+    required=True,
+    type=str,
+    help="Reason for addition",
```
elainethale, line 699:
```suggestion
    help="Please specify the reason for this addition in your log message",
```

dsgrid/cli/registry.py
```diff
@@ -686,6 +686,36 @@ def update_project(
...
+)
+def register_supplemental_dimensions(
+    ctx, registry_manager, project_id, filename: Path, log_message
+):
+    """Register new supplemental dimensions to a project."""
```
elainethale, line 704:
```suggestion
    """Register new supplemental dimensions with a project."""
```

dsgrid/config/supplemental_dimension.py
```diff
@@ -19,3 +20,23 @@ class SupplementalDimensionModel(DimensionModel):
...
+
+class SupplementalDimensionsListModel(DSGBaseModel):
+    """Defines a list of supplemental dimensions."""
+
+    supplemental_dimensions: Annotated[
+        list[SupplementalDimensionModel],
+        Field(
+            title="supplemental_dimensions",
+            description="List of supplemental dimensions. They will be automatically registered "
+            "during project registration and then converted to supplemental_dimension_references.",
```
elainethale, line 33: Is this model only used for the new CLI functionality to add supplemental dimensions to an already-registered project? If so, should this description say something different than "They will be automatically registered during project registration"?

dsgrid/cli/registry.py
elainethale: Probably my suggested language doesn't fit with the rest of the CLI. How about, "Reason for adding these supplemental dimensions to the project"?

dsgrid/config/supplemental_dimension.py
```diff
@@ -19,3 +20,12 @@ class SupplementalDimensionModel(DimensionModel):
             title="mapping",
         ),
     ]
+
+
+class SupplementalDimensionsListModel(DSGBaseModel):
+    """Defines a list of supplemental dimensions."""
+
+    supplemental_dimensions: Annotated[
```
lixiliu, line 28: Is this true, a container for both dimensions and mappings? If so, can we name it as such?

daniel-thom: This is directly analogous to this field: https://github.com/dsgrid/dsgrid/blob/main/dsgrid/config/project_config.py#L262. Would you prefer to call this `supplemental_dimensions_and_mappings` and lose the symmetry? Also, I could make an argument that a mapping is an attribute of a supplemental dimension and is therefore covered under this simpler name.

lixiliu: If this is already in place in other config types, I am fine with keeping it as is. At one point, I think we have dimensions and mappings separate? So in my mental model, they are two distinct entities. Maybe it's time for me to update that.

daniel-thom: We store them separately in the registry. However, at some point we changed the project config definition to allow them to be specified in one object, as in this example: https://github.com/dsgrid/dsgrid-project-DECARB/blob/main/project/project.json5#L320.

lixiliu (approved):

Issue #307, opened 2024-07-02
 Add CLI command to replace dataset requirements in a project config


daniel-thom:
Pre-conditions:

- A project is registered and has datasets submitted to it.
- User has new requirements for one or more datasets in a JSON5 file.

Post-conditions:

- The `dataset_requirements` section for each dataset in the project config is updated with the new requirements.
- The project's version has its patch number increased (1.0.0 -> 1.0.1).
- All affected datasets have that project-level status set to unregistered.
- All unaffected datasets are unchanged.

daniel-thom:
This needs to support adding a new dataset.

PR #310, opened 2024-07-03
Add CLI command to replace dataset dimension requirements

daniel-thom:
Closes GitHub Issue #307


elainethale (commented):
Should this PR also include a documentation change/addition?

daniel-thom:
I'm still thinking about how to refer to the required data models in the CLI help.

elainethale:
> I'm still thinking about how to refer to the required data models in the CLI help.

This is exactly my concern--how do people know the correct format for the files they are submitting?

daniel-thom:
> > I'm still thinking about how to refer to the required data models in the CLI help.
> 
> This is exactly my concern--how do people know the correct format for the files they are submitting?

Note that we don't give any links for any other command. We expect readers to use the documentation where we discuss it. These are a bit more obscure, so I added some hyperlinks. It doesn't look very good in the terminal, though. I added printing with rich + rich_click, but I gave up on figuring out how to display the link name instead of the full URL. The printing with Rich is significantly better.

dsgrid/config/input_dataset_requirements.py
```diff
@@ -0,0 +1,39 @@
...
+            title="required_dimensions",
+            description="Defines required record IDs that must exist for each dimension.",
+        ),
+    ]
+
+
+class InputDatasetDimensionRequirementsListModel(DSGBaseModel):
+    """Defines a list of dataset dimension requirements."""
+
+    datasets: Annotated[
```
lixiliu, line 26: Is this name correct given description? It's list of dataset dimension requirements

dsgrid/registry/project_registry_manager.py
```diff
@@ -789,6 +793,61 @@ def register_supplemental_dimensions(
...
+        filename: Path,
+        submitter: str,
+        log_message: str,
+    ):
+        """Replace dataset requirements in a project."""
+        config = self.get_by_id(project_id)
+        model = InputDatasetDimensionRequirementsListModel.from_file(filename)
+        for dataset in model.datasets:
+            found = False
+            for i in range(len(config.model.datasets)):
```
lixiliu, line 830: `For I, dataset in enumerate(config.model.datasets):`

dsgrid/registry/project_registry_manager.py
```diff
@@ -789,6 +793,61 @@ def register_supplemental_dimensions(
...
+                if config.model.datasets[i].dataset_id == dataset.dataset_id:
+                    config.model.datasets[i].required_dimensions = dataset.required_dimensions
+                    if config.model.datasets[i].status == DatasetRegistryStatus.REGISTERED:
```
lixiliu, line 833: Why are we unregistering here?

dsgrid/registry/project_registry_manager.py
daniel-thom: I used an index to make it more obvious that we are changing the config as well as to avoid an odd variable name (`dataset` is already used in this scope).

dsgrid/registry/project_registry_manager.py
daniel-thom: The project has new requirements for the dataset, and so it needs to be re-registered.

dsgrid/config/input_dataset_requirements.py
daniel-thom: If this is confusing, I can change it. Each entry in this list is a set of requirements for a dataset. `datasets` is shorter than `dataset_dimension_requirements`. It's also like an overlay to this class: https://github.com/dsgrid/dsgrid/blob/main/dsgrid/config/project_config.py#L695, and I like the symmetry.

lixiliu: I think the symmetry is good, but if the label is the same as the InputDatasetListModel, I would expect it to take those model/input rather than the RequirementsListModel here. So personally, I would prefer a more nuanced label.

lixiliu (approved):
Leaving the option up to you to change the config parameter name, otherwise things look good, thanks for addressing my comments.

dsgrid/config/input_dataset_requirements.py
daniel-thom: This parameter is user facing and so consensus would be good. @elainethale and @ashreeta can also weigh in.

I'll elaborate on the symmetry argument. In this scenario the user will most likely copy the existing dataset section of the project config into a new JSON file, remove unrelated fields, edit the requirements, and then upload it through this new CLI command. While calling the field `dataset_dimension_requirements` is more specific, I think that `datasets` is simpler, provides sufficient explanation, and is the same name that is used in the project config (allowing copy/paste) to work. The new blob would look something like this:
```
datasets: [
  dataset_id: 'decarb_2023_dgen',
  required_dimensions: {
    single_dimensional: {
      sector: {base: ['res', 'com']},
      metric: {base: ['electricity_dpv']}
    }
  }
]
```

elainethale: I also prefer the more accurate name,`dataset_dimension_requirements`. The user already has to modify what they've copy-pasted ...

Issue #320, opened 2024-08-08
dsgrid-admin datasets remove


ashreeta:
dsgrid-admin datasets remove does not remove registered datasets from projects (it should).

Issue #309, opened 2024-07-03
Need to track source datasets in a derived dataset


daniel-thom:
When we create a derived dataset, we are not linking that dataset with its source datasets in the registry graph. This is challenging because there is nothing in the dataset config that indicates that it is a derived dataset. We likely need to add a field to that dataset config that specifies the source datasets and versions. Then, at registration time, we can link them in the registry graph. If something is later updated in the project or a source dataset causing that dataset to be unregistered, we can also unregister the derived dataset.

PR #323, opened 2024-08-09
Unregister datasets in projects on dataset removal

daniel-thom:
Closes GitHub Issue #320

### DIFF
--- a/dsgrid/cli/dsgrid_admin.py
+++ b/dsgrid/cli/dsgrid_admin.py
@@ -1,5 +1,6 @@
 """Main CLI command for dsgrid."""
 
+import getpass
 import logging
 import shutil
 import sys
@@ -12,6 +13,7 @@ from dsgrid.common import LOCAL_REGISTRY, REMOTE_REGISTRY
 from dsgrid.config.simple_models import RegistrySimpleModel
 from dsgrid.dsgrid_rc import DsgridRuntimeConfig
 from dsgrid.loggers import setup_logging, check_log_file_size
+from dsgrid.registry.common import DatasetRegistryStatus, VersionUpdateType
 from dsgrid.registry.registry_database import DatabaseConnection
 from dsgrid.registry.registry_manager import RegistryManager
 from dsgrid.registry.filter_registry_manager import FilterRegistryManager
@@ -114,25 +116,25 @@ def registry(ctx, remote_path):
 
 @click.group()
 @click.pass_obj
-def dimensions(registry_manager):
+def dimensions(registry_manager: RegistryManager):
     """Dimension subcommands"""
 
 
 @click.group()
 @click.pass_obj
-def dimension_mappings(registry_manager):
+def dimension_mappings(registry_manager: RegistryManager):
     """Dimension mapping subcommands"""
 
 
 @click.group()
 @click.pass_obj
-def projects(registry_manager):
+def projects(registry_manager: RegistryManager):
     """Project subcommands"""
 
 
 @click.group()
 @click.pass_obj
-def datasets(registry_manager):
+def datasets(registry_manager: RegistryManager):
     """Dataset subcommands"""
 
 
@@ -182,7 +184,7 @@ Dimension Commands
 @click.command(name="remove")
 @click.argument("dimension-id")
 @click.pass_obj
-def remove_dimension(registry_manager, dimension_id):
+def remove_dimension(registry_manager: RegistryManager, dimension_id: str):
     """Remove a dimension from the dsgrid repository."""
     registry_manager.dimension_manager.remove(dimension_id)
 
@@ -195,7 +197,7 @@ Dimension Mapping Commands
 @click.command(name="remove")
 @click.argument("dimension-mapping-id")
 @click.pass_obj
-def remove_dimension_mapping(registry_manager, dimension_mapping_id):
+def remove_dimension_mapping(registry_manager: RegistryManager, dimension_mapping_id: str):
     """Remove a dimension mapping from the dsgrid repository."""
     registry_manager.dimension_mapping_manager.remove(dimension_mapping_id)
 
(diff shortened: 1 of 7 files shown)


dsgrid/registry/dataset_registry_manager.py
```diff
@@ -361,8 +361,8 @@ def _update(
 
         return model
 
-    def remove(self, dataset_id: str):
-        config = self.get_by_id(dataset_id)
+    def remove(self, dataset_id: str, version: Optional[str] = None):
+        config = self.get_by_id(dataset_id, version=version)
```
elainethale, line 365: So `version is None` returns just the latest dataset with that id?

dsgrid/registry/dimension_mapping_registry_manager.py
```diff
@@ -411,7 +412,7 @@ def _update(self, config, submitter, update_type, log_message):
 
         return model
 
-    def remove(self, mapping_id):
+    def remove(self, mapping_id: str, version: Optional[str] = None):
```
elainethale, line 415: version isn't used below. Should it be?

dsgrid/registry/dimension_registry_manager.py
```diff
@@ -384,7 +384,7 @@ def _update(self, config, submitter, update_type, log_message):
         self._dimensions[new_key] = get_dimension_config(model)
         return model
 
-    def remove(self, dimension_id):
+    def remove(self, dimension_id, version: Optional[str] = None):
```
elainethale, line 387: Ditto here

dsgrid/registry/project_registry_manager.py
```diff
@@ -1259,7 +1259,7 @@ def _update_config(self, config, submitter, update_type, log_message):
         self._projects[new_key] = new_config
         return new_config
 
-    def remove(self, project_id: str):
+    def remove(self, project_id: str, version: Optional[str] = None):
```
elainethale, line 1262: Ditto

tests/test_registry_management.py
```diff
@@ -330,6 +331,40 @@ def test_add_supplemental_dimension(tmp_registry_db):
...
+            "registry",
+            "datasets",
+            "remove",
+            dataset_id,
+            "1.0.0",
+        ],
+    )
+    assert result.exit_code == 0
+    config = project_mgr.get_by_id(project_id)
+    assert config.model.datasets[0].status == DatasetRegistryStatus.UNREGISTERED
```
elainethale, line 365: Is there anything else to check? It should still work to ask for the dataset_id since it's still in the project config, but are there any other data available from the project that should no longer be accessible and/or be None instead of actual data?

dsgrid/cli/dsgrid_admin.py
```diff
@@ -220,10 +222,28 @@ def remove_project(registry_manager, project_id):
...
+                and dataset.status == DatasetRegistryStatus.REGISTERED
+            ):
+                dataset.status = DatasetRegistryStatus.UNREGISTERED
+                msg = (
+                    f"Set dataset {dataset_id} status to unregistered in project {project_id} "
+                    "after removal."
+                )
+                project_mgr.update(
+                    config, VersionUpdateType.MINOR, msg, submitter=getpass.getuser()
+                )
```
elainethale, line 246: I'm surprised that this is it. Do we need to unlist the data filepath or anything else like that? I was kind of expecting to see some more undo/remove code in project_registry_manager.py

dsgrid/registry/dataset_registry_manager.py
daniel-thom: Using the version for removals was a mistake. We always remove all versions. Removing by specific version would be a larger change. We might not want to ever do that. In most cases we want to remove everything and start over (for project or dataset).

tests/test_registry_management.py
daniel-thom: Derived datasets would also be impacted. But I don't think this should ever be a normal activity outside of the case of "wipe this one out and start over."

dsgrid/cli/dsgrid_admin.py
daniel-thom: On line 228 we call into the dataset registry manager and delete the data.

dsgrid/cli/dsgrid_admin.py
```diff
@@ -219,11 +221,38 @@ def remove_project(registry_manager, project_id):
...
+    for project_id in project_mgr.list_ids():
+        config = project_mgr.get_by_id(project_id)
+        removed_dataset_ids = []
+        for dataset in config.iter_datasets():
+            if (
+                dataset.dataset_id in dataset_ids_set
+                and dataset.status == DatasetRegistryStatus.REGISTERED
+            ):
+                dataset.status = DatasetRegistryStatus.UNREGISTERED
+                removed_dataset_ids.append(dataset.dataset_id)
```
daniel-thom, line 248: This needs to clear the registered dimension mappings for the dataset also.

tests/test_registry_management.py
elainethale: Do we know which derived datasets rely on which initial datasets? Can we remove those too?

elainethale: Check-in conversation: We do not have those tracking data available today.

elainethale (approved):
Dan's going to add the dimension mappings to this PR. Then I think this is okay/good enough for power users to try out.

If we don't have one yet, we should add an issue to track the provenance of derived datasets, so we could remove derived datasets that depend on the dataset being removed as well.

daniel-thom:
> Dan's going to add the dimension mappings to this PR. Then I think this is okay/good enough for power users to try out.
> 
> If we don't have one yet, we should add an issue to track the provenance of derived datasets, so we could remove derived datasets that depend on the dataset being removed as well.

https://github.com/dsgrid/dsgrid/issues/309

PR #324, opened 2024-08-12
Add bulk register CLI command

daniel-thom:
This PR adds the CLI command listed below in order to make it easier to register projects and datasets in bulk.

The logic is mostly existing. We used it in tests. I refactored the code so that we can run it within the dsgrid package.

I also changed the default registry mode to offline. It is unlikely that we will be syncing data to any cloud location in the near future.

```
dsgrid registry bulk-register --help
                                                                                                                                                                                                                                                                
 Usage: dsgrid registry bulk-register [OPTIONS] REGISTRATION_FILE                                                                                                                                                                                               
                                                                                                                                                                                                                                                                
 Bulk register projects, datasets, and their dimensions. The JSON/JSON5 filename must match the data model defined by this documentation:                                                                                                                       
 https://dsgrid.github.io/dsgrid/reference/data_models/project.html#dsgrid.config.registration_models.RegistrationModel                                                                                                                                         
                                                                                                                                                                                                                                                                
╭─ Options ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --base-data-dir  -d  PATH  Base directory for input data. If set, and if the dataset paths are relative, prepend them with this path.                                                                                                                        │
│ --base-repo-dir  -r  PATH  Base directory for dsgrid project/dataset repository. If set, and if the config file paths are relative, prepend them with this path.                                                                                             │
│ --help                     Show this message and exit.                                                                                                                                                                                                       │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

```


dsgrid/cli/registry.py
```diff
@@ -988,6 +998,149 @@ def update_dataset(
...
+    help="Base directory for input data. If set, and if the dataset paths are relative, prepend "
+    "them with this path.",
+)
+@click.option(
+    "-r",
+    "--base-repo-dir",
+    type=click.Path(exists=True),
+    callback=_path_callback,
+    help="Base directory for dsgrid project/dataset repository. If set, and if the config file "
+    "paths are relative, prepend them with this path.",
```
elainethale, line 1017: This is making an assumption about the structure of repositories like https://github.com/dsgrid/dsgrid-project-DECARB?

elainethale: I see ... not really since this is just being applied to the file paths listed in the bulk registration .json.

dsgrid/config/registration_models.py
```diff
@@ -0,0 +1,129 @@
+"""Contains data models to control registration of test projects and datasets."""
```
elainethale, line 1:
```suggestion
"""Contains data models to control bulk registration of projects and datasets."""
```

dsgrid/config/registration_models.py
```diff
@@ -0,0 +1,129 @@
...
+        ),
+    ]
+    replace_dimension_mapping_names_with_ids: Annotated[
+        bool,
+        Field(
+            description="Replace the dimension mapping entries with IDs of dimension mappings "
+            "in the database with matching names. Typically only useful for tests.",
+            default=False,
+        ),
+    ]
```
elainethale, line 52: This is about referring to already-registered dimensions with names instead of IDs? First, the description is a bit unclear on this front. Second, are we sure this is only useful for tests? If a dimension is reused a bunch of times across datasets, could we list it using its name (and nothing else) after it gets defined once (e.g., by the project)?

dsgrid/config/registration_models.py
```diff
@@ -0,0 +1,129 @@
...
+            default=None,
+        ),
+    ]
+    dimension_mapping_references_file: Annotated[
+        Optional[Path],
+        Field(
+            description="Path to file containing references to mappings of dataset-to-project dimensions",
+            default=None,
+        ),
+    ]
```
elainethale, line 36: Should these fields be associated with a specific project_id?

dsgrid/config/registration_models.py
```diff
@@ -0,0 +1,129 @@
...
+            "in the database with matching names. Typically only useful for tests.",
+            default=False,
+        ),
+    ]
+    register_dataset: Annotated[
+        bool, Field(description="If True, register the dataset", default=True)
+    ]
+    submit_to_project: Annotated[
+        bool, Field(description="If True, submit the dataset to the project", default=True)
+    ]
```
elainethale, line 58: Which project?

dsgrid/config/registration_models.py
```diff
@@ -0,0 +1,129 @@
...
+    autogen_reverse_supplemental_mappings: Annotated[
+        set[DimensionType],
+        Field(
+            description="Dimensions on which to attempt create reverse mappings from supplemental dimensions.",
+            default=set(),
+        ),
+    ]
```
elainethale, line 65: When do we do this/why would a user want to do this?

dsgrid/config/registration_models.py
elainethale: I see--this structure goes below the project structure.

But ... if the project_id were supplied here, the bulk register data structure could be flatter, i.e., it could list both projects and datasets, with datasets at the same level as, rather than under, individual projects.

dsgrid/config/registration_models.py
```diff
@@ -0,0 +1,129 @@
...
+
+class ProjectRegistrationModel(DSGBaseModel):
+    """Defines a project to be registered."""
+
+    project_id: Annotated[str, Field(description="Project ID")]
+    config_file: Annotated[Path, Field(description="Path to project.json5")]
+    datasets: Annotated[
+        list[DatasetRegistrationModel],
+        Field(description="List of datasets to register and optionally submit to the project"),
+    ]
```
elainethale, line 99: Per comment above, this could be removed from here and added to the RegistrationModel instead.

dsgrid/cli/dsgrid_admin.py
```diff
@@ -152,16 +152,21 @@ def datasets(registry_manager):
     help="local dsgrid registry data path.",
 )
 @click.option(
-    "-f", "--force", is_flag=True, default=False, help="Delete registry_path if it already exists."
+    "-f",
+    "--overwrite",
+    "--force",
```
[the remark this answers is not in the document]
daniel-thom: This click option makes them synonyms. I've decided that overwrite is more clear, but wasn't hesitant to remove force. I would have to remove it in lots of other places. If you feel strongly, I can (1) revert the change or (2) do a global replace.

dsgrid/cli/registry.py
daniel-thom: Correct. It allows for a git-committed-file to work for different users.

dsgrid/config/registration_models.py
daniel-thom: The `ProjectRegistrationModel` contains a list of `DatasetRegistrationModel` instances, so that is implicit. If this registration file defined a dataset in multiple projects, there would be two `DatasetRegistrationModel` instances.

dsgrid/cli/dsgrid_admin.py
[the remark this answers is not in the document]
elainethale: Should we mark "--force" as a deprecated option that will eventually go away?

dsgrid/config/registration_models.py
daniel-thom: It's for obscure things like this: https://github.com/dsgrid/dsgrid-test-data/blob/main/test_efs/dsgrid_project/datasets/modeled/comstock_unpivoted/dataset.json5#L87

That is based on a very old precedent. What we are doing with real projects and datasets now is better. They always have to explicitly list the paths to dimension record files and dsgrid de-duplicates as needed. It's possible that we could remove this old stuff.

dsgrid/config/registration_models.py
daniel-thom: Sure, we could redesign the structure as part of making this functionality public. We have a few of these files floating around, like https://github.com/dsgrid/dsgrid/blob/main/tests/data/standard_scenarios_registration.json5. It's been this way a long time.

What's the best way to present these concepts to the user? I typically think of projects containing datasets while understanding that dsgrid treats them separately. We haven't had a case yet where two projects share datasets, though I know it's likely.

If I were to make this change, I would do something like this so that it lines with dsgrid completely:
```
{
  projects: [{project_id: "project1", ...}],
  datasets: [{dataset_id: "dataset1", ...}],
  dataset_submissions_to_projects: [
    {
       dataset_id: "dataset1",
       project_id: "project1",
       dimension_mapping_file: "file_path_project1.json5",
       autogen-reverse-supplemental-mappings: false,
    },
    {
       dataset_id: "dataset1",
       project_id: "project2",
       dimension_mapping_file: "file_path_project2.json5"
       autogen-reverse-supplemental-mappings: false
    },
  ],
}
```

daniel-thom: I would add that the main use case for this feature is to register a project and all of its datasets. Another option is to remove this list of project objects from this data structure. The project fields could be at the root level. Could we use it to bulk register many projects and datasets? Sure.

dsgrid/config/registration_models.py
daniel-thom: This is the help provided by the `submit-dataset` CLI command:
```
│    --autogen-reverse-supplemental-mappings  -a  [metric|geography|sector|subsector|time|weather_year|model_year|scenario]  For any dimension listed here, if the dataset's dimension is a project's supplemental dimension and no mapping is provided,       │
│                                                                                                                            create a reverse mapping from that supplemental dimension. 
```

dsgrid/config/registration_models.py
elainethale: This conversation got cut short by a fire alarm, but I think the decision was to move forward with the above, since it is future-proofed.

daniel-thom:
@elainethale @ashreeta I redesigned this command. This [test](https://github.com/dsgrid/dsgrid/pull/324/files#diff-c68f826de13c93ea03f0485b122cc02dd69b6562e33be7a7ab124b0b009c7724R322) demonstrates the workflow.

elainethale (approved):
LGTM

dsgrid/cli/registry.py
```diff
@@ -1252,6 +1263,186 @@ def update_dataset(
...
+)
+@click.pass_obj
+@click.pass_context
+def bulk_register(
+    ctx,
+    registry_manager: RegistryManager,
+    registration_file: Path,
+    base_data_dir: Optional[Path],
+    journal_file: Optional[Path],
+    base_repo_dir: Optional[Path],
```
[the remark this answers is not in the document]
daniel-thom: Thanks. It actually doesn't matter because click passes arguments by name. But I want them to be consistent and will fix.

Issue #321, opened 2024-08-08
Query related - Order of dataset expression / aggregation / filtering


ashreeta:
The order of dataset handling in the query should be modified to allow the expression to be applied after aggregation and filtering?

PR #329, opened 2024-08-14
Support aggregating mapped datasets individually

daniel-thom:
Closes GitHub Issue #321 

## Description
I added a parameter called `aggregate_each_dataset` to the query data model and then use it to
1.	Apply the expression (such as union, add, subtract) to make one dataset and then apply the aggregation.
2.	Apply the aggregation to each dataset and then apply the expression. Required if performing addition or subtraction on datasets with different dimensionality.

`aggregate_each_dataset` is not always valid. Suppose the query drops the subsector and sector dimensions to find total electricity demand. The result will have duplicate values because the aggregation needs to be applied again.

It’s possible that we could determine when a second aggregation is needed, but that could be challenging. What should we do now? I vote for the easy way out and leave this to the query writer. I wouldn’t judge the added complexity to be worthwhile.

What do you think?

### DIFF
--- a/dsgrid/query/models.py
+++ b/dsgrid/query/models.py
@@ -486,6 +486,18 @@ class QueryResultParamsModel(CacheableQueryBaseModel):
             default=[],
         ),
     ]
+    aggregate_each_dataset: Annotated[
+        bool,
+        Field(
+            description="Aggregate each dataset before applying the expression to create one "
+            "overall dataset. The default behavior is to perform one aggregation on the overall "
+            "dataset. This parameter must be set to True for queries that will be adding or "
+            "subtracting datasets with different dimensionality. Warning: setting this value to "
+            "True could produce rows with duplicate dimension combinations when performing a "
+            "union of datasets while also dropping one or more dimensions.",
+            default=False,
+        ),
+    ]
     reports: Annotated[
         list[ReportInputModel],
         Field(description="Run these pre-defined reports on the result.", default=[]),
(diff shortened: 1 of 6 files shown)


ashreeta (approved):
Yes agree, this is more complex than I previously thought.

dsgrid/query/models.py
```diff
@@ -486,6 +486,18 @@ class QueryResultParamsModel(CacheableQueryBaseModel):
...
+    aggregate_each_dataset: Annotated[
+        bool,
+        Field(
+            description="Aggregate each dataset before applying the expression to create one "
+            "overall dataset. The default behavior is to perform one aggregation on the overall "
+            "dataset. This parameter must be set to True for queries that will be adding or "
+            "subtracting datasets with different dimensionality. Warning: setting this value to "
+            "True could produce rows with duplicate dimension combinations when performing a "
+            "union of datasets while also dropping one or more dimensions.",
+            default=False,
```
elainethale, line 498:
```suggestion
            description="If True, aggregate each dataset before applying the expression to create one "
            "overall dataset. This parameter must be set to True for queries that will be adding or "
            "subtracting datasets with different dimensionality. Defaults to False, which corresponds to "
            "the default behavior of performing one aggregation on the overall dataset. WARNING: "
            "For a standard query that performs a union of datasets, setting this value to True could "
            "produce rows with duplicate dimension combinations, especially if one or more "
            "dimensions are also dropped.",
            default=False,
```

elainethale (approved):
LGTM

Issue #207, opened 2023-01-12
Decide how to handle CSV files output from Spark in queries


daniel-thom:
When Spark writes CSV files, it creates a directory and makes one CSV per partition. Readers using Spark won't even notice. Pandas users might find it problematic.

PR #330, opened 2024-08-16
Improve query cache identification

daniel-thom:
Closes GitHub Issue #328 and #207

## Description
The cached queries relied on the project version. If the version changed for reasons that did not affect datasets, the cached versions would still get invalidated. This change makes the identification of the cached queries far more accurate.

### DIFF
--- a/dsgrid/project.py
+++ b/dsgrid/project.py
@@ -1,11 +1,13 @@
 """Interface to a dsgrid project."""
 
+import json
 import logging
 import shutil
 from pathlib import Path
 from typing import Optional
 
 from pyspark.sql import SparkSession
+from semver import VersionInfo
 
 from dsgrid.common import VALUE_COLUMN
 from dsgrid.config.project_config import ProjectConfig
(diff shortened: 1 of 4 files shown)


dsgrid/project.py
```diff
@@ -409,3 +404,38 @@ def get_myear_column(dataset_id):
...
+        """Create a hash that can be used to identify whether the mapping of the dataset to
+        project dimensions can be skipped based on a previous query.
+
+        If a directory with the hash does not already exist, create it and serialize the content
+        used to create the hash.
+
+        Examples of changes that will invalidate the query:
+          - Bump to project major version number
+          - Change to a dataset version
+          - Change to a project's dimension requirements for a dataset
```
elainethale, line 420: Where is this captured? In the project major version number?

dsgrid/project.py
```diff
@@ -409,3 +404,38 @@ def get_myear_column(dataset_id):
...
+          - Change to a dataset dimension mapping
```
elainethale, line 421: Similarly, is this captured in the project major version number?

elainethale (approved):
LGTM

dsgrid/project.py
daniel-thom: It _should_ be captured by the project major version number. In case someone messes up, though, it's caught more explicitly by line 431. That pulls the `InputDatasetModel` from the project config which includes dimension requirements and dimension mappings.

PR #331, opened 2024-08-19
Add example CLI commands

daniel-thom:


dsgrid/cli/query.py
```diff
@@ -90,7 +90,14 @@ def _add_options(func):
 )
 
 
-@click.command("create")
+_create_project_query_epilog = """
+Examples:\n
+$ dsgrid query project create my-result-name my-project-id my-dataset-id\n
+$ dsgrid query project create --default-result-aggregation my-result-name my-project-id my-dataset-id\n
+"""
```
elainethale, line 97:
```suggestion
_create_project_query_epilog = """
Examples:\n
$ dsgrid query project create my-query-result-name my-project-id my-dataset-id\n
$ dsgrid query project create --default-result-aggregation my-query-result-name my-project-id my-dataset-id\n
"""
```

dsgrid/cli/query.py
```diff
@@ -148,7 +147,6 @@ def create_project_query(
     dataset_id,
     filters,
     aggregation_function,
-    default_per_dataset_aggregation,
```
elainethale, line 151: Is this a feature no one is using?

dsgrid/cli/query.py
```diff
@@ -387,7 +391,13 @@ def query_composite_dataset(
     # CompositeDatasetQuerySubmitter.submit(project, output).submit(query, force=force)
 
 
-@click.command()
+_create_derived_dataset_config_epilog = """
+Examples:\n
+$ dsgrid query project create-derived-dataset-config query_output/my_result my_dataset_config\n
+"""
```
elainethale, line 397: I'm not sure what `my_dataset_config` is ... is it a name, a folder, a file name?

`my_result` is also slightly unclear ... should it be `my-query-result-name` (from `_create_project_query_epilog`)?

dsgrid/cli/registry.py
```diff
@@ -119,18 +119,24 @@ def list_(registry_manager):
...
 @click.option(
     "-f",
     "--filter",
     multiple=True,
     type=str,
     help=f"""
-    filter table with a case-insensitive expression in the format 'field operation value',
-    accept multiple flags\b\n
+    Filter table with a case-insensitive expression in the format 'field operation value',
+    accepts multiple flags\b\n
```
elainethale, line 138: Should we also list "valid fields"?

dsgrid/cli/registry.py
```diff
@@ -139,7 +145,13 @@ def list_dimensions(registry_manager, filter):
     registry_manager.dimension_manager.show(filters=filter)
 
 
-@click.command(name="register")
+_register_dimensions_epilog = """
+Examples:\n
+$ dsgrid registry dimensions register -l "Registry dimensions for my-project" dimensions.json5\n
```
elainethale, line 150:
```suggestion
$ dsgrid registry dimensions register -l "Register dimensions for my-project" dimensions.json5\n
```

dsgrid/cli/registry.py
```diff
@@ -149,8 +161,13 @@ def list_dimensions(registry_manager, filter):
 )
 @click.pass_obj
 @click.pass_context
-def register_dimensions(ctx, registry_manager, dimension_config_file, log_message):
-    """Register new dimensions with the dsgrid repository."""
+def register_dimensions(
+    ctx, registry_manager: RegistryManager, dimension_config_file: Path, log_message: str
+):
+    """Register new dimensions with the dsgrid repository. The JSON/JSON5 filename must
+    match the data model defined by this documentation:
```
elainethale, line 168: The filename must match the data model, or the file contents must match the data model?

dsgrid/cli/registry.py
```diff
@@ -160,7 +177,13 @@ def register_dimensions(ctx, registry_manager, dimension_config_file, log_messag
         return 1
 
 
-@click.command(name="dump")
+_dump_dimension_epilog = """
+Examples:\n
+$ dsgrid registry dimensions dump 17565829\n
```
elainethale, line 182: The number at the end is cryptic

elainethale: Ah, it's a dimension id? Is it always a plain int?

elainethale: If it is a plain int, how do people know where to find it? (This is also relevant to the next example CLI call.)

dsgrid/cli/registry.py
```diff
@@ -251,18 +286,23 @@ def update_dimension(
...
 @click.option(
     "-f",
     "--filter",
     multiple=True,
     type=str,
     help=f"""
-    filter table with a case-insensitive expression in the format 'field operation value',
-    accept multiple flags\b\n
+    Filter table with a case-insensitive expression in the format 'field operation value',
+    accepts multiple flags\b\n
```
elainethale, line 304: Similar to list command above, can we list the valid fields? (I would not have expected "Type [From, To]" to work ...

dsgrid/cli/registry.py
```diff
@@ -271,7 +311,13 @@ def list_dimension_mappings(registry_manager, filter):
     registry_manager.dimension_mapping_manager.show(filters=filter)
 
 
-@click.command(name="register")
+_register_dimension_mappings_epilog = """
+Examples:\\
+$ dsgrid registry dimension-mappings list -l "Register dimension mappings for my-project" dimension_mappings.json5\n
```
elainethale, line 316: Is `list` a typo here?

dsgrid/cli/registry.py
```diff
@@ -283,8 +329,13 @@ def list_dimension_mappings(registry_manager, filter):
 )
 @click.pass_obj
 @click.pass_context
-def register_dimension_mappings(ctx, registry_manager, dimension_mapping_config_file, log_message):
-    """Register new dimension mappings with the dsgrid repository."""
+def register_dimension_mappings(
+    ctx, registry_manager: RegistryManager, dimension_mapping_config_file: Path, log_message: str
+):
+    """Register new dimension mappings with the dsgrid repository. The JSON/JSON5 filename must
```
elainethale, line 335: As above, filename or file contents?

dsgrid/cli/registry.py
```diff
@@ -318,17 +375,35 @@ def register_dimension_mappings(ctx, registry_manager, dimension_mapping_config_
...
+        force=force,
     )
     if res[1] != 0:
         return 1
 
 
-@click.command(name="update")
+_update_dimension_mapping_epilog = """
+Examples:\n
+$ dsgrid registry dimension-mappings update -d 17565575 -l "Update dimension mappings" -u major -v 1.0.0 dimension_mappings.json5"\n
```
elainethale, line 402:
```suggestion
$ dsgrid registry dimension-mappings update -d 17565575 -l "Swap out the state to county mapping for my-dataset to that-project" -u major -v 1.0.0 dimension_mappings.json5"\n
```

dsgrid/cli/registry.py
```diff
@@ -362,14 +437,17 @@ def dump_dimension_mapping(ctx, registry_manager, dimension_mapping_id, version,
...
-    version,
+    registry_manager: RegistryManager,
+    dimension_mapping_config_file: Path,
+    dimension_mapping_id: str,
+    log_message: str,
+    update_type: VersionUpdateType,
+    version: str,
 ):
-    """Update an existing dimension mapping registry."""
+    """Update an existing dimension mapping registry. The JSON/JSON5 filename must
```
elainethale, line 447: filename or file contents?

dsgrid/cli/registry.py
```diff
@@ -387,18 +465,23 @@ def update_dimension_mapping(
...
 @click.option(
     "-f",
     "--filter",
     multiple=True,
     type=str,
     help=f"""
-    filter table with a case-insensitive expression in the format 'field operation value',
-    accept multiple flags\b\n
+    Filter table with a case-insensitive expression in the format 'field operation value',
+    accepts multiple flags\b\n
```
elainethale, line 483: Ditto on valid fields

dsgrid/cli/registry.py
```diff
@@ -426,7 +515,10 @@ def register_project(
     project_config_file,
     log_message,
 ):
-    """Register a new project with the dsgrid repository."""
+    """Register a new project with the dsgrid repository. The JSON/JSON5 filename must
```
elainethale, line 518: file contents?

dsgrid/cli/registry.py
```diff
@@ -719,7 +878,16 @@ def register_subset_dimensions(ctx, registry_manager, project_id, filename: Path
         return 1
 
 
-@click.command()
+_register_supplemental_dimensions_epilog = """
+Examples:\n
+$ dsgrid registry projects register-supplemental-dimensions \\ \n
+    -l "Register supplemental dimension for states for my-project-id" \\ \n
```
elainethale, line 884:
```suggestion
    -l "Register states supplemental dimension for my-project-id" \\ \n
```

dsgrid/cli/registry.py
```diff
@@ -820,7 +1004,17 @@ def replace_dataset_dimension_requirements(
         return 1
 
 
-@click.command(name="list-dimension-query-names")
+_list_project_dimension_query_names_epilog = """
+Examples:\n
+$ dsgrid registry projects list-dimension-query-names\n
+$ dsgrid registry projects list-dimension-query-names --exclude-subset\n
+$ dsgrid registry projects list-dimension-query-names --exclude-supplemental\n
+"""
```
elainethale, line 1012: Can you filter by project_id?

dsgrid/cli/registry.py
```diff
@@ -898,18 +1092,23 @@ def list_project_dimension_query_names(
...
 @click.option(
     "-f",
     "--filter",
     multiple=True,
     type=str,
     help=f"""
-    filter table with a case-insensitive expression in the format 'field operation value',
-    accept multiple flags\b\n
+    Filter table with a case-insensitive expression in the format 'field operation value',
+    accepts multiple flags\b\n
```
elainethale, line 1110: valid fields?

dsgrid/cli/registry.py
```diff
@@ -930,7 +1135,10 @@ def list_datasets(registry_manager, filter):
 @click.pass_obj
 @click.pass_context
 def register_dataset(ctx, registry_manager, dataset_config_file, dataset_path, log_message):
-    """Register a new dataset with the dsgrid repository."""
+    """Register a new dataset with the registry. The JSON/JSON5
+    filename must match the data model defined by this documentation:
```
elainethale, line 1139: file contents? Or revert to language like, "Must mach the data model defined by"

dsgrid/cli/registry.py
```diff
@@ -969,7 +1183,17 @@ def dump_dataset(registry_manager, dataset_id, version, directory, force):
...
 
 
-@click.command(name="update")
+_update_dataset_epilog = """
+Examples:\n
+$ dsgrid registry datasets update \\ \n
+    -l "Update the description for dataset my-dataset-id." \\ \n
+    -u patch \\ \n
+    -v 1.0.0 \\ \n
+    my-dataset-id\n
```
elainethale, line 1192: Should a .json5 file be listed here somewhere?

dsgrid/cli/registry.py
```diff
@@ -1004,7 +1228,10 @@ def dump_dataset(registry_manager, dataset_id, version, directory, force):
 def update_dataset(
     ctx, registry_manager, dataset_config_file, dataset_id, log_message, update_type, version
 ):
-    """Update an existing dataset registry."""
+    """Update an existing dataset in the registry. The JSON/JSON5
+    filename must match the data model defined by this documentation:
```
elainethale, line 1232: File contents?

elainethale (approved):
Lots of language suggestions/questions, but generally just thanks for doing this!

dsgrid/cli/query.py
daniel-thom: It was a feature that was removed from the code long ago.

dsgrid/cli/query.py
daniel-thom: The command doctoring is `Create a derived dataset configuration and dimensions from a query result.` `my_dataset_config` will be a directory that contains a dataset.json5 as well as possibly some dimension files.

dsgrid/cli/registry.py
daniel-thom: That's on the next line (ACCEPTED_OPS).

dsgrid/cli/registry.py
daniel-thom: I'll change `filename` to `file.`

dsgrid/cli/registry.py
elainethale: Those are operators/operations. I'm wondering what I can apply the operators to. I.e., in "field operation value", we list the accepted operations, and given field and operation the acceptable values should be relatively obvious, but what fields are valid in this case? Is that information programmatically available here? If not, how does the user figure it out?

dsgrid/cli/registry.py
daniel-thom: Yes, it is always an integer. This is the help message for the command:
```
 Usage: dsgrid registry dimensions dump [OPTIONS] DIMENSION_ID                                                                                                                                  
                                                                                                                                                                                                
 Dump a dimension config file (and any related data) from the registry.
```
`dsgrid registry dimensions list` prints a table like this:
```
+------------------------------------------------------------------------------------------------------------+
|                                                 Dimensions                                                 |
+--------------+------------------------+----------+---------+------------+-----------+----------------------+
| Type         | Query Name             | ID       | Version | Date       | Submitter | Description          |
+--------------+------------------------+----------+---------+------------+-----------+----------------------+
| geography    | county                 | 17681312 | 1.0.0   | 2024-08-19 | dthom     | Initial registration |
|              |                        |          |         | 11:01:14   |           |                      |
```
Is this sufficient?

dsgrid/cli/registry.py
daniel-thom: The fields are the columns printed in the table. Presumably, the user will run the command at least once and display all rows.

dsgrid/cli/registry.py
daniel-thom: project_id is a required argument. I apparently forgot to include it.

dsgrid/cli/registry.py
daniel-thom: The fields are not available programmatically. They are hard-coded in each of the registry manager files.

elainethale: Got it. Would `column` or `table column` be more clear than `field`?

dsgrid/cli/registry.py
elainethale: Does/could the description of "DIMENSION_ID" say something like "ID listed in the table output by 'dsgrid registry dimensions list'" or "dsgrid registry ID for the dimension, as listed in the table output by 'dsgrid registry dimensions list'"?

daniel-thom: click doesn't provide an explicit way of defining help for a positional argument. We could write that text in the command docstring. The implication of what I think you're asking is that for every command that takes project_id/dataset_id/dimension_id/mapping_id, I would add text redirecting to the list command. For example, this is one project command:

```
 Usage: dsgrid registry projects list-dimension-query-names [OPTIONS] PROJECT_ID                                                                                                                
                                                                                                                                                                                                
 List the project's dimension query names.
```
Change to
```
 Usage: dsgrid registry projects list-dimension-query-names [OPTIONS] PROJECT_ID                                                                                                                
                                                                                                                                                                                                
 List the project's dimension query names. Run "dsgrid registry projects list" to display a table of all project IDs.
```
Is that what you're asking for?

elainethale: I'm most concerned about the _ID being defined if it's an ArangoDB ID rather than the dataset_ids and project_ids that are in the config files. I.e., I find the DIMENSION_ID arguments the least intuitive. But maybe I haven't spent enough time working with the CLI ...

daniel-thom: Regardless of whether the ID is a human-readable string or integer, the CLI user needs to know how to find out what they are. We do not expect that users will dig through config files on GitHub to find them, right?

Here are the options as I see them:
1. The CLI is sufficiently intuitive as it is. The user observes that each of the projects/datasets/dimensions/dimension-mappings command groups has register, list, dump, and update commands, and understands how they are connected.
2. The main registry CLI docstring describes how the commands are connected. Repo documentation could also describe it, such as [here](https://dsgrid.github.io/dsgrid/reference/cli_fundamentals.html). But people almost never read this stuff. 
3. In every instance where the CLI refers to an ID in a docstring, it refers to the appropriate list command.

elainethale: On the integer dimension and dimension IDs: I agree this is the best option because we don't want to require uniqueness.

Suggestion 3 might be most helpful for users, but I understand that it's kind of a pain to maintain--what do you think?

I do also think it would be worth revising the documentation page you link (Suggestion 2) to place more emphasis on things like "how do you know what ID to use" and less on "how do you most efficiently configure your shell."

elainethale: Let's just implement 2 for now. I'm not a frequent enough user to be a good judge of whether 3 is needed, and I am persuaded by the too much information/clutter argument. Presumably people will start by learning how to `list` things ...

daniel-thom: I added content for 2. Let me know what you think. I think I addressed all other comments as well.

docs/source/reference/cli_fundamentals.rst
```diff
@@ -16,10 +16,37 @@ The dsgrid CLI commands are hierarchical with help at every level. For example:
...
    $ dsgrid query project run --help
 
+Registry CLI Commands
+---------------------
+The ``dsgrid registry`` command group has a subgroup for each type of configuration stored in the
+registry: ``projects``, ``datasets``, ``dimensions``, and ``dimension-mapppings``.
+
+Each of those subgroups has four main commands:
+
+- ``register``: Register a new configuration item in the registry. This command typically takes a
```
elainethale, line 34:
```suggestion
- ``register``: Register a new item (project, dataset, dimension, or dimension-mapping). This command typically takes a
```

docs/source/reference/cli_fundamentals.rst
```diff
@@ -16,10 +16,37 @@ The dsgrid CLI commands are hierarchical with help at every level. For example:
...
+  JSON5 file as an input. After registration, all other commands will refer to the item by its ID.
+  Projects and datasets have a user-created string as their ID (``project_id`` and ``dataset_id``)
+  stored in the JSON5 file. Dimensions and dimension mappings receive integer IDs from the registry
+  during registration.
```
elainethale, line 38:
```suggestion
  JSON5 file as an input. After registration, all other commands will refer to the item by its ID.
  Projects and datasets have a user-created string as their ID (``project_id`` and ``dataset_id``),
  which is defined in the original JSON5 file. Dimensions and dimension mappings receive integer 
  IDs from the registry during registration.
```

docs/source/reference/cli_fundamentals.rst
```diff
@@ -16,10 +16,37 @@ The dsgrid CLI commands are hierarchical with help at every level. For example:
...
+- ``list``: List configuration items stored in the registry in a table. Each command offers the
```
elainethale, line 39: Change "configuration items" to just "items"? Or maybe change to "item configurations"

dsgrid/cli/registry.py
elainethale: Looks good. I made a few minor editing suggestions.

elainethale (approved):

docs/source/reference/cli_fundamentals.rst
daniel-thom: I used configuration item because we refer "project config" and "dataset config" all over the place. I could be ok with just item. Document is another option. That is what the db uses.

elainethale: To me it seems most natural to just talk about "items," i.e., projects, datasets, etc. that are registered with dsgrid. 

I get that, for example, only the dataset configuration is available on the command line through the registry (and that, for example, the actual data is not so near at hand), but I think for users that's all just one thing, i.e., the "dataset", that is "in dsgrid" and by browsing the registry they are browsing the "dataset," not its "configuration" or "document". Does that make sense/seem right?

daniel-thom: I'll go with that. We may need to revisit the documentation where we often talk about "project config" and "dataset config." The documentation should probably focus most on the "dataset contributor" users and be consistent with terms. It makes sense to "go look at the project for dataset requirements" instead of "go look at the project config".

PR #334, opened 2024-10-31
Update pydantic models to v2.9

daniel-thom:
There is no functional change in this PR.

- Update Pydantic to v2.9.
- Simplify definition dsgrid data models in conformance with latest pydantic. Mostly, remove verbose uses of Annotated.
- Replace uses of `typing.List` and `typing.Dict` with `list` and `dict`.
- Remove use of `Optional` when there is a default value - those cases should not need to support a value of None.


elainethale (approved):
LGTM. I like the streamlined interface. If I'm recalling correctly, this is more like an earlier version? (I seem to remember a previous PR changing all the "Field"s to "Annotated"s ...)

daniel-thom:
> LGTM. I like the streamlined interface. If I'm recalling correctly, this is more like an earlier version? (I seem to remember a previous PR changing all the "Field"s to "Annotated"s ...)

Yes, that's true. I switched to Annotated when I upgraded from pydantic 1.x to 2. Type hinting didn't work correctly. I'm not sure if that was my fault or some incompatibility with the initial version. Anyway, everything works well now. I did have to convert a few things to new pydantic 2.9 formats.

PR #338, opened 2024-12-20
Remove AWS packages

daniel-thom:


.github/workflows/codecov.yml
```diff
@@ -9,6 +9,7 @@ jobs:
     name: Codecov
     runs-on: ubuntu-latest
     steps:
+    - uses: actions/checkout@v4
```
daniel-thom, line 12: I inadvertently removed this line in the previous PR.

.github/workflows/codecov.yml
```diff
@@ -23,7 +24,7 @@ jobs:
         # Credentials would need to be configured.
         git submodule init
         git submodule update
-        python -m pytest -v --disable-warnings --cov=./ --cov-report=xml:coverage.xml tests
+        python -m pytest -v --disable-warnings --cov=./ --cov-report=xml:coverage.xml tests dsgrid/utils/py_expression_eval/tests.py
```
elainethale, line 27: What is `dsgrid/utils/py_expression_eval/tests.py` about? Also, I don't see that file in this PR, does it already exist?

elainethale (approved):
One question that's probably not important. Otherwise LGTM

PR #350, opened 2025-04-29
Generate project and dataset config files

daniel-thom:
This PR enables a workflow to generate a filesystem layout with starting project and dataset config files. The dataset dimension files are generated from either the dataset itself or matching dimension IDs from the registry.
```bash
$ dsgrid registry projects generate-config my_project comstock resstock
```
```bash
$ tree my_project
my_project
├── datasets
│   ├── historical
│   └── modeled
└── project
    ├── dimension_mappings
    ├── dimensions
    │   ├── subset
    │   └── supplemental
    └── project.json5
```
```bash
$ dsgrid registry datasets generate-config comstock dsgrid-test-data/datasets/test_efs_comstock_unpivoted -o my_project/datasets/modeled
```
```bash
$ tree my_project
my_project
├── datasets
│   ├── historical
│   └── modeled
│       └── comstock
│           ├── dataset.json5
│           └── dimensions
│               ├── geography.csv
│               ├── metric.csv
│               ├── sector.csv
│               └── subsector.csv
└── project
    ├── dimension_mappings
    ├── dimensions
    │   ├── subset
    │   └── supplemental
    └── project.json5
```

### DIFF
--- a/dsgrid/config/common.py
+++ b/dsgrid/config/common.py
@@ -1,0 +1,31 @@
+from typing import Any
+
+from dsgrid.dimension.base_models import DimensionType
+
+
+def make_base_dimension_template(
+    exclude_dimension_types: set[DimensionType] | None = None,
+) -> list[dict[str, Any]]:
+    exclude: set[DimensionType] = exclude_dimension_types or set()
+    dimensions = [
+        {
+            "type": x.value,
+            "class": x.value.title(),
+            "name": "",
+            "description": "",
+            "file": f"dimensions/{x.value}.csv",
+            "module": "dsgrid.dimension.standard",
+        }
+        for x in DimensionType
+        if x not in exclude
+    ]
+    dimensions.append(
+        {
+            "type": DimensionType.TIME.value,
+            "time_type": "",
+            "class": DimensionType.TIME.value.title(),
+            "name": "",
+            "description": "",
+        }
+    )
+    return dimensions
--- a/dsgrid/config/common.py
+++ b/dsgrid/config/common.py
@@ -1,12 +1,15 @@
 from typing import Any
 
 from dsgrid.dimension.base_models import DimensionType
+from dsgrid.dimension.time import MeasurementType, TimeDimensionType, TimeIntervalType
 
 
 def make_base_dimension_template(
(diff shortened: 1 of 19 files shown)


dsgrid/config/common.py
```diff
@@ -0,0 +1,31 @@
+from typing import Any
+
+from dsgrid.dimension.base_models import DimensionType
+
+
+def make_base_dimension_template(
+    exclude_dimension_types: set[DimensionType] | None = None,
```
elainethale, line 7: Should this default to `{DimensionType.TIME}`?

dsgrid/config/common.py
```diff
@@ -0,0 +1,31 @@
...
+        if x not in exclude
+    ]
+    dimensions.append(
+        {
+            "type": DimensionType.TIME.value,
+            "time_type": "",
+            "class": DimensionType.TIME.value.title(),
+            "name": "",
+            "description": "",
+        }
```
elainethale, line 29: This template is much less comprehensive than the one provided for the file-based dimension types. Could there be an option to specify which time_type you want to use?

tests/cli/test_dataset_config_generation.py
```diff
@@ -0,0 +1,135 @@
...
+        dataset_id,
+        f"dsgrid-test-data/datasets/test_efs_{dataset_id}",
+        "--project-id",
+        "test_efs",
+        "--schema-type",
+        "standard",
+        "--pivoted-dimension-type",
+        "metric",
+        "--time-columns",
+        "timestamp",
```
elainethale, line 31: This is the default, so could (also) test that the command works without this option.

dsgrid/config/common.py
daniel-thom: You should generally never set a default function argument to a mutable instance because that argument is only evaluated once.

dsgrid/config/common.py
daniel-thom: I expanded the template.

tests/cli/test_dataset_config_generation.py
daniel-thom: The next dataset does that.

PR #352, opened 2025-05-02
Allow users to customize the dataset mapping plan in a query

daniel-thom:


dsgrid/dataset/dataset_schema_handler_one_table.py
```diff
@@ -113,7 +114,7 @@ def _check_one_table_data_consistency(self):
     def make_dimension_association_table(self) -> DataFrame:
         dim_cols = self._list_dimension_columns(self._load_data)
         df = self._load_data.select(*dim_cols).distinct()
-        df = self._remap_dimension_columns(df)
+        df = self._remap_dimension_columns(df, self.build_default_dataset_mapping_plan())
```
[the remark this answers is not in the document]
elainethale: And same question seems to pop up more below.

dsgrid/utils/dataset.py
```diff
@@ -402,10 +402,23 @@ def repartition_if_needed_by_mapping(
     df: DataFrame,
     mapping_type: DimensionMappingType,
     scratch_dir_context: ScratchDirContext,
-) -> DataFrame:
-    """Repartition the dataframe if the mapping might cause data skew."""
+    repartition: bool | None = None,
+) -> tuple[DataFrame, Path | None]:
+    """Repartition the dataframe if the mapping might cause data skew.
```
elainethale, line 407: Define data skew? (I don't recall exactly what you mean by the term ... something about natural partitions being vastly different sizes???)

dsgrid/query/dataset_mapping_plan.py
```diff
@@ -0,0 +1,83 @@
...
+class MapDimensionOperation(DSGBaseModel):
+
+    dimension_name: str = Field(
+        description="Name of the project dimension to which the dataset will be mapped.",
+    )
+    handle_data_skew: bool | None = Field(
+        default=None,
+        description="Use a salting technique to handle data skew in this mapping "
+        "operation. This will automatically trigger a persist to the filesystem. "
+        "If the value is None, dsgrid will determine what to do at runtime.",
```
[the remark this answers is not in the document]
elainethale: I see below that our default is to `handle_data_skew` for `DimensionMappingType.ONE_TO_MANY_DISAGGREGATION`. I don't think we want to list anything that specific here, but maybe: `If the value is None, dsgrid will use a salting technique when applying DimensionMappingTypes that have caused data skew issues in the past."

tests/test_queries.py
```diff
@@ -819,6 +837,80 @@ def validate(self, expected_values=None):
...
+                project_id=self.get_project_id(),
+                include_dsgrid_dataset_components=False,
+                dataset=DatasetModel(
+                    dataset_id="tempo_mapped",
+                    source_datasets=[
+                        StandaloneDatasetModel(dataset_id="tempo_conus_2022"),
+                    ],
+                ),
+                mapping_plan=[
+                    DatasetMappingPlan(
```
elainethale, line 861: Question in a strange place because I missed it above: When does time mapping happen?

tests/test_queries.py
```diff
@@ -819,6 +837,80 @@ def validate(self, expected_values=None):
...
+                        aggregation_function="sum",
+                    ),
+                ],
+                output_format="parquet",
+                table_format=UnpivotedTableFormatModel(),
+            ),
+        )
+        return self._model
+
+    def validate(self, expected_values=None) -> bool:
```
elainethale, line 910: A few questions about these tests:
- I assume this test is working on a small version of tempo_conus_2022? 
- Can we test some expected_values?
- Should we also verify that mapping in a different order doesn't change the result (which would also require comparing to expected_values)?
- Would it be worthwhile to test that the journal is logging dimension names in the expected order?

PR #354, opened 2025-06-03
Convert model_year and weather_year from int to str

daniel-thom:
This allows dataset contributors to partition their datasets on model year and weather year. As-is, dataset registration will fail because Spark interprets the columns as integers. Also, contributors frequently set the schema of these columns as integers and we can be forgiving.


elainethale (approved):
LGTM. 

Overall, this seems to make us robust to reading in years that we expect to be string record ids but are currently stored as integers. Is there anything we should be doing on the back end to sometimes save years out as integers? For example, can dsgrid data output tables be partitioned on model_year?

Issue #353, opened 2025-05-22
Make it easier for dataset submitters to acknowledge missing data


elainethale:
Right now, it's often the case that submitting a data to project results in an error because there are unexpected missing data. dsgrid outputs a file containing all of the missing data combinations, but to resolve the error, data submitters have to go back to the original dataset and add NULLS, which essentially requires a manual mapping from project dimensions back to dataset dimensions. 

We want to alert dataset submitters to data that are missing, in case any of the missing data indicate an error, but more often than not the missing data basically are expected, e.g., because a specific subsector is not present in a particular county. Therefore, it would be good to make it easier than it is currently to resolve this type of error.

One idea for doing this (credit to @ashreeta) is to be able to specify the file of missing data combinations *in the project base dimensions, i.e., the file that dsgrid outputs* in the `dataset_requirements` section of `project.json`.

A drawback of this idea is that the current pattern would result in this file getting incorporated directly into the `project.json`, such that the registered version of that file would become much bigger. **Is that acceptable, or might it be possible in this case to register a supplemental data file alongside the registered `project.json`?**

daniel-thom:
Let's make sure we are clear on what this means. The simplest solution that meets the request is that we add a new mechanism to specify expected-missing data.

During dataset submission, the user specifies the path to a Parquet file that contains one row per combination of dimensions that are expected-missing. dsgrid stores this file in a project directory in the registry. Note: dsgrid already creates this table in CSV format when dataset submission fails because of missing data. The user would have to confirm its contents and then rewrite it as a Parquet file. We would add the file path to project config so that we know what happened.

An alternate solution is that dsgrid attempts to create the multi-dimensional section of the project configuration file from the missing records. Producing an optimized JSON object programmatically might not be straightforward. In the worst case, this is a JSON file in the hundreds of MBs. The file created by @ashreeta for the industry dataset is 337 MBs. We would store this object in our SQLite registry database. We would also likely add it to the project's git repository. We really don't want to do either of those things.

daniel-thom:
Another idea is that the project submitter is allowed to pass a flag to the dataset submission command saying, "I have confirmed that the missing dimension combinations are expected." dsgrid then writes the table of missing records to the registry. The result is the same as the solution above with fewer user steps.

daniel-thom:
One more comment:

[...]

PR #361, opened 2025-07-31
Add support for datasets isolated from projects

### LINKED ISSUE #148: Optimize handling of null values in load_data_lookup
Whenever we load a registered dataset with a load_data_lookup.parquet we have to filter out rows with NULL data `id` values. Those are needed to validate intentionally excluded dimension combinations during dataset registration. Per discussion with @lixiliu we could do two things to make this better:
1. At registration time, after validation and before we copy the file to the registry, strip the NULLs from the dataframe and write them to a secondary file. That secondary file would not be used by dsgrid; it would only provide a way to retrieve the original dataset.
2. Allow the dataset contributor to provide the secondary file during registration.

### LINKED ISSUE #301: Dataset-to-project mappings should use the registry graph
When we map a dataset to user-requested project supplemental dimensions in a query, we use this simplistic procedure:
- Map from dataset to the project's base dimension.
- Map from project's base dimension to project's supplemental dimension.

In many cases, this is fine. We either need to go through both steps or want to have a cached version of a dataset in the project's base dimensions.

In other cases, it's a bit silly. If a dataset has state-level values, the project's base dimension is county, and the user wants state-level data, we will disaggregate the data to county level and then immediately aggregate back to state. This is expensive and time-consuming.

Here is a proposal to make this more efficient - at least for mappings involving only two dimensions.
1. When we store a dimension mapping, create a `to` edge connecting the two dimensions.
2. Store the mapping type, metadata, and records in the edge as a JSON document (ArangoDB supports this). We currently store dimension mappings as vertexes, and that may need to change.
3. When we map a dataset to a project in a query, instead of mapping to the project's base dimension, submit a shortest-path query to ArangoDB, and map to each vertex returned.
4. We will likely want an option to choose the current behavior or the new behavior.

daniel-thom:
- Map datasets to an arbitrary list of destination dimensions, includes support for multiple hops. (Fixes #301)
- Check dataset cross-joins at dataset registration time. (#353)
- If expected errors exist, allow the user to pass the error file created by the registration process as an override to a second run.
- Store the union of NULL rows in the data files (existing feature) with the new files in the registry. NULLs will now be removed from data files during queries (which could be debated). This would fully fix #148 
- Removed requirements for all datasets to have all dimension types. Projects continue to require that datasets have all dimension types.
- Add a new interface for the registry data store. Default is our existing filesystem store. Added a DuckDB store. This allows parent applications to run queries efficiently.


dsgrid/cli/registry.py
```diff
@@ -1336,6 +1455,12 @@ def dump_dataset(registry_manager, dataset_id, version, directory, force):
     type=str,
     help="reason for submission",
 )
+@click.option(
```
daniel-thom, line 1458: I'll note a few alternative implementations.
1. Allow the user to place the missing_associations file in the same directory as the load_data.parquet. We could look for it during registration.
2. The dataset.json5 could specify the path to the missing_associations file.

The result of the registration process is a directory in the registry containing the parquet files and this CSV, so there is a strong argument for 1. The current implementation is based on this idea:

- User runs `dsgrid registry datasets register`
- It fails and dumps the missing associations to a file in the current directory.
- User runs `dsgrid registry datasets register --missing-association=xyz.csv`

I wouldn't be opposed to instructing them to copy the file to their dataset directory instead.

dsgrid/dimension/base_models.py
```diff
@@ -150,6 +149,37 @@ def check_required_dimensions(dimensions, tag):
...
+    tag : str
+        User-defined string to include in exception messages
+
+    Raises
+    ------
+    ValueError
+        Raised if a required dimension is not provided.
+
+    """
+    # dimension_types = {x.dimension_type for x in dimensions}
```
daniel-thom, line 168: We talked about requiring a subset of dimensions, but I'm not sure that any are absolutely required.

dsgrid/dataset/dataset_schema_handler_one_table.py
```diff
@@ -167,7 +228,48 @@ def make_project_dataframe(
...
+                to_geo_dim=project_config.get_base_dimension(DimensionType.GEOGRAPHY),
             )
             return self._finalize_table(context, ld_df, project_config)
+
+    def make_mapped_dataframe(
+        self,
+        context: QueryContext,
+        geography_dimension: DimensionBaseConfigWithFiles | None = None,
+        metric_dimension: DimensionBaseConfigWithFiles | None = None,
+        time_dimension: TimeDimensionBaseConfig | None = None,
```
elainethale, line 248: Can the user not map sector, subsector, end use, model_year, scenario? Why these three specific dimension types?

elainethale (commented):
From the high-level description, the only thing that gives me pause is:

- Removed requirements for all datasets to have all dimension types. **Projects continue to require that datasets have all dimension types.**

I understand that we don't want to change the project behavior right now. Would it make sense to add an issue with a title like, "Enable projects to use datasets that are missing some dimension types"? It seems like a dataset not having a dimension type is similar to a trivial dimension, and that that information could be specified in a project-specific way upon submission ... ?

I'm ~1/3 of the way through the files. The code mostly looks good so far, but I don't know why dataset mapping is restricted to geography, metric, and time.

dsgrid/cli/registry.py
```diff
@@ -356,7 +389,7 @@ def register_dimension_mappings(
 
 _dump_dimension_mapping_epilog = """
 Examples:\n
-$ dsgrid registry dimension-mappings dump 17565575\n
+$ dsgrid registry dimension-mappings dump 8c575746-18fa-4b65-bf1f-516079d634a5\n
```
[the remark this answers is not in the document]
daniel-thom: The friendly IDs were generated by ArangoDB.

dsgrid/cli/registry.py
daniel-thom: I don't think that we would want to proactively write to their dataset directory. The errors might be real. The user should make the choice to use the overrides.

I started with the file in the current directory because that's what we discussed. I thought of the file in the dataset directory later. Here are my pros and cons:

- Current directory: Simplest for the user. They inspect the file and then re-run the command without copying/moving files. Downside is that we have to pipe the CLI option everywhere...and the user has to type it. 
- Dataset directory: We are moving toward a workflow where a user could have all these files in the dataset directory: `load_data.csv`, `load_data_schema.csv`, `load_data_lookup.csv`, `load_data_lookup_schema.csv`, `missing_associations.csv`. For a user building a new dataset, this is all consistent and no additional CLI args are required. The only downside is that the user has to copy the file, which isn't a big deal. I mostly like this option.

dsgrid/dataset/dataset_schema_handler_one_table.py
daniel-thom: They can. We have special handling for these three. We may add time zone through geography. We convert units through metric. Time is its own thing. The others don't have special handling.

daniel-thom:
> From the high-level description, the only thing that gives me pause is:
> 
> * Removed requirements for all datasets to have all dimension types. **Projects continue to require that datasets have all dimension types.**
> 
> I understand that we don't want to change the project behavior right now. Would it make sense to add an issue with a title like, "Enable projects to use datasets that are missing some dimension types"? It seems like a dataset not having a dimension type is similar to a trivial dimension, and that that information could be specified in a project-specific way upon submission ... ?
> 
> I'm ~1/3 of the way through the files. The code mostly looks good so far, but I don't know why dataset mapping is restricted to geography, metric, and time.

It is critical for all datasets in a project to have the same columns because we will combine them together in one table. How do we add those columns for datasets that don't have a dimension? I could imagine solutions other than trivial dimensions. Note: I consider a trivial dimension to be the case where the dataset _has_ a dimension, but simply don't want to specify it because it is obvious. For example, there is no reason for transportation data to include a sector column. Not having a dimension seems like a different case to me.

I wonder if we could expand the project config to specify how datasets can be combined in a more flexible manner. We don't need complete overlap in all cases.

dsgrid/cli/registry.py
elainethale: The latter does seem good in that the directions are straightforward--put the required files in a single folder. I'm fine with requiring that at this point if that's what you'd prefer as well.

elainethale (approved):
LGTM

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -155,19 +305,31 @@ def make_project_dataframe(self, context, project_config) -> DataFrame:
...
 
+    @abc.abstractmethod
+    def make_mapped_dataframe(
+        self,
+        context: QueryContext,
+        geography_dimension: DimensionBaseConfigWithFiles | None = None,
+        metric_dimension: DimensionBaseConfigWithFiles | None = None,
+        time_dimension: TimeDimensionBaseConfig | None = None,
+    ) -> DataFrame:
+        """Return a load_data dataframe with dimensions mapped as specified by the QueryContext."""
```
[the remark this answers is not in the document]
daniel-thom: I take it back. The code can be improved. We can lookup these "to dimensions" later. We don't need to pass them through the stack. It's a couple of extra calls but leads to better readability.

dsgrid/cli/registry.py
daniel-thom: This is now implemented.

dsgrid/registry/registry_manager.py
```diff
@@ -407,6 +414,8 @@ def copy(
         """
         src_db = RegistryDatabase.connect(src)
         src_data_path = src_db.get_data_path()
+        # TODO: This does not support the duckdb data store. Need to implement this copy operation
+        # in the DataStoreInterface.
```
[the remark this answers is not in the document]
daniel-thom: No. I created #368

dsgrid/query/query_submitter.py
```diff
@@ -648,6 +672,158 @@ def _save_composite_dataset(self, context: QueryContext, df, repartition):
...
+        to_dimension_mapping_refs: list[DimensionMappingReferenceModel] = []
+        mapped_dimension_types: set[DimensionType] = set()
+        dimensions: dict[DimensionType, DimensionBaseConfig] = {}
+        with mgr.dimension_mapping_manager.db.engine.connect() as conn:
+            graph = mgr.dimension_mapping_manager.build_graph(conn=conn)
+            for to_dim_ref in query.to_dimension_references:
+                to_dim = mgr.dimension_manager.get_by_id(
+                    to_dim_ref.dimension_id, version=to_dim_ref.version
+                )
+                if to_dim.model.dimension_type in mapped_dimension_types:
```
lixiliu, line 735: What if a user wants an aggregation at PCA region by state?

dsgrid/registry/bulk_register.py
```diff
@@ -0,0 +1,161 @@
...
+            tmp_files,
+            base_data_dir,
+            base_repo_dir,
+            journal,
+        )
+    except Exception:
+        failure_occurred = True
+        raise
+    finally:
+        if failure_occurred and journal.has_entries():
```
lixiliu, line 59: is it possible to have `failure_occurred and not journal.has_entries()`

dsgrid/cli/registry.py
```diff
@@ -1149,6 +1240,13 @@ def list_project_dimension_names(
 @click.command(name="generate-config", epilog=_generate_project_config_epilog)
 @click.argument("project_id")
 @click.argument("dataset_ids", nargs=-1)
+@click.option(
```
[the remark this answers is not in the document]
daniel-thom: I made it plural because you can enter the option multiple times. It is a bit confusing, but it probably better to  optimize for the user instead of variable names in the code. I'll change it.

dsgrid/cli/registry.py
```diff
@@ -1476,6 +1610,7 @@ def generate_dataset_config_from_dataset(
     dataset_id: str,
     dataset_path: Path,
     schema_type: DataSchemaType,
+    metric_type: str,
```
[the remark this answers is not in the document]
daniel-thom: These are actually names of metric classes, not an enum.

PR #367, opened 2025-08-11
Add support for CSV schema

daniel-thom:
This PR will allow users to use CSV files for any data or lookup table. There is one caveat: they have to supply a schema file in the same directory as the data/lookup file. My reasoning is this: dsgrid's default behavior is to use county FIPS codes that look like integers but are actually strings that have leading zeros. These are a constant source of error. If we rely on schema inference from Spark and DuckDB, our users will constantly get errors and contact us for help. It will save everyone a lot of time if we require the schema. I don't think that this is a big ask.

If we can't require the schema, I'll ask if we can stop using these problematic FIPS codes and instead follow what the buildings teams do - prefix the codes with a letter.


ashreeta (approved):
Can we support timestamp columns that are strings or objects? Do they *have* to be timestamps with or without a timezone? Also geography is not the limiting factor for data submission, a lot of the examples of raw data actually have some version of county names listed for the county. Its usually the mapping file that is the issue since that needs to contain the dsgrid convention for counties.

daniel-thom:
> Can we support timestamp columns that are strings or objects? Do they _have_ to be timestamps with or without a timezone? Also geography is not the limiting factor for data submission, a lot of the examples of raw data actually have some version of county names listed for the county. Its usually the mapping file that is the issue since that needs to contain the dsgrid convention for counties.

You seem to be asking about timestamps in Parquet files, which is outside the scope of this PR. We do not currently support timestamps as strings in Parquet files. We've discussed it. It could be considered as a new issue.

ashreeta:
> > Can we support timestamp columns that are strings or objects? Do they _have_ to be timestamps with or without a timezone? Also geography is not the limiting factor for data submission, a lot of the examples of raw data actually have some version of county names listed for the county. Its usually the mapping file that is the issue since that needs to contain the dsgrid convention for counties.
> 
> You seem to be asking about timestamps in Parquet files, which is outside the scope of this PR. We do not currently support timestamps as strings in Parquet files. We've discussed it. It could be considered as a new issue.

No I have some examples of csv files with the timestamp column as a string.

daniel-thom:
> > > Can we support timestamp columns that are strings or objects? Do they _have_ to be timestamps with or without a timezone? Also geography is not the limiting factor for data submission, a lot of the examples of raw data actually have some version of county names listed for the county. Its usually the mapping file that is the issue since that needs to contain the dsgrid convention for counties.
> > 
> > 
> > You seem to be asking about timestamps in Parquet files, which is outside the scope of this PR. We do not currently support timestamps as strings in Parquet files. We've discussed it. It could be considered as a new issue.
> 
> No I have some examples of csv files with the timestamp column as a string.

Can you elaborate? Timestamps in CSV files are always strings.

daniel-thom:
> > > > > Can we support timestamp columns that are strings or objects? Do they _have_ to be timestamps with or without a timezone? Also geography is not the limiting factor for data submission, a lot of the examples of raw data actually have some version of county names listed for the county. Its usually the mapping file that is the issue since that needs to contain the dsgrid convention for counties.
> > > > 
> > > > 
> > > > You seem to be asking about timestamps in Parquet files, which is outside the scope of this PR. We do not currently support timestamps as strings in Parquet files. We've discussed it. It could be considered as a new issue.
> > > 
> > > 
> > > No I have some examples of csv files with the timestamp column as a string.
> > 
> > 
> > Can you elaborate? Timestamps in CSV files are always strings.
> 
> Ok, I am referring to the example json file where you have the timestamp column as a 'timestamp' with or without the timezone. If that column is a string in the csv doesn't it make sense that the data submitter will indicate it is a string? .. code-block:: json { "timestamp": "TIMESTAMP_TZ", "geography": "STRING", "scenario": "STRING", "subsector": "STRING", "value": "FLOAT", }

In CSV the initial type is always string. The schema file designates the final type. After reading the CSV file into a Spark DataFrame or DuckDB table, we want the type of the column to be a timestamp, not a string.

daniel-thom:
> > > > > > > Can we support timestamp columns that are strings or objects? Do they _have_ to be timestamps with or without a timezone? Also geography is not the limiting factor for data submission, a lot of the examples of raw data actually have some version of county names listed for the county. Its usually the mapping file that is the issue since that needs to contain the dsgrid convention for counties.
> > > > > > 
> > > > > > 
> > > > > > You seem to be asking about timestamps in Parquet files, which is outside the scope of this PR. We do not currently support timestamps as strings in Parquet files. We've discussed it. It could be considered as a new issue.
> > > > > 
> > > > > 
> > > > > No I have some examples of csv files with the timestamp column as a string.
> > > > 
> > > > 
> > > > Can you elaborate? Timestamps in CSV files are always strings.
> > > 
> > > 
> > > Ok, I am referring to the example json file where you have the timestamp column as a 'timestamp' with or without the timezone. If that column is a string in the csv doesn't it make sense that the data submitter will indicate it is a string? .. code-block:: json { "timestamp": "TIMESTAMP_TZ", "geography": "STRING", "scenario": "STRING", "subsector": "STRING", "value": "FLOAT", }
> > 
> > 
> > In CSV the initial type is always string. The schema file designates the final type. After reading the CSV file into a Spark DataFrame or DuckDB table, we want the type of the column to be a timestamp, not a string.
> 
> Ok thanks for that. It would have been helpful (to me atleast) the comments or explanation including this sentence where the schema file designates the final type. If it does designate the final type, why don't we just set the final type to what dsgrid can read in by default? Why does the data submitter need to specify this information when they might not have any clue as to the 'final type'?

I really hope the user knows what the expected types are. dsgrid will not have any clue. The other option I mentioned is to let Spark/DuckDB infer the schema. The result will be errors with county IDs cast as integers. I'm not completely opposed to this option - but would prefer we do something different with county FIPS codes.

PR #369, opened 2025-08-13
Refactor expected missing associations

daniel-thom:
This is an addendum to #361. @lixiliu and I arrived at this simplification:

While we will continue to support null rows in load_data/load_data_lookup data tables, the behavior is different.
- If nulls are present, filter them out when writing the tables to the registry.
- Split the null rows into a missing_associations table.
- The user is not allowed to provide null rows and a missing associations table.


dsgrid/dataset/dataset_schema_handler_one_table.py
```diff
@@ -104,57 +104,9 @@ def _check_one_table_data_consistency(self):
                     msg = f"dimension column {column} must have data type = StringType"
                     raise DSGInvalidDataset(msg)
                 dimension_types.add(dim_type)
+        check_for_nulls(self._load_data, exclude_columns={VALUE_COLUMN})
```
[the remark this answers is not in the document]
daniel-thom: Correct, I think this is an artifact of the original code. The order of operations is now
write to registry
- split nulls out to separate file
run checks

So there should never be nulls here.

dsgrid/registry/dataset_registry_manager.py
```diff
@@ -332,13 +337,27 @@ def _check_time_consistency(
...
             lk_path = path / filename
             if lk_path.exists():
-                return read_dataframe(lk_path)
-
-        msg = f"Did not find any lookup data files in {path}. Expected one of {ALLOWED_LOAD_DATA_LOOKUP_FILENAMES}"
+                df = read_dataframe(lk_path)
+                if "id" not in df.columns:
+                    msg = "load_data_lookup does not include an 'id' column"
+                    raise DSGInvalidDataset(msg)
+                missing = df.filter("id IS NULL").drop("id")
```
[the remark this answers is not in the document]
daniel-thom: We do later.

dsgrid/dataset/dataset_schema_handler_one_table.py
```diff
@@ -248,6 +168,7 @@ def make_mapped_dataframe(self, context: QueryContext) -> DataFrame:
         plan = query.mapping_plan
         if plan is None:
             plan = self.build_default_dataset_mapping_plan()
+        breakpoint()
```
lixiliu, line 171: Stale

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -251,15 +184,47 @@ def _check_dimension_associations(
...
+        )
+        if missing_associations is not None:
+            missing_associations = self._union_not_covered_dimensions(
+                missing_associations, context
+            )
+            assert sorted(df.columns) == sorted(missing_associations.columns)
+            df = union([df, missing_associations.select(*df.columns)])
+        mapping_plan = self.build_default_dataset_mapping_plan()
+        with DatasetMappingManager(self.dataset_id, mapping_plan, context) as mapping_manager:
+            df = self._remap_dimension_columns(df, mapping_manager).drop("fraction")
```
lixiliu, line 205: Need to call distinct() I think because of potentially many to one mapping.

lixiliu (approved):
Minor cleanup and comment.

dsgrid/dataset/dataset_schema_handler_base.py
lixiliu: NVM, I think _remap_dimension_columns() reduces it automatically.

PR #380, opened 2025-10-08
Review and edits of dsgrid metadata

elainethale:
## Description

This PR:
- Reduces the number of required meta-data fields
- Streamlines dataset meta-data
- Addresses and/or removes a few old to-dos and blocks of commented-out code

For this WIP stage, the PR also includes some comments for discussion.

## Checklist

- [x] All tests pass
- [x] Consider transferring TODOs to GitHub

### DIFF
--- a/dsgrid/config/dataset_config.py
+++ b/dsgrid/config/dataset_config.py
@@ -13,7 +13,6 @@ from dsgrid.config.dimension_config import (
 )
 from dsgrid.config.time_dimension_base_config import TimeDimensionBaseConfig
 from dsgrid.dataset.models import (
-    TableFormatModelBase,
     PivotedTableFormatModel,
     UnpivotedTableFormatModel,
     TableFormatModel,
@@ -357,7 +356,7 @@ class DatasetConfigModel(DSGBaseDatabaseModel):
     )
     # This field must be listed before dimensions.
     use_project_geography_time_zone: bool = Field(
-        default=False,  # ETH@20251008 - Default here conflicts with make_unvalidated_dataset_config
+        default=False,
         description="If true, time zones will be applied from the project's geography dimension. "
         "If false, the dataset's geography dimension records must provide a time zone column.",
     )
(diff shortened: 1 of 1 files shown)


dsgrid/config/dataset_config.py
```diff
@@ -252,76 +258,106 @@ class DatasetConfigModel(DSGBaseDatabaseModel):
...
-    )
     enable_unit_conversion: bool = Field(
         default=True,
         description="If the dataset uses its dimension mapping for the metric dimension to also "
         "perform unit conversion, then this value should be false.",
     )
     # This field must be listed before dimensions.
     use_project_geography_time_zone: bool = Field(
-        default=False,
+        default=False,  # ETH@20251008 - Default here conflicts with make_unvalidated_dataset_config
```
daniel-thom: Do we need to change something?

dsgrid/config/dataset_config.py
```diff
@@ -444,17 +494,18 @@ def check_time_zone(self) -> "DatasetConfigModel":
 def make_unvalidated_dataset_config(
     dataset_id,
     metric_type: str,
-    table_format: dict[str, str] | None = None,
-    data_classification=DataClassificationType.MODERATE.value,
-    dataset_type=InputDatasetType.MODELED,
+    table_format: TableFormatModelBase = UnpivotedTableFormatModel(),
```
daniel-thom: This violates the rule of assigning a mutable value to a default parameter. This is ok if we don't change it. But maybe someone will change it in the future. This is why I usually default to None in this case.

daniel-thom: I'll defer to you.

dsgrid/config/dataset_config.py
elainethale: They don't actually conflict anymore. Removing the comment.

daniel-thom (approved):

PR #389, opened 2025-12-03
Refactor CLI

daniel-thom:
1. Move the `dsgrid-admin` registry commands to the main `dsgrid registry` CLI. We don't have the notion of an official registry and so the "admin" concepts were confusing.
2. Remove/disable the `--offline` registry options. We don't have a remote registry are now always offline. This was always confusing to users.

### DIFF
--- a/dsgrid/cli/dsgrid.py
+++ b/dsgrid/cli/dsgrid.py
@@ -42,16 +42,6 @@ logger = logging.getLogger(__name__)
 @click.option(
     "-n", "--no-prompts", default=False, is_flag=True, show_default=True, help="Do not prompt."
 )
-# Offline mode is permanently on because we do not currently support a remote, official registry.
-# @click.option(
-#     "--offline/--online",
-#     is_flag=True,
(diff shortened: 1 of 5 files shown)


dsgrid/cli/dsgrid.py
```diff
@@ -116,6 +121,7 @@ def cli(
         timer_stats_collector.enable()
     else:
         timer_stats_collector.disable()
+    ctx.params["offline"] = True
```
elainethale: Why keep this around?

elainethale (approved):
LGTM. My comments are only questions about whether we want to delete even more lines of code.

dsgrid/cli/dsgrid.py
daniel-thom: There is code elsewhere that depends on it. I tried to only comment-out code that we might eventually use again. It is conceivable that we have local and remote registries that can be synced. If you say, "no way", I can fully delete.

PR #392, opened 2025-12-05
Initial refactor of dsgrid documentation

elainethale:
## Description

This implements a new structure for the dsgrid documentation. There is still a lot of content creation, review, and reorganization to do at the subsection and page level, but this seems like a good point to double-check with everyone that we like this direction. Then we can divide and conquer writing, reviewing, and making other improvements by signing up for the various subsections and/or topics.

@daniel-thom @ashreeta @lixiliu 

Instructions for generating docs:

```
pip install -e .[doc,spark]
cd docs
python -m sphinx -b html source _build/html
```

Then you can open `docs/_build/html/index.html` in your browser.

## Checklist

- [ ] All tests pass


elainethale:
@daniel-thom, please check out the new tables for data models and enums ...

docs/doc_generators/generate_dimension_classes.py
```diff
@@ -0,0 +1,177 @@
...
+    lines.append("| Field | Type | Default | Description |")
+    lines.append("|-------|------|---------|-------------|")
+    lines.append("| `id` | `str` | *(required)* | Unique identifier within a dimension |")
+    lines.append("| `name` | `str` | *(required)* | User-defined name |")
+    for field_name, field_info in _get_own_fields(cls).items():
+        type_str = _format_type(field_info.annotation)
+        default_str = _format_default(field_info)
+        desc = (field_info.description or "").replace("\n", " ").strip()
+        desc = " ".join(desc.split()).replace("|", "\\|")
+        lines.append(f"| `{field_name}` | {type_str} | {default_str} | {desc} |")
```
daniel-thom, line 99: Is it worth asking the AI to use a templating system (jinja2, mustache)? I'm fine as is, but I have seen this behavior, asked it about templating, to which it responded, "that's a great idea!"

daniel-thom (approved):

docs/doc_generators/generate_dimension_classes.py
elainethale: I don't want to mess with that right now. Not a bad idea, just not high enough priority.

Also, AI is often highly complementary ;)

PR #393, opened 2025-12-08
Adds WeatherVariable metric type and DAILY time type

elainethale:
## Description

Additional dimension types support daily BAIT data needed for STRIDE.

## Checklist

- [ ] All tests pass

### DIFF
--- a/tests/test_daily_time.py
+++ b/tests/test_daily_time.py
@@ -114,6 +114,48 @@ def daily_time_dimension_drop_feb29():
     )
 
 
+@pytest.fixture
+def daily_time_dimension_drop_dec31():
+    """Create a daily time dimension config that drops Dec 31."""
+    yield DailyTimeDimensionConfig(
+        DailyTimeDimensionModel(
+            dimension_type=DimensionType.TIME,
+            class_name="DailyTime",
+            module="dsgrid.dimension.standard",
+            name="daily_time_no_dec31",
+            description="test daily time without Dec 31",
+            ranges=[
+                DailyRangeModel(start="2012-12-28", end="2013-01-03"),
+            ],
+            str_format="%Y-%m-%d",
+            time_interval_type=TimeIntervalType.PERIOD_BEGINNING,
+            leap_day_adjustment=LeapDayAdjustmentType.DROP_DEC31,
+            year_column="year",
+        )
+    )
+
+
+@pytest.fixture
+def daily_time_dimension_drop_jan1():
+    """Create a daily time dimension config that drops Jan 1."""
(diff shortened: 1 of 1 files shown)


dsgrid/config/daily_time_dimension_config.py
```diff
@@ -0,0 +1,200 @@
...
+            msg = (
+                f"DailyTimeDimensionConfig expects columns {expected_cols}, "
+                f"but got {actual_cols}"
+            )
+            raise ValueError(msg)
+
+        # Get the year column name (time_year, weather_year, or model_year)
+        year_col = self.get_year_column()
+        month_col = "time_month"
+        day_col = "time_day"
```
elainethale: @lixiliu, does representative time allow user discretion for all column names, or does it require certain column names?

dsgrid/config/dimensions.py
```diff
@@ -20,6 +20,7 @@
     TimeDimensionType,
     RepresentativePeriodFormat,
     DatetimeFormat,
+    LeapDayAdjustmentType,
```
elainethale: Why wasn't this needed before? Maybe the Daily time models are in the wrong place?

dsgrid/dimension/time.py
```diff
@@ -20,6 +20,7 @@ class TimeDimensionType(DSGEnum):
     REPRESENTATIVE_PERIOD = "representative_period"
     INDEX = "index"
     NOOP = "noop"
+    DAILY = "daily"
```
elainethale: Would it be okay to move this up higher in the list?

dsgrid/config/daily_time_dimension_config.py
elainethale: I see that it requires specific column names: (`month`, `day_of_week`, `hour`) OR (`month`, `hour`, `is_weekday`). Per that patter, I'll change my column names to `month`, `day`.

dsgrid/config/daily_time_dimension_config.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def build_time_dataframe(self) -> DataFrame:
+        model_time = self.list_expected_dataset_timestamps()
+        df_time = get_spark_session().createDataFrame(
+            model_time,
+            schema=StructType(
+                [StructField(col, IntegerType(), False) for col in ["year", "month", "day"]]
+            ),
+        )
+        df_time = df_time.withColumnRenamed("year", self.get_year_column())
```
elainethale: @lixiliu, if daily time is using the `weather_year` or `model_year` column at first, should this method continue using those names, or should it just produce a `year` column?

PR #400, opened 2025-12-16
IANA time zones

daniel-thom:
Replace custom dsgrid time zone names with IANA names. I also disabled the full Spark tests so that we can quickly get through several PRs. Will re-enable at the end.

### DIFF
--- a/dsgrid/utils/dataset.py
+++ b/dsgrid/utils/dataset.py
@@ -455,7 +455,7 @@ def convert_time_zone_with_chronify_spark_hive(
     df: DataFrame,
     value_column: str,
     from_time_dim: TimeDimensionBaseConfig,
-    time_zone: str | str,
+    time_zone: str,
     scratch_dir_context: ScratchDirContext,
 ) -> DataFrame:
     """Create a single time zone-converted table with chronify and Spark and a Hive Metastore."""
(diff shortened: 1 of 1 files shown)


dsgrid/dimension/base_models.py
```diff
@@ -207,12 +208,17 @@ def check_timezone_in_geography(dimension, err_msg=None):
...
+            ZoneInfo(tz)
+        except ZoneInfoNotFoundError:
+            invalid_tzs.append(tz)
+
+    if invalid_tzs:
         msg = (
-            f"Geography dimension {dimension.dimension_id} has invalid time_zone(s) in records: "
-            f"{dimension.filename}: {diff}. Use dsgrid TimeZone enum values only ({tz})."
+            f"Geography dimension {dimension.dimension_id} has invalid time zone(s) in records "
+            f"{dimension.filename}: {invalid_tzs}. Use IANA time zone names only."
```
elainethale: Is there an official, human readable website we can send users to?

dsgrid/query/models.py
```diff
@@ -528,7 +527,7 @@ class QueryResultParamsModel(CacheableQueryBaseModel):
         default=[],
     )
     # TODO #205: implement
```
elainethale: TODO can be removed now.

dsgrid/dimension/time.py
```diff
@@ -140,187 +140,6 @@ class MeasurementType(DSGEnum):
     )
 
 
-class TimeZone(DSGEnum):
```
lixiliu: Turn this into a lookup for users in the doc?

PR #401, opened 2025-12-16
Support DateTime with time parts specified in columns

daniel-thom:


dsgrid/registry/dataset_registry_manager.py
```diff
@@ -313,47 +331,184 @@ def _register(
...
+
+        if not isinstance(time_dim.model, DateTimeDimensionModel) or not isinstance(
+            time_dim.model.column_format, TimeFormatInPartsModel
+        ):
+            return
+
+        # This code only exists because we lack full support for time zone naive timestamps.
+        df = handler.get_base_load_data_table()
+        col_format = time_dim.model.column_format
+        hour_col = col_format.hour_column
```
elainethale: I'm not sure why this is defined here.

dsgrid/registry/dataset_registry_manager.py
```diff
@@ -313,47 +331,183 @@ def _register(
...
+    def _build_timestamp_sql(
+        timestamp_str_expr: str, fixed_tz: str | None
+    ) -> tuple[str, TimeFormatDateTimeTZModel | TimeFormatDateTimeNTZModel]:
+        """Build the final timestamp SQL expression and determine the column format."""
+        if fixed_tz is not None:
+            if use_duckdb():
+                sql = f"cast({timestamp_str_expr} || ' {fixed_tz}' as timestamptz) as timestamp"
+            else:
+                tz = ZoneInfo(fixed_tz)
+                offset = datetime(2012, 1, 1, tzinfo=tz).strftime("%z")
```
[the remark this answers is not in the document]
elainethale: I think I see--the actual year doesn't matter, right? I guess this is something weird for making sure you get the right timestamp in Spark?

elainethale (commented):
LGTM. A couple of minor comments.

dsgrid/registry/dataset_registry_manager.py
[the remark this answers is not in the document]
daniel-thom: yes

dsgrid/config/dimensions.py
```diff
@@ -237,6 +237,87 @@ def serialize_cls(self, val: str, _) -> None:
...
+        "for all rows.",
+        default=None,
+    )
+    time_zone: str | None = Field(
+        default=None,
+        title="time_zone",
+        description="Time zone of the timestamps. Use None for time zone-naive timestamps.",
+    )
+
+    def converted_to_datetime(self):
```
lixiliu: Can remove these methods as they are not being used.

lixiliu (approved):

dsgrid/config/dimensions.py
```diff
@@ -237,6 +237,87 @@ def serialize_cls(self, val: str, _) -> None:
...
+        title="month_column",
+        description="Name of the month column in the dataset.",
+    )
+    day_column: str = Field(
+        title="day_column",
+        description="Name of the day column in the dataset.",
+    )
+    hour_column: str | None = Field(
+        title="hour_column",
+        description="Name of the hour column in the dataset. If None, the hour will be set to 0 "
```
lixiliu: Value should represent hour of day (0-23), not hour of year

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -135,6 +139,10 @@ def check_consistency(self, missing_dimension_associations: dict[str, DataFrame]
     def check_time_consistency(self):
         """Check the time consistency of the dataset."""
 
+    @abc.abstractmethod
+    def get_base_load_data_table(self) -> DataFrame:
```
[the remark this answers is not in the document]
daniel-thom: The DatasetRegistryManager needs a public interface method on the handler to retrieve the user's load data table. This is before we have joined the lookup table. The purpose is to convert the time format in that table from time-in-parts to datetime.

dsgrid/config/dimensions.py
```diff
@@ -237,6 +237,87 @@ def serialize_cls(self, val: str, _) -> None:
...
+    )
+    month_column: str = Field(
+        title="month_column",
+        description="Name of the month column in the dataset.",
+    )
+    day_column: str = Field(
+        title="day_column",
+        description="Name of the day column in the dataset.",
+    )
+    hour_column: str | None = Field(
```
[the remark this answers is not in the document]
daniel-thom: Given limited budget for development and test, I think we should exclude this for now.

dsgrid/config/dimensions.py
```diff
@@ -238,6 +237,78 @@ def serialize_cls(self, val: str, _) -> None:
...
+    def get_time_columns(self) -> list[str]:
+        return [self.time_column]
+
+
+class TimeFormatDateTimeNTZModel(DSGBaseModel):
+    """Format of timestamps in a dataset is timezone-naive datetime,
+    requiring localization to time zones."""
+
+    dtype: Literal["TIMESTAMP_NTZ"] = "TIMESTAMP_NTZ"
+    time_column: str = Field(
```
elainethale: nit: inconsistent spacing between dtype and time_column

dsgrid/config/dimensions.py
```diff
@@ -238,6 +237,78 @@ def serialize_cls(self, val: str, _) -> None:
         return None
 
 
+class TimeFormatDateTimeTZModel(DSGBaseModel):
+    """Format of timestamps in a dataset is timezone-aware datetime."""
+
+    dtype: Literal["TIMESTAMP_TZ"] = "TIMESTAMP_TZ"
```
elainethale: I find this syntax odd-looking, but I assume it's right/works.

dsgrid/config/index_time_dimension_config.py
```diff
@@ -49,7 +49,9 @@ def to_chronify(
 
     def get_frequency(self) -> timedelta:
         freqs = [trange.frequency for trange in self.model.ranges]
-        assert set(freqs) == {freqs[0]}, freqs
+        if len(set(freqs)) > 1:
```
elainethale: In another place in this PR a similar check also looked for (not set(freqs)) (len 0).

dsgrid/registry/dataset_registry_manager.py
```diff
@@ -313,56 +331,193 @@ def _register(
...
+        timestamp_sql, new_col_format = self._build_timestamp_sql(timestamp_str_expr, fixed_tz)
+
+        cols_to_drop = self._get_time_columns_to_drop(col_format)
+        reformatted_df = self._apply_timestamp_transformation(df, cols_to_drop, timestamp_sql)
+
+        self._update_config_for_timestamp(
+            config, reformatted_df, scratch_dir_context, cols_to_drop, new_col_format
         )
-        schema_handler.check_time_consistency()
+        self._update_time_dimension(time_dim, new_col_format, col_format.hour_column)
```
elainethale: The transformations here are important. Maybe add an INFO message that lists the before and after columns?

dsgrid/registry/dataset_registry_manager.py
```diff
@@ -313,56 +331,193 @@ def _register(
...
+
+    @staticmethod
+    def _build_timestamp_string_expr(col_format: TimeFormatInPartsModel) -> str:
+        """Build SQL expression that creates a timestamp string from time-in-parts columns."""
+        str_type = "varchar" if use_duckdb() else "string"
```
elainethale: An example of "if duckdb" logic scattered about. Should there be/is there already a module of helper functions? This one would be something like:

```
def get_str_type():
    return "varchar" if use_duckdb() else "string"
```

dsgrid/registry/dataset_registry_manager.py
```diff
@@ -313,56 +331,193 @@ def _register(
...
+            f"cast({col_format.year_column} as {str_type}) || '-' || "
+            f"lpad(cast({col_format.month_column} as {str_type}), 2, '0') || '-' || "
+            f"lpad(cast({col_format.day_column} as {str_type}), 2, '0') || ' ' || "
+            f"{hour_expr} || ':00:00'"
+        )
+
+    @staticmethod
+    def _resolve_timezone(
+        df: DataFrame, config: DatasetConfig, col_format: TimeFormatInPartsModel
+    ) -> tuple[DataFrame, str | None]:
```
elainethale: Will all these methods that depend on TimeFormatInPartsModel be removed once Lixi's changes are merged in?

dsgrid/config/dimensions.py
daniel-thom: Agree that it's odd. This is the discriminator field for Pydantic when de-serializing from JSON. From a Python perspective, it is a string field that must contain this value - and that value is provided by default.

dsgrid/registry/dataset_registry_manager.py
daniel-thom: We need to either have all this replaced by that work or make time-in-parts a first-class entity in chronify. I'm hoping for the former, but we will see.

PR #403, opened 2026-01-13
Make dataset checks optional

daniel-thom:
Not all consumers of dsgrid want the rigorous dataset dimension checks that are enabled by default. This PR allows them to bypass those checks. The bypass mechanism does not apply to the dataset-submission-to-project process, but it could in the future.

### DIFF
--- a/dsgrid/dimension/base_models.py
+++ b/dsgrid/dimension/base_models.py
@@ -165,6 +165,8 @@ def check_required_dataset_dimensions(dimensions, requirements: DatasetDimension
     ----------
     dimensions : list
         list of DimensionReferenceModel
+    requirements : DatasetDimensionRequirements
+        Defines the dimension requirements in the dataset that must be checked.
     tag : str
         User-defined string to include in exception messages
 


dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -130,6 +130,7 @@ def check_consistency(
         self,
         missing_dimension_associations: dict[str, DataFrame],
         scratch_dir_context: ScratchDirContext,
+        requirements: DatasetDimensionRequirements,
```
elainethale: Should there be a default here and throughout?

daniel-thom:
> Looks like CoPilot might have caught some minor issues.
> 
> I see that in some (more public-facing) functions there are default requirements. My only minor comment is to double-check where defaults are provided versus where they are not.

I fixed one docstring suggested by Copilot but am choosing to ignore the other comments as being not relevant. The `check_consistency` method is not user-facing, so I don't think we need a default value (the same method requires ScratchDirContext, which is even harsher for the user to provide. The relevant public methods are `DatasetRegistryManager.register` and `bulk_register`.

PR #406, opened 2026-01-23
Ll/local time part2

lixiliu:
Closes GitHub Issue #

## Description

Local time support.

Allows config to describe tz-naive timestamps that are:
- `aligned_in_absolute_time`(with or without time zone) or
- `aligned_in_local_std_time` (where time zones are given by a TIME_ZONE_COLUMN in the data table)

Time zone localization is triggered during dataset registration when:
- timestamps in data table are tz-naive (TIMESTAMP_NTZ) but time config has a time zone.
- For `aligned_in_local_std_time`, localization is based on the TIME_ZONE_COLUMN from the geography dimension.

No time zone localization available when dataset is submitted to project

Chronify: https://github.com/NatLabRockies/chronify/pull/61

## Checklist

- [x] All tests pass


dsgrid/config/date_time_dimension_config.py
```diff
@@ -30,36 +31,32 @@ def to_chronify(self) -> chronify.DatetimeRange:
...
+                    dtype=col_dtype,
+                    time_column=time_cols[0],
+                    start=start,
+                    length=self.get_lengths()[0],
+                    resolution=self.get_frequency(),
+                    time_zone_column=TIME_ZONE_COLUMN,
+                    time_zones=self._get_chronify_time_zones(),
+                    measurement_type=self._model.measurement_type,
+                    interval_type=self._model.time_interval_type,
+                )
```
daniel-thom, line 60: What if the value is LOCAL_AS_STRINGS? Need a catch-all case.

dsgrid/config/date_time_dimension_config.py
```diff
@@ -117,20 +109,61 @@ def get_tzinfo(self) -> tzinfo | None:
...
+        match self.model.column_format.dtype:
+            case "TIMESTAMP_NTZ":
+                return chronify.TimeDataType.TIMESTAMP_NTZ
+            case "TIMESTAMP_TZ":
+                return chronify.TimeDataType.TIMESTAMP_TZ
+            case _:
+                msg = f"Unsupported datetime dtype for chronify: {self.model.column_format.dtype}"
+                raise ValueError(msg)
+
+    def _get_chronify_start_time(self) -> pd.Timestamp:
```
daniel-thom: I'm confused as to why we are switching between `datetime` and `pd.Timestamp`. `get_start_times` is annotated as returning `list[pd.Timestamp]` but it actually returns datetimes. Then we convert to pd.Timestamp here. Can we only use one timestamp type?

dsgrid/utils/dataset.py
```diff
@@ -828,3 +984,98 @@ def split_expected_missing_rows(
     drop_columns = time_columns + [VALUE_COLUMN]
     missing_associations = null_df.drop(*drop_columns)
     return df.filter(f"{VALUE_COLUMN} IS NOT NULL"), missing_associations
+
+
+def localize_timestamps_if_necessary(
```
daniel-thom, line 1152: Is this redundant with the method in dataset_schema_handler_base.py line 456?

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -451,6 +453,98 @@ def _check_dataset_time_consistency_by_time_array(self, time_cols, load_data_df)
...
+                    # use Spark with CSV data files.
+                    src_path = context.get_temp_filename(suffix=".parquet")
+                    write_dataframe(load_data_df, src_path)
+                else:
+                    src_path = data_file_path
+                store_file = context.get_temp_filename(suffix=".db")
+                with create_store(store_file) as store:
+                    # This performs all of the checks.
+                    store.create_view_from_parquet(src_path, chronify_schema)
+                    if localization_plan == "localize_to_single_tz":
```
daniel-thom: Seems like we need validation of localization_plan at the beginning of the method. There will undefined errors on `new_chronify_schema` if someone adds a new option without handling.

dsgrid/utils/dataset.py
```diff
@@ -828,3 +984,98 @@ def split_expected_missing_rows(
...
+
+    # This is a workaround for pivoted tables to still use Chronify, which only supports stacked tables.
+    value_columns = config.get_value_columns()
+    assert len(value_columns) > 0, value_columns
+    value_column = next(iter(value_columns))
+
+    runtime_config = dsgrid.runtime_config
+    match localization_plan:
+        case "localize_to_single_tz":
+            to_time_zone = time_dim._get_chronify_time_zone()
```
daniel-thom: This shouldn't rely on a private method. Why do we need a chronify-specific method? Can't it be `time_dim.get_time_zone`?

tests/test_localize_timestamps_if_necessary.py
```diff
@@ -0,0 +1,465 @@
...
+
+    sdf = spark.createDataFrame(df)
+    res_df, changed = localize_timestamps_if_necessary(
+        sdf, config, scratch_dir_context=MagicMock()
+    )
+    assert changed is True
+    if isinstance(res_df, DataFrame):
+        res_df = res_df.toPandas()
+    assert sorted(res_df[TIME_COLUMN]) == sorted(called_df[TIME_COLUMN])
+    # target.assert_called_once()
```
daniel-thom: Why are many of these lines commented out?

daniel-thom (changes requested):

dsgrid/dataset/dataset_schema_handler_base.py
lixiliu: N/A since it is removed from the handler

dsgrid/utils/dataset.py
lixiliu: Because we need the ZoneInfo obj version of time_dim.get_time_zone.

I don't want to make it a public method, though I can, because the method is only for the datetime time config.

tests/test_localize_timestamps_if_necessary.py
lixiliu: They work in local purest but don't seem to work for the CI version. I'll remove

dsgrid/config/dimensions.py
```diff
@@ -253,13 +252,13 @@ def get_time_columns(self) -> list[str]:
 
 class TimeFormatDateTimeNTZModel(DSGBaseModel):
     """Format of timestamps in a dataset is timezone-naive datetime,
-    requiring localization to time zones."""
+    timestamps can be localized to time zone(s)."""
```
elainethale: Perhaps delete, "timestamps can be localized to time zone(s)". This model is just documenting what the time format of the dataset is, which is covered by the first line.

dsgrid/config/dimensions.py
```diff
@@ -290,14 +289,21 @@ class TimeFormatInPartsModel(DSGBaseModel):
...
     )
-    time_zone: str | None = Field(
+    offset_column: str | None = Field(
+        title="offset_column",
+        description="Name of the offset column in the dataset. Value is the UTC offset in hours (e.g., -8 or -08:00). "
+        "If None, the offset will not be set.",
         default=None,
-        title="time_zone",
-        description="IANA time zone of the timestamps. Use None for time zone-naive timestamps.",
     )
```
elainethale, line 322: Does this only support standard time, or can offset change over the year?

It seems like it would be best to support time_zone strings as well ... Can we provide guidance/support for data that includes daylight savings? I.e., be robust to repeated hours in fall and skipped hour in spring?

Timezone offset is very confusing. Can we provide an example that clarifies our sign convention whenever it is mentioned?

dsgrid/config/dimensions.py
elainethale:
```suggestion
    """All geographies have data with the same set of timestamps in absolute time.
```

elainethale: This is a great example. We should have examples like this everywhere, especially for enum fields and data models / data model fields.

elainethale: This doesn't seem to be true any longer.

dsgrid/config/dimensions.py
```diff
@@ -457,38 +463,43 @@ def handle_legacy_fields(cls, values):
...
 
+    def get_time_zones(self) -> list[str]:
+        return [self.time_zone] if self.time_zone else []
+
 
 class LocalTimeMultipleTimeZones(DSGBaseModel):
-    """For each geography, data has the same set of timestamps when interpreted as local clock time by adjusting
-    for the time zone of each geography.
-    Timestamps in the data must be tz-aware.
+    """For each geography, data has the same set of timestamps when interpreted in local standard time.
```
elainethale:
```suggestion
    """Across geographies, timestamps cover the same interval of local time, e.g., all of 2018 as experienced locally, in clock time. This means, for example, that all data would have the same set of timestamps if timestamps are first interpreted in geography-specific local standard time and are then made timezone unaware.
```

dsgrid/dimension/time.py
```diff
@@ -27,9 +27,21 @@ class TimeZoneFormat(str, Enum):
     """Defines the time format of the datetime config model"""
 
     ALIGNED_IN_ABSOLUTE_TIME = "aligned_in_absolute_time"
-    ALIGNED_IN_CLOCK_TIME = "aligned_in_clock_time"
+    ALIGNED_IN_LOCAL_STD_TIME = "aligned_in_local_std_time"
```
elainethale: I'm not sure about the name `ALIGNED_IN_LOCAL_STD_TIME`. By its nature, this format is in one way "aligned" and in another way "not aligned". Alternatives:

- `ALIGNED_IN_NTZ_LOCAL_STD_TIME` - Clarifies that you have to drop the time zone in order for the timestamps to appear aligned.
- `ALIGNED_IN_STD_CLOCK_TIME` - I realize you dropped "clock time" in this PR, but it is resonant with intuition, i.e., what everyone sees when they look up at the clock. Adding "STD" clarifies how this format deals with daylight savings.

The last option also brings up my earlier comment--should we be more robust to data that directly contains daylight savings ugliness?

dsgrid/common.py
```diff
@@ -27,6 +27,7 @@ def on_hpc():
 SYNC_EXCLUDE_LIST = ["*.DS_Store", "**/*.lock"]
 TIME_ZONE_COLUMN = "time_zone"
 VALUE_COLUMN = "value"
+TIME_COLUMN = "timestamp"
```
elainethale, line 33: Are we standardizing to this now?

If so, can data submitters map other column names to this one if necessary? The same question applies to these other columns: "value" and "time_zone".

dsgrid/utils/dataset.py
```diff
@@ -359,19 +363,20 @@ def convert_time_zone_with_chronify_duckdb(
...
     value_column: str,
     from_time_dim: TimeDimensionBaseConfig,
-    time_zone: str,
+    time_zone: tzinfo | None,
     scratch_dir_context: ScratchDirContext,
 ) -> DataFrame:
     """Create a single time zone-converted table with chronify and DuckDB.
     All operations are performed in memory.
+    Time zone conversion converts from tz-aware timestamps to
+    tz-naive timestamps with the specified time zone as a new column.
```
elainethale, line 376: Time zone "conversion" is a vague term. Is it always tz-aware to tz-naive + time_zone column throughout the dsgrid and chronify codebases, or is it overloaded?

dsgrid/utils/dataset.py
```diff
@@ -499,6 +555,61 @@ def convert_time_zone_by_column_with_chronify_spark_hive(
...
 
 
+def localize_time_zone_with_chronify_spark_hive(
+    df: DataFrame,
+    value_column: str,
+    from_time_dim: TimeDimensionBaseConfig,
+    time_zone: tzinfo | None,
+    scratch_dir_context: ScratchDirContext,
+) -> DataFrame:
+    """Create a single time zone-localized table with chronify and Spark and a Hive Metastore."""
```
elainethale: Check that the doc strings for the spark methods repeat relevant information from the duckdb methods.

dsgrid/utils/dataset.py
```diff
@@ -828,3 +984,98 @@ def split_expected_missing_rows(
...
     missing_associations = null_df.drop(*drop_columns)
     return df.filter(f"{VALUE_COLUMN} IS NOT NULL"), missing_associations
+
+
+def localize_timestamps_if_necessary(
+    df: DataFrame,
+    config: DatasetConfig,
+    scratch_dir_context: ScratchDirContext,
+) -> tuple[DataFrame, bool]:
+    """Localize tz-naive timestamps to time zone(s) in the dataframe if necessary using Chronify."""
```
elainethale: Under what conditions is localization "necessary"?

dsgrid/config/date_time_dimension_config.py
```diff
@@ -30,36 +32,35 @@ def to_chronify(self) -> chronify.DatetimeRange:
...
+                return chronify.DatetimeRange(
+                    dtype=col_dtype,
+                    time_column=time_cols[0],
+                    start=start,
+                    length=self.get_lengths()[0],
+                    resolution=self.get_frequency(),
+                    measurement_type=self._model.measurement_type,
+                    interval_type=self._model.time_interval_type,
+                )
+            case TimeZoneFormat.ALIGNED_IN_LOCAL_STD_TIME:
```
elainethale: Should we also support local time data with local timestamps like America/Denver? _TZ type, "aligned in standard time", but timestamps are localized to clock time, including DST.

elainethale (commented):
Looking at this code brings two main questions to mind:

1. Is it possible to know what operations will be/are applied to bring one time convention in line with another?

2. Are we supporting all time dimension types (when reasonably compatible) as possible project time base dimensions and/or supplemental dimensions (are they queryable)?

dsgrid/dimension/time.py
elainethale: Removed by this PR and needs to be deleted or marked for deprecation, right?

dsgrid/dimension/time.py
```diff
@@ -27,9 +27,21 @@ class TimeZoneFormat(str, Enum):
...
     LOCAL_AS_STRINGS = "local_as_strings"
```
elainethale: Is this used? Claude thinks this might cause a silent failure if a user selected it.

dsgrid/dimension/time.py
elainethale, line 63: This Adjustment class and the next two seem one-directional and/or ambiguous to me. Should we have separate "adjustment types" for with leap day -> without leap day and without leap day - with leap day, and likewise for multiple types of DST conversions (standard timestamps <-> timestamps with DST, representative or indexed timestamps with different DST handling <-> standard timestamps or timestamps with DST)?

elainethale (commented):
Here are a couple of additional points raised by Claude that make sense to me and we might want to address:

```
If someone set an ANNUAL or REPRESENTATIVE_PERIOD [project] base dimension, there's no guard — it would fail only at dataset registration or query time with a confusing error. Consider adding a project-level validator or at least documenting that DATETIME is the expected base type.
```

```
AlignedTimeSingleTimeZone.time_zone is now str | None. When None, get_time_zones() returns []. This flows through to get_localization_plan() returning None (no localization). This enables true tz-naive datetime data, which is useful. But it also means ALIGNED_IN_ABSOLUTE_TIME with time_zone=None and TIMESTAMP_NTZ produces data with no timezone information anywhere — which may be confusing for downstream consumers. Worth documenting what this combination means.
```

dsgrid/dimension/time.py
elainethale: This is related to my comment about whether we want to explicitly support standard timezone types like `America/Denver`. Claude raised a similar question.

dsgrid/config/date_time_dimension_config.py
```diff
@@ -117,20 +113,61 @@ def get_tzinfo(self) -> tzinfo | None:
...
-                    f"Unsupported combination of format_type {self.model.time_zone_format.format_type} "
-                    f"and localize_to_time_zone {self.model.localize_to_time_zone}"
+                    f"Unsupported combination of time zone format: {self.model.time_zone_format.format_type}, and "
+                    f"time zone(s): {self.model.time_zone_format.get_time_zones()}"
                 )
                 raise ValueError(msg)
+
+    def _get_chronify_dtype(self) -> chronify.TimeDataType:
+        match self.model.column_format.dtype:
+            case "TIMESTAMP_NTZ":
```
daniel-thom: Having this string be all-caps is unfortunate. All strings in our config files are lowercase. Can we still change this in chronify?

dsgrid/common.py
lixiliu: These are just enums for default values, users can still pass in their own column names.

dsgrid/config/date_time_dimension_config.py
lixiliu: This only describes alignement and is independent of timestamps being tz-aware. If tz-naive however, the timestamps must be localizable in standard time for accuracy, otherwise we would not know how to parse duplicated timestamps during fallback. This is enforced elesewhere. aligned_in_clock_time is technically more narrow and not correct when comparing a timeseries that observes only standard time like Hawii and another that observes daylight savings. The alignment occurs only when both are converted to local standard time.

dsgrid/config/date_time_dimension_config.py
lixiliu: I'll change it all to lower caps here.

dsgrid/config/dimensions.py
lixiliu: This lets the uses define the column in the dataset that contains the offset information. So the offset can vary by row. This is one way of ingesting the data, similar to timestamps as strings. This class supports time parsing that is broken out into parts, the parts can contain offset info to be converted to tz-aware timestamps.

dsgrid/dimension/time.py
lixiliu: A few places already raise warning when the selection is not one of the above.

dsgrid/dimension/time.py
lixiliu: Regarding data that observes daylight savings, we only accept it as tz-aware (options: tz-aware datetime config or DatetimeInParts config with a utc-offset column in the data). We don't accept tz-naive timestamps with skips and duplicates because we cannot accurately parse which duplicate comes first.

dsgrid/utils/dataset.py
lixiliu: It's the former, and I don't understand what overloaded mean here.

elainethale: Does "time zone conversion" always mean the same thing whenever it's used throughout the code base (not overloading the term), or does it mean one thing the the context of one function and another thing in another context (overloading the term)?

missing_associations/geography__subsector.csv
elainethale, line 1: This should not be here.

elainethale (changes requested):
Generating the documentation from a clean state shows errors related to chronify and the new how-to not being in a toc-tree.

elainethale:
> Generating the documentation from a clean state shows errors related to chronify and the new how-to not being in a toc-tree.

In addition to simply addressing the errors, please look for additional places (beyond the how-to index page) where the new how-to should be referenced.

dsgrid/config/dimensions.py
```diff
@@ -290,14 +291,21 @@ class TimeFormatInPartsModel(DSGBaseModel):
         "If None, the hour will be set to 0 for all rows.",
         default=None,
     )
-    time_zone: str | None = Field(
+    offset_column: str | None = Field(
+        title="offset_column",
+        description="Name of the offset column in the dataset. Value is the UTC offset in hours (e.g., -8 or -08:00). "
```
elainethale: I don't know how to interpret +/- here. Please provide a specific example. (What should this value be for Eastern or Pacific Standard Time?)

dsgrid/config/dimensions.py
```diff
@@ -457,38 +465,71 @@ def handle_legacy_fields(cls, values):
...
+    def get_time_zones(self) -> list[str]:
+        return [self.time_zone] if self.time_zone else []
+
 
 class LocalTimeMultipleTimeZones(DSGBaseModel):
-    """For each geography, data has the same set of timestamps when interpreted as local clock time by adjusting
-    for the time zone of each geography.
-    Timestamps in the data must be tz-aware.
+    """Across geographies, timestamps cover the same interval of local time,
+    e.g., all of 2018 as experienced locally, in clock time.
```
elainethale: In clock time, or in standard time?

dsgrid/config/dimensions.py
```diff
@@ -457,38 +465,71 @@ def handle_legacy_fields(cls, values):
...
+    This means, for example, that all data would have the same set of timestamps
+    if timestamps are first interpreted in geography-specific local standard time
+    and are then made timezone unaware.
+
+    data table must contain TIME_ZONE_COLUMN column with IANA time zones for each record.
```
elainethale: IANA time zones or UTC offsets?

elainethale: I see below that one needs to use IANA of the form Etc/GMT[+-]X. That should be clarified here.

tests/test_convert_time_format_if_necessary.py
```diff
@@ -0,0 +1,383 @@
...
+    expected = [
+        "2019-12-31 19:00:00",  # +05:00
+        "2020-01-01 08:00:00",  # -08:00
+        "2020-01-01 07:45:00",  # -07:45
+    ]
+    assert got == expected
+
+
+def test_offset_parsing_in_parts_backend_agnostic():
+    # Backend-agnostic: run under current engine (DuckDB or Spark).
```
elainethale: It seems like the other tests do this as well (they're not marked as spark-only or duckdb-only). Does this test add anything?

tests/test_convert_time_format_if_necessary.py
```diff
@@ -0,0 +1,383 @@
...
+
+    spark = get_spark_session()
+    set_current_time_zone("UTC")
```
elainethale: Seems like an odd place to set the time zone.

tests/test_convert_time_format_if_necessary.py
```diff
@@ -0,0 +1,383 @@
...
+        {
+            "year": [2020, 2020],
+            "month": [1, 1],
+            "day": [1, 1],
+            "hour": [0, 0],
+            "utc_offset": [24.5, -24.5],
+        }
+    )
+    df = spark.createDataFrame(pdf)
+    with pytest.raises(Exception):
```
elainethale, line 156: Since this would be a user error, should we be raising a more specific exception type? This comment also applies to the next several tests.

dsgrid/utils/dataset.py
elainethale: The larger issue is that the function name is also vague. `convert_time_zone_with_chronify_duckdb` could mean a variety of things ...

dsgrid/utils/dataset.py
```diff
@@ -446,20 +504,22 @@ def convert_time_zone_with_chronify_spark_hive(
...
     from_time_dim: TimeDimensionBaseConfig,
-    time_zone: str,
+    time_zone: tzinfo | None,
     scratch_dir_context: ScratchDirContext,
 ) -> DataFrame:
-    """Create a single time zone-converted table with chronify and Spark and a Hive Metastore."""
+    """Create a single time zone-converted table with chronify and Spark and a Hive Metastore.
+    Time zone conversion converts from tz-aware timestamps to
+    tz-naive timestamps with the specified time zone as a new column.
+    """
```
elainethale, line 517: Related to above comments, instead of `convert_time_zone` I think we need a term that conveys "fully formed timestamp to naive timestamp + time zone column". Maybe `convert_to_time_zone_column` or something similar? @daniel-thom, do you have ideas?

dsgrid/utils/dataset.py
```diff
@@ -479,6 +539,8 @@ def convert_time_zone_by_column_with_chronify_spark_hive(
 ) -> DataFrame:
     """Create a multiple time zone-converted table (based on a time_zone_column)
     using chronify and Spark and a Hive Metastore.
+    Time zone conversion converts from tz-aware timestamps to
+    tz-naive timestamps of the time zones specified in the time_zone_column.
```
elainethale:
```suggestion
    tz-naive timestamps with time zones specified in the time_zone_column.
```

dsgrid/utils/dataset.py
elainethale: Furthermore, now I see that this is the inverse of "localization" ... Maybe that helps with naming?

dsgrid/config/dimensions.py
daniel-thom: except that it uses "EST", which we don't support. :(

Obviously, it is what the average American understands. I suppose that I am commenting on the contradiction and not asking for a change.

tests/test_convert_time_format_if_necessary.py
daniel-thom: It doesn't get unset. We have a context manager for this.

dsgrid/registry/dataset_registry_manager.py
```diff
@@ -513,26 +610,53 @@ def _update_config_for_timestamp(
...
+        time_column = time_dim.model.column_format.time_column
+        time_dim.model.column_format = TimeFormatDateTimeTZModel(time_column=time_column)
+
+        # update data_file columns schema, if present, to reflect TIMESTAMP_TZ
+        data_file = config.model.data_layout.data_file
+        if data_file.columns is not None:
+            updated_columns: list[Column] = []
+            for col in data_file.columns:
+                if col.name == time_column:
+                    updated_columns.append(Column(name=time_column, data_type="TIMESTAMP_TZ"))
```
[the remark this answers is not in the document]
lixiliu: updated_columns takes DuckDB parameters, which uses uppercase for data_type.

dsgrid/utils/dataset.py
lixiliu: It is always used to mean coverting tz-aware timestamps from one time zone to another, so not overloaded. We have to output tz-naive timestamps because keeping the converted timestamps tz-aware does nothing because the timestamp display is superseded by the system time zone. Recall tz-aware timestamps gets recoverted to system time zone when converting from database to pandas.

dsgrid/dimension/time.py
lixiliu: For leap day adjustment, I agree that we don't currently support adding an extra day when mapping a non-leap year to a leap year.

For daylight savings, as long as the timestamps are tz-aware, all the complexity between time zone conversion is taken care of. So, dsgrid only need to support manipulating tz-naive timestamps to a format that can be localized to a time zone. The only case I can think of is the industrial time for IEF. And we have a pathway for that through index time.

dsgrid/utils/dataset.py
lixiliu: There are two types of functions here: `convert_time_zone` (which takes a single time zone) and `convert_time_zon_by_column` (which takes a time zone column instead). Both pandas and Ibis uses the term "convert" for the same intention. (Ibis uses `convert_timezone` and pandas uses `dt.tz_convert`)

docs/source/user_guide/dataset_registration/data_file_formats.md
elainethale, line 1: @lixiliu, please review all the Time content on this page for accuracy and clarity, starting from the Time heading.

docs/source/user_guide/dataset_registration/dimension_concepts.md
```diff
@@ -62,34 +61,128 @@ Records can also be listed directly in the configuration. For example:
...
+    month_column: "month",
+    day_column: "day",
+    hour_column: "hour",       // optional; omit to default all hours to 0
+    offset_column: "utc_offset",  // optional
+  }
+```
+
+Although the schema accepts multiple `ranges` entries, dsgrid currently only supports a single continuous range. The time zone is specified through `time_zone_format`, which supports two variants:
+
+- **`aligned_in_absolute_time`** — all geographies share the same timestamps in absolute time. Provide a single `time_zone` (an IANA time zone string such as `"America/New_York"` or `"Etc/GMT+5"`).
```
elainethale: Is `America/New_York` supported? If not, please don't list it and also add some clarifying text so users know they have to use constant offset time zones. If so, are there specific requirements for timestamps around DST transitions?

docs/source/user_guide/dataset_registration/dimension_concepts.md
```diff
@@ -62,34 +61,128 @@ Records can also be listed directly in the configuration. For example:
...
+- **`aligned_in_std_clock_time`** — timestamps cover the same interval of standard clock time across geographies (e.g., all of 2012 as experienced locally in standard time). The data table must have a `time_zone` column with per-row IANA time zones. Provide a `time_zones` list of all unique time zones in the data table.
```
elainethale: This description needs to clarify that only constant-offset IANA time zones are allowed.

Also, can we provide users with a reason why they need to list the time_zones they are using? Do we use that list to pre-check that the time zones are all constant-offset, for example?

docs/source/user_guide/dataset_registration/data_file_formats.md
elainethale: (Feel free to streamline/remove content in favor of links to other locations if that's appropriate.)

docs/source/user_guide/dataset_registration/dimension_concepts.md
elainethale: I see that this clarification is below, but it's worth repeating early, often, and very clearly. We can't count on people reading everything carefully and all the way through.

docs/source/user_guide/dataset_registration/index.md
```diff
@@ -16,6 +16,8 @@ registration_checks
 ## How-Tos
 
 - [How to Create Dataset Dimensions](../how_tos/how_to_dimensions)
+- [How to Define a Time Dimension](../how_tos/how_to_time_dimension)
+- [How to Convert Time Zones](../how_tos/how_to_convert_time_zone)
```
elainethale: Does time zone conversion belong here or under dataset submittal?

docs/source/user_guide/how_tos/how_to_convert_time_zone.md
```diff
@@ -0,0 +1,226 @@
...
+2012-12-31 23:00:00  | Etc/GMT+5  | g_east    | 10.9
+...
+2011-12-31 23:00:00  | Etc/GMT+6  | g_west    | 5.3
+2012-01-01 00:00:00  | Etc/GMT+6  | g_west    | 2.0
+2012-12-31 21:00:00  | Etc/GMT+6  | g_west    | 15.3
+2012-12-31 22:00:00  | Etc/GMT+6  | g_west    | 12.0
+...
+```
+
+The same absolute UTC instant (`2012-01-01 05:00:00+00`) appears as midnight Eastern but one hours before midnight for the Central time zone.
```
elainethale:
```suggestion
The same absolute UTC instant (`2012-01-01 05:00:00+00`) appears as midnight Eastern but one hour before midnight for the Central time zone.
```

docs/source/user_guide/how_tos/how_to_convert_time_zone.md
```diff
@@ -0,0 +1,226 @@
...
+
+The `convert_time_zone_by_column` functions accept an optional `wrap_time_allowed: bool` parameter (default `False`). It controls what happens when the per-row timezone offsets shift some timestamps **outside** the nominal time range of the source data.
+
+**Why this arises**: suppose your tz-aware source data covers the calendar year 2012 in UTC. Converting a row in Eastern standard time (`UTC-5`) shifts the first 5 hours backward to `2011-12-31 19:00–23:00 EST`, which falls before the 2012 boundary. Without wrapping, those hours are preserved at their true local clock positions, so the output starts on `2011-12-31` for Eastern rows.
+
+**Effect of `wrap_time_allowed=True`**: Chronify reorders the timestamps cyclically so the output covers the *same nominal range* as the source schema in tz-naive clock time. The 5 early-morning hours that would have landed in 2011 are instead placed at the end of the year (after `2012-12-31 19:00 EST`). The result is still 8760 hours, all within 2012, just reordered.
+
+| `wrap_time_allowed` | Behavior |
+|---|---|
+| `False` (default) | Output timestamps reflect true local clock time; range may shift relative to the source |
```
elainethale:
```suggestion
| `False` (default) | Output timestamps reflect true local clock time; range may shift relative to the input data, or may vary by geography |
```

docs/source/developer_guide/how_tos/how_to_convert_time_zone.md
```diff
@@ -0,0 +1,226 @@
...
+---
+
+## Choosing between the two patterns
+
+| Scenario | Use |
+|---|---|
+| All data should reflect the same local time | `convert_time_zone` |
+| Each geography has its own local time zone | `convert_time_zone_by_column` |
+
+---
```
elainethale, line 195: I find this kind of confusing. dsgrid users don't generally choose the pattern--they describe what data they are providing and what they want out and then dsgrid converts as needed to do what the user wants.

docs/source/developer_guide/how_tos/how_to_convert_time_zone.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Query backend
```
elainethale, line 197: This information seems more like detailed information than a how-to. Perhaps this should be a page under "Project Queries"? And then I do think a link from this how-to to that page would be appropriate.

docs/source/developer_guide/how_tos/how_to_convert_time_zone.md
```diff
@@ -0,0 +1,226 @@
...
+
+---
+
+## Relationship to `localize_time_zone`
+
+Time zone *conversion* and time zone *localization* are inverses:
+
+| Operation | Input | Output |
+|---|---|---|
+| `localize_time_zone` | `timestamp_ntz` | `timestamp_tz` |
```
elainethale, line 217: Above didn't you say that dsgrid always outputs timestamp_ntz because the backends like to report in system time zone???

docs/source/developer_guide/how_tos/how_to_convert_time_zone.md
elainethale: And as its own page, or as part of its own page, it could be expanded. What might users want to know about the differences between DuckDB and Spark? (And do we have some of that information somewhere else already?)

elainethale (changes requested):
Not quite all the way through, but I want to provide feedback ahead of a meeting. 

Changes generally look good. There is a little more work to do on the documentation. Please see my comments and generally take one more pass through from a user perspective.

docs/source/user_guide/dataset_registration/dimension_concepts.md
lixiliu: Alignment is separate from column_format. Localization is only required when column_format is NTZ and the only time time zones are restricted to fixed offset. In short, time zone requirement is tied to column_format, not alignment. It is why all that is already mentioned in the column_format section preceding it.

docs/source/user_guide/dataset_registration/dimension_concepts.md
lixiliu: See column_format preceding it.

docs/source/user_guide/dataset_registration/index.md
lixiliu: Good point, it should fall under project query rather than dataset submittal. However I don't see a section on that.

docs/source/developer_guide/how_tos/how_to_convert_time_zone.md
lixiliu: Localize_time_zone makes the timestamps tz-aware...
Convert_time_zone is when dsgrid outputs the timestamps as tz-naive to show the actual time zone conversion. This was what we talked about during the meeting.

dsgrid/config/dimensions.py
```diff
@@ -632,13 +713,31 @@ def handle_legacy_fields(cls, values):
...
+        is specified in `time_zone_format`. DST-observing zones are rejected because they
+        produce ambiguous or missing timestamps at daylight saving time transitions.
+        """
+        is_tz_naive = self.column_format.dtype == "timestamp_ntz" or (
+            self.column_format.dtype == "time_format_in_parts"
+            and self.column_format.offset_column is None
+        )
+        if not is_tz_naive:
+            return self
+        for tz_name in self.time_zone_format.get_time_zones():
```
elainethale, line 764: Is this the only real use of the time_zones list in the config?

elainethale (approved):
LGTM, @lixiliu  -- thanks for sticking with it! @daniel-thom, any remaining concerns?

daniel-thom (approved):

PR #414, opened 2026-04-11
Replace Spark/DuckDB APIs with Ibis

daniel-thom:
Depends on https://github.com/NatLabRockies/chronify/pull/66

Closes GitHub Issue #

## Checklist

- [ ] Tests exercising the new feature or bug fix
- [ ] All tests pass
- [ ] At least one code review approval
- [ ] Consider transferring TODOs to GitHub

PR #422, opened 2026-06-25
Add cross-backend time-zone resolution tests

elainethale:
## What

Adds cross-backend (DuckDB + Spark) tests that pin the time-zone resolution the time migration depends on. Targets `feat/use-ibis` so the Spark CI job runs them alongside DuckDB.

Addresses three open review comments on #414:
- [r3162523396](https://github.com/dsgrid/dsgrid/pull/414#discussion_r3162523396) — "has all the time-zone stuff been thoroughly checked?"
- [r3453966611](https://github.com/dsgrid/dsgrid/pull/414#discussion_r3453966611) — audit/assert the DuckDB naive-vs-`TIMESTAMPTZ` extraction trap.
- [r3453966617](https://github.com/dsgrid/dsgrid/pull/414#discussion_r3453966617) — add a leap-year/TZ test for the annual→datetime divisor.

## Tests

**`tests/test_ibis_tz.py`**
- `test_custom_time_zone_affects_tz_aware_extraction` — under `custom_time_zone`, `.year()` on a TZ-aware instant flips `2021` (UTC) → `2020` (LA). Runs on both backends.
- `test_custom_time_zone_ignored_on_naive_duckdb_timestamp` (DuckDB-only) — pins the trap from the `tz.py` docstring: a naive `TIMESTAMP` ignores the connection `TimeZone`.

**`tests/test_annual_time.py`**
- `test_map_annual_time_leap_year_tz_boundary` — a DateTime dimension whose hours span two UTC years but one local **leap** year (`America/Los_Angeles`, `2020-12-31 14:00..23:00`). The map must resolve a single local year (not raise "more than one year") and apply the `366*24` divisor. This is a genuine discriminator: without TZ-aware resolution `.year()` yields `{2020, 2021}`.
- `test_datetime_chronify_timestamps_are_tz_aware` — pins that chronify timestamps reach DuckDB as `timestamp with time zone`.

## Verification

Run locally on **both** backends:
- `DSGRID_BACKEND_ENGINE=duckdb pytest tests/test_ibis_tz.py tests/test_annual_time.py` → 9 passed
- `DSGRID_BACKEND_ENGINE=spark  pytest tests/test_ibis_tz.py tests/test_annual_time.py` → 8 passed, 1 skipped (the DuckDB-only naive case)

## Note on r3431568749

I considered adding direct tests for the `from_utc_timestamp`/`to_utc_timestamp` test helpers. I left them out deliberately: they are test-only, collapse to a single DuckDB `timezone()` call that does **not** distinguish the two directions, and the existing code comment already flags them as "not fully correct or ideal" (production time mapping goes through chronify). A cross-backend equivalence test would codify known-imperfect behavior rather than protect production. The chronify-based mapping path — which these tests cover — is the production-relevant one.

### DIFF
--- a/dsgrid/config/annual_time_dimension_config.py
+++ b/dsgrid/config/annual_time_dimension_config.py
@@ -17,7 +17,7 @@ from dsgrid.ibis.operations import cross_join, filter_sql
 from dsgrid.ibis.table_utils import table_column_to_list
 from dsgrid.utils.timing import timer_stats_collector, track_timing
 from dsgrid.ibis.session import get_runtime_session
-from dsgrid.ibis.tz import custom_time_zone
+from dsgrid.ibis.tz import assert_tz_aware_extraction, custom_time_zone
 from .dimensions import AnnualTimeDimensionModel
 from .time_dimension_base_config import TimeDimensionBaseConfig
 
@@ -165,6 +165,10 @@ def map_annual_time_to_date_time(
     # ``timestamp('UTC')`` (TIMESTAMPTZ). See dsgrid.ibis.tz for the broader
     # cross-backend contract.
     with custom_time_zone(dt_dim.model.time_zone_format.time_zone):
+        # Guard the load-bearing assumption: on DuckDB, .year() honors the connection
+        # TimeZone only for TZ-aware columns, so a regression to a naive timestamp would
+        # silently resolve the year in UTC. Fail loudly instead.
+        assert_tz_aware_extraction(dt_df[time_col])
         years = table_column_to_list(
             dt_df.select(year=dt_df[time_col].year()).distinct(),
             "year",
--- a/dsgrid/config/annual_time_dimension_config.py
+++ b/dsgrid/config/annual_time_dimension_config.py
@@ -17,7 +17,7 @@ from dsgrid.ibis.operations import cross_join, filter_sql
 from dsgrid.ibis.table_utils import table_column_to_list
 from dsgrid.utils.timing import timer_stats_collector, track_timing
 from dsgrid.ibis.session import get_runtime_session
-from dsgrid.ibis.tz import assert_tz_aware_extraction, custom_time_zone
+from dsgrid.ibis.tz import custom_time_zone
 from .dimensions import AnnualTimeDimensionModel
 from .time_dimension_base_config import TimeDimensionBaseConfig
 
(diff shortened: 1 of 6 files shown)


tests/test_annual_time.py
```diff
@@ -244,6 +250,94 @@ def test_map_annual_time_total_to_datetime_with_existing_model_year(
...
+    dt_dim = date_time_dimension_year_boundary
+    timestamps = make_time_range_generator(dt_dim.to_chronify()).list_timestamps()
+    time_col = dt_dim.get_load_data_time_columns()[0]
+    dt_df = get_runtime_session().createDataFrame(
+        [(x.to_pydatetime(),) for x in timestamps], schema=[time_col]
+    )
+    if use_duckdb():
+        assert str(dt_df.schema()[time_col]).startswith("timestamp('UTC'")
+    else:
+        assert str(dt_df.schema()[time_col]).startswith("timestamp")
```
elainethale: These seem like lame ways to test for being TZ-aware. Would it make sense to use type definitions in ibis.types (or similar)?

dsgrid/ibis/tz.py
```diff
@@ -29,11 +29,49 @@
...
+    Parameters
+    ----------
+    column : ibis.Column
+        The timestamp column about to be extracted from inside a ``custom_time_zone`` block.
+
+    Raises
+    ------
+    DSGInvalidOperation
+        On DuckDB, if ``column`` is not a TZ-aware timestamp.
+    """
```
elainethale: This doc string seems overly long for a simple function.

dsgrid/config/annual_time_dimension_config.py
```diff
@@ -165,6 +165,10 @@ def map_annual_time_to_date_time(
     # ``timestamp('UTC')`` (TIMESTAMPTZ). See dsgrid.ibis.tz for the broader
     # cross-backend contract.
     with custom_time_zone(dt_dim.model.time_zone_format.time_zone):
+        # Guard the load-bearing assumption: on DuckDB, .year() honors the connection
+        # TimeZone only for TZ-aware columns, so a regression to a naive timestamp would
+        # silently resolve the year in UTC. Fail loudly instead.
+        assert_tz_aware_extraction(dt_df[time_col])
```
elainethale: This isn't a general guard. I was expecting there to just be an assert within the `custom_time_zone` function itself.

elainethale: Done — folded the check into `custom_time_zone` itself in 789d51e9. It now takes optional `*tz_aware_columns` and validates each on entry (raising `DSGInvalidOperation` on a naive DuckDB column); the standalone `assert_tz_aware_extraction` helper is gone. The annual call site is now just `with custom_time_zone(tz, dt_df[time_col]):`. The varargs default to empty, so the whole-query callers in `query_submitter.py` are unaffected — and those blocks set the session TZ for downstream SQL/chronify rather than doing any direct ibis `.year()`/`.hour()` extraction, which is the only thing this guard protects.

dsgrid/ibis/tz.py
elainethale: Trimmed in 789d51e9 — the long standalone docstring is gone now that the logic lives inside `custom_time_zone`, which carries a short note plus a `Raises` section.

tests/test_annual_time.py
elainethale: Agreed — replaced the dtype-string matching in 789d51e9. Added `is_tz_aware_timestamp(dtype)` to `dsgrid.ibis.types` (`isinstance(dtype, Timestamp) and dtype.timezone is not None`), alongside the existing `is_string_type`/`is_string_column` predicates, and use it in the tests and in the `custom_time_zone` guard. The Spark branch uses ibis's own `dtype.is_timestamp()`, since Spark timestamps report tz-naive but render via the session TZ.

PR #423, opened 2026-07-10
Standardize aggregation and filter-operator vocabularies behind registries

elainethale:
Addresses the design flagged in dsgrid/dsgrid#414 [comment 4675630295](https://github.com/dsgrid/dsgrid/pull/414#issuecomment-4675630295): the aggregation and filter-operator vocabularies were scattered across hard-coded sets, special cases, and `match` arms, while data types had a single source of truth in `ibis/types.py:TYPE_SPECS`. This PR gives both vocabularies the same treatment.

## What changed

**Aggregations — `dsgrid/ibis/aggregations.py` (new).** Each `AggregationSpec` carries the user-facing name, the Ibis reduction method for the native fast path, and per-backend SQL templates for the raw-SQL fallback. `_aggregate_value` derives both its fast-path allowlist and its fallback SQL from the registry; the `mean→AVG` special case and the no-op `ibis_op = "mean" if op_name == "mean" else op_name` are gone (both are now the `mean` row's data).

*Policy: registry + passthrough.* Unregistered identifiers are still forwarded verbatim to the backend, preserving current flexibility, but `AggregationModel` now logs a warning listing the registered names — so "what can a user invoke and how do they discover it?" is answered at query-build time, in the CLI `--aggregation-function` help, and in the module docstring. `ColumnModel` scalar extractions (`hour`, `year`, …) deliberately do not warn.

**Operators — `dsgrid/dimension/filter_operators.py` (new).** Each `FilterOperatorSpec` carries the name, a family bucket (string/null/membership/range), the value shape it expects, and the callable building its Ibis expression. The operator set, the pydantic validator, and the dispatch all derive from the registry, replacing the three manually-synced copies (`DIMENSION_COLUMN_FILTER_OPERATORS`, `check_operator`, the `match` arms) plus the standalone `_apply_between`. `between` is now a registered operator; `DimensionFilterBetweenColumnOperatorModel` keeps its exact shape for saved-query compatibility and routes through the same dispatcher.

## Testing strategy: characterize first

The first commit adds characterization tests that pass against **unmodified** production code on both backends (43 tests), pinning both `_aggregate_value` paths for every op, path agreement for the same logical query (the dual-engine concern from r3453966621), verbatim passthrough of unregistered names, and unit-level coverage gaps in the filter tests. The refactor commits were then required to keep those tests green verbatim, with one documented exception (below). A mutation-style test (`test_registry_drives_both_paths`) redefines `sum` in the registry and asserts both execution paths follow it.

## Behavior changes (all additive or bug fixes)

- **Bug fix:** the SQL fallback used to uppercase `approx_median` into `APPROX_MEDIAN(value)`, which exists on **neither** backend (DuckDB: `Scalar Function with name approx_median does not exist`; Spark SQL has `percentile_approx`). The registry emits `APPROX_QUANTILE(x, 0.5)` / `PERCENTILE_APPROX(x, 0.5)`. The characterization test pinning the old failure was flipped to pin the fix — the only Phase-0 test the refactor touched. Every other registered name produces byte-identical SQL to the old `.upper()` path.
- `operator="between"` with a two-element value now validates on the column-operator models — exactly what `DimensionFilterColumnOperatorModel.value`'s field description and the how-to docs have promised all along (that documented usage previously raised `ValidationError`).
- `check_operator`'s error message sorts the allowed names (was nondeterministic set ordering).
- Docs: the `between_column_operator` example in `how_to_filter.md` used fields the model doesn't have (`column1`/`column2`/`operator`/`value`); fixed to `column`/`lower_bound`/`upper_bound`.

Out of scope, deliberately: the expression models' raw-SQL operator interpolation (deferred separately in r3453966647) and chronify (no aggregation names cross that boundary; `MeasurementType` is a data-semantics tag, not a reduction dsgrid runs).

## Verification

- DuckDB: 108 passed (`test_unpivoted_table_aggregation`, `test_dimension_filters`, `test_query_models`, `test_ibis_operations`)
- Spark: 104 passed, 4 skipped (pre-existing DuckDB-only relocation tests)
- `pre-commit` and `ty check` clean

### DIFF
--- a/dsgrid/cli/query.py
+++ b/dsgrid/cli/query.py
@@ -127,7 +127,7 @@ $ dsgrid query project create --default-result-aggregation my_query_result_name 
     show_default=True,
     help="Aggregation function for any included default aggregations. "
     f"Registered: {', '.join(sorted(SUPPORTED_AGGREGATIONS))}. Other names are "
-    "forwarded to the backend as-is.",
+    "uppercased and forwarded to the backend.",
 )
 @click.option(
     "-f",
--- a/dsgrid/cli/query.py
+++ b/dsgrid/cli/query.py
@@ -26,6 +26,7 @@ from dsgrid.dimension.dimension_filters import (
     SupplementalDimensionFilterColumnOperatorModel,
 )
 from dsgrid.filesystem.factory import make_filesystem_interface
+from dsgrid.ibis.aggregations import SUPPORTED_AGGREGATIONS
 from dsgrid.query.dataset_mapping_plan import DatasetMappingPlan
 from dsgrid.query.derived_dataset import create_derived_dataset_config_from_query
 from dsgrid.query.models import (
@@ -124,7 +125,9 @@ $ dsgrid query project create --default-result-aggregation my_query_result_name 
     "--aggregation-function",
     default="sum",
     show_default=True,
-    help="Aggregation function for any included default aggregations.",
+    help="Aggregation function for any included default aggregations. "
+    f"Registered: {', '.join(sorted(SUPPORTED_AGGREGATIONS))}. Other names are "
+    "forwarded to the backend as-is.",
 )
 @click.option(
     "-f",
--- a/dsgrid/dataset/unpivoted_table.py
+++ b/dsgrid/dataset/unpivoted_table.py
@@ -129,18 +129,12 @@ def _get_metric_column_name(context: QueryContext, metric_query_name):
 
 
 def _aggregate_value(df: ibis.Table, group_by_cols: list[str], op_name: str) -> ibis.Table:
-    # Fast path: when all group-by entries are bare column references and
-    # ``op_name`` is a registered reduction, dispatch through native Ibis to
-    # keep the chain lazy and let the planner fuse adjacent aggregations.
-    #
-    # Only registered names take this path because ``FunctionReference``
-    # accepts any backend SQL identifier. Without that restriction, names
-    # like ``round`` would resolve to Ibis's scalar column method
-    # ``df[col].round`` instead of a reduction, and
-    # ``df.group_by(...).aggregate(c=df[col].round())`` is a non-aggregate
-    # inside an aggregate — which either errors or silently produces wrong
-    # SQL. Unregistered names fall through to the SQL-string path below,
-    # which forwards them verbatim to the backend.
+    # Fast path: registered reductions over bare group-by columns dispatch
+    # through native Ibis, keeping the chain lazy. Everything else takes the
+    # SQL path below. Unregistered names must not reach the fast path:
+    # resolving an arbitrary identifier against the Ibis column API could
+    # pick a scalar method (e.g. ``round``), which is invalid inside
+    # ``aggregate()``.
     spec = find_aggregation_spec(op_name)
     bare_cols = [col for col in group_by_cols if _looks_like_bare_column(col, df.columns)]
     if spec is not None and len(bare_cols) == len(group_by_cols):
--- a/dsgrid/dataset/unpivoted_table.py
+++ b/dsgrid/dataset/unpivoted_table.py
@@ -5,6 +5,7 @@ import ibis
 
 from dsgrid.common import VALUE_COLUMN
 from dsgrid.dimension.base_models import DimensionType
+from dsgrid.ibis.aggregations import find_aggregation_spec, sql_aggregate_expression
 from dsgrid.ibis.backend import get_runtime_backend
 from dsgrid.ibis.operations import create_temp_view, handle_column_spaces
 from dsgrid.query.models import (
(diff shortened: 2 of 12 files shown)


dsgrid/dataset/unpivoted_table.py
```diff
@@ -127,28 +128,23 @@ def _get_metric_column_name(context: QueryContext, metric_query_name):
...
+    # accepts any backend SQL identifier. Without that restriction, names
+    # like ``round`` would resolve to Ibis's scalar column method
+    # ``df[col].round`` instead of a reduction, and
     # ``df.group_by(...).aggregate(c=df[col].round())`` is a non-aggregate
     # inside an aggregate — which either errors or silently produces wrong
-    # SQL. Anything not in the allowlist falls through to the SQL-string
-    # path below, which forwards the name verbatim to the backend.
-    ibis_op = "mean" if op_name == "mean" else op_name
+    # SQL. Unregistered names fall through to the SQL-string path below,
+    # which forwards them verbatim to the backend.
```
elainethale: Could this explanation be more concise/simpler?

dsgrid/ibis/filter_operators.py
elainethale, line 1: Should this be in the ibis subpackage? I can see it going either way, but if it is was in `ibis` the module name could be filter_operators.py.

dsgrid/query/models.py
```diff
@@ -176,6 +181,13 @@ class AggregationModel(DSGBaseModel):
     @classmethod
     def check_aggregation_function(cls, aggregation_function):
         if isinstance(aggregation_function, str):
+            if aggregation_function not in SUPPORTED_AGGREGATIONS:
+                logger.warning(
+                    "aggregation_function=%r is not a registered dsgrid aggregation; it will "
+                    "be forwarded verbatim to the backend, which may reject it. Registered: %s",
+                    aggregation_function,
+                    sorted(SUPPORTED_AGGREGATIONS),
+                )
```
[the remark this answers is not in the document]
elainethale: Good catch — the wording was wrong, not the behavior. Fixed in 6a014f8e: the warning now says "uppercased and forwarded". I kept the uppercasing itself because it is the pre-registry behavior (pinned by the characterization tests), and since SQL function names are case-insensitive on both backends, forwarding truly verbatim would change the emitted SQL for zero semantic gain.

dsgrid/cli/query.py
```diff
@@ -124,7 +125,9 @@ def _add_options(func):
     "--aggregation-function",
     default="sum",
     show_default=True,
-    help="Aggregation function for any included default aggregations.",
+    help="Aggregation function for any included default aggregations. "
+    f"Registered: {', '.join(sorted(SUPPORTED_AGGREGATIONS))}. Other names are "
+    "forwarded to the backend as-is.",
 )
```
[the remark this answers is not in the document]
elainethale: Fixed in 6a014f8e — help text now reads "Other names are uppercased and forwarded to the backend."

dsgrid/dataset/unpivoted_table.py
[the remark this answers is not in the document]
elainethale: Fixed in 6a014f8e — the comment was rewritten (and condensed, per the other thread on these lines); it no longer claims verbatim forwarding.

dsgrid/ibis/aggregations.py
```diff
@@ -0,0 +1,116 @@
...
+:class:`~dsgrid.query.models.AggregationModel`. Every per-function
+representation (Ibis reduction method for the native fast path, DuckDB and
+Spark SQL expressions for the raw-SQL fallback) lives on an
+:class:`AggregationSpec`.
+
+dsgrid deliberately does not reject unregistered names: any plain SQL
+identifier accepted by :class:`~dsgrid.query.models.FunctionReference` is
+forwarded verbatim to the backend, which may reject it at execution time.
+Registered names get the native fast path and backend-correct SQL spellings;
+:func:`sql_aggregate_expression` implements both rules in one place.
```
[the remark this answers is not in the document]
elainethale: Fixed in 6a014f8e — the module docstring now says "uppercased and forwarded" and notes that SQL function names are case-insensitive on both backends, so the uppercasing does not change which function runs.

dsgrid/ibis/aggregations.py
```diff
@@ -0,0 +1,116 @@
...
+    return spec
+
+
+def sql_aggregate_expression(op_name: str, column: str) -> str:
+    """Build the SQL aggregation expression for ``op_name`` over ``column``.
+
+    Registered names use their per-backend template. Unregistered names are
+    uppercased and forwarded verbatim, preserving dsgrid's passthrough
+    contract for any backend function a user cares to name; the backend
+    rejects names it does not know at execution time.
```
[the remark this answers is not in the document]
elainethale: Fixed in 6a014f8e — now "uppercased and forwarded otherwise unchanged".

dsgrid/dataset/unpivoted_table.py
elainethale: Condensed from 12 lines to 6 in 6a014f8e. It now leads with what the fast path is and keeps only the one non-obvious fact: unregistered names must not reach the fast path because resolving an arbitrary identifier against the Ibis column API could pick a scalar method (e.g. `round`), which is invalid inside `aggregate()`.

dsgrid/ibis/filter_operators.py
elainethale: Agreed — moved to `dsgrid/ibis/filter_operators.py` in 6a014f8e (module name unchanged, as you suggested it could be there). The symmetry argument won me over: both registries map user-facing names to Ibis expressions, so they now sit side by side with `TYPE_SPECS` in the abstraction layer, and the import direction stays clean (`dsgrid.dimension.dimension_filters` already imports from `dsgrid.ibis`).

dsgrid/query/models.py
elainethale: Is this treatment okay? Does this wrap registered aggregation functions in FunctionReference, or do those fall through to the `return`? Should there be a type check on non str and not None `aggregation_function` values?

elainethale: Good questions — answering each, then what changed in e91f7544:

1. **Registered functions were wrapped, not fall-through.** The wrap was unconditional inside the `isinstance(str)` branch; only the *warning* was conditional on registration. But the nesting made that genuinely hard to read, which is a problem in itself.
2. **Non-str/non-None values fell through untouched — and that was a latent trap.** Only a `FunctionReference` actually works downstream (execution reads `op.name` in `unpivoted_table.py`, and `ColumnModel.function` is consumed as `column.function.name`). Anything else — e.g. a pyspark callable like `F.sum` from pre-Ibis habits — validated fine and then died at query time with an opaque `AttributeError`.
3. **Yes, there should be a type check, and now there is.** The validator handles each case explicitly: `None` rejected, `FunctionReference` passed through, `str` warned-if-unregistered and wrapped, anything else rejected at parse time with the expected types in the message. `ColumnModel.handle_function` had the same fall-through and got the same treatment (`None` stays allowed there, since no function is valid for a column).

All in-repo callers pass strings, so nothing changes for existing queries; new tests pin the `FunctionReference` pass-through (no warning, identity preserved) and the rejection of ints/callables/lists on both models.

PR #425, opened 2026-08-10
Fix regressions found in review of the Ibis migration (follow-up to #414)

daniel-thom:
## Summary

Follow-up to review findings on #414 (the Ibis migration). Correctness reviews against `main` surfaced a set of behavioral regressions introduced by the migration; this branch restores the pre-Ibis contracts and pins each one with tests.

### Restored / fixed behavior

- **FileSchema type handling** (`69ae815b`): declared types are now authoritative for JSON (cross-family casts such as STRING on an inferred numeric column are applied, matching CSV) and validated against Parquet's self-describing schema (a conflicting declaration raises instead of being silently ignored). Removes the now-unused `strict_family` mode.
- **NULL handling**: `first`/`last` skip NULLs on both aggregation paths (`286ea16c`); the association/distinct validators pin NULL semantics (`2cc7a663`); `get_unique_values_per_column` removed so NULLs are not dropped (`86cd3f9a`).
- **Spark runtime**: reading JSON uses FAILFAST again so malformed records raise instead of nulling out (`895c3ac2`); a live session's TZ is no longer reset (`7a7789e0`); the salted key is repartitioned again (`b97df084`).
- **Data safety**: `write_dataframe` refuses to overwrite an existing path again (`56b14c4d`).
- **Query/config validation**: filter values that would match zero rows are rejected (`fcdf44fe`); annual-time mapping rejects mismatched `model_year` (`b6f9afe0`); group-by metadata is only dropped for function/alias columns (`a2aee626`); the legacy rc-field deprecation is visible again (`f766e155`).

### Cleanup

Dead pre-Ibis branches removed (`3db7eace`), typing fixes for `ty` (`02c44fb0`, `bea4fbbe`), a backend-specific test assumption fixed (`9eb938f6`), and `rust/Cargo.lock` tracked with `ethnum` floored at 1.5.3 (`e380f5d3`).

## Test plan

- `tests/test_file_schema.py` and the other touched suites pass on both backends (DuckDB and Spark, `DSGRID_BACKEND_ENGINE=spark`).
- `ty check` and `ruff` clean.


dsgrid/ibis/io.py
```diff
@@ -600,25 +614,37 @@ def write_dataframe(df: ibis.Table, filename: str | Path, overwrite: bool = Fals
...
+        write_table(df, name, "parquet", overwrite=overwrite)
     elif suffix == ".csv":
-        write_table(df, name, "csv")
+        write_table(df, name, "csv", overwrite=overwrite)
     elif suffix == ".json":
         if use_duckdb():
             new_name = name.replace(".json", ".parquet")
-            write_table(df, new_name, "parquet")
+            write_table(df, new_name, "parquet", overwrite=overwrite)
         else:
```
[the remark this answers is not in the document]
daniel-thom: Fixed in aa2e79ee by removing the redirect entirely. It predated `write_table` gaining a line-delimited JSON writer on DuckDB (`COPY ... (FORMAT JSON)`), so it was vestigial — and beyond the overwrite check landing on the wrong path, it left `file_schema`'s rename path recording a `schema.path` that was never written. JSON now writes to the requested path on both backends, so the overwrite contract applies to the user-provided path.

dsgrid/utils/dataset.py
```diff
@@ -1084,35 +1049,22 @@ def merge_expected_associations_tables(
...
                 set_dim_cols = sorted(c for c in col_set if c in all_dim_records)
-                merged_unique = (
-                    get_unique_values_per_column(merged, covered_dim_cols)
-                    if covered_dim_cols
-                    else {}
-                )
-                df_unique_overlap = (
-                    get_unique_values_per_column(df, set_dim_cols) if set_dim_cols else {}
-                )
                 pre_join_values: dict[str, set[str]] = {}
```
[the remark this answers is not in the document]
daniel-thom: Fixed in 15923de1 — the annotation is now `dict[str, set]`, matching what the distinct-value sets actually hold (they can contain `None`). The loops here also switched to the batched `get_unique_values_per_column` (see the reply on `_check_distinct_column_values`).

dsgrid/utils/utilities.py
```diff
@@ -115,6 +115,39 @@ def check_uniqueness(iterable: Iterable, tag: str) -> set[str]:
...
+    [None, 'a', 'b']
+    """
+    non_null = []
+    has_null = False
+    for value in values:
+        if value is None:
+            has_null = True
+        else:
+            non_null.append(value)
+    non_null.sort()
```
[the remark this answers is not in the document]
daniel-thom: Fixed in d7cdd4f9. Agreed that a TypeError here would mask the error being reported — `sorted_with_nulls` now falls back to ordering by `(type name, str(value))` when direct comparison raises, with a test covering the `{6037, "6037", None}` case.

dsgrid/registry/project_registry_manager.py
```diff
@@ -1595,23 +1594,18 @@ def _check_distinct_column_values(project_table: ibis.Table, mapped_dataset_tabl
...
-    project_values = get_unique_values_per_column(project_table, columns)
-    mapped_values = get_unique_values_per_column(mapped_dataset_table, columns)
-
     has_mismatch = False
-    for column in columns:
-        diff = project_values[column].difference(mapped_values[column])
+    for column in project_table.columns:
+        project_values = get_unique_values(project_table, column)
+        mapped_values = get_unique_values(mapped_dataset_table, column)
+        diff = project_values.difference(mapped_values)
```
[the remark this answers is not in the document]
daniel-thom: Fixed in 15923de1. `get_unique_values_per_column` is restored with NULL preservation: the same single aggregation computes `collect(distinct=True)` plus `count(col)` per column, and `count(col) < count(*)` — exactly the condition under which `.distinct()` would have produced a NULL row — adds `None` to that column's set. So it's one execute per side again, with sets identical to the per-column loop (and to pre-Ibis main). All three per-column loops now use it: `_check_distinct_column_values`, the association check in `dataset_schema_handler_base`, and the loops in `merge_expected_associations_tables`, where each per-column call had been re-executing the whole join pipeline. Verified on both backends, including NULL and empty-table edge cases.

dsgrid/dataset/table_format_handler_base.py
```diff
@@ -215,9 +215,14 @@ def _build_group_by_columns(
...
                     group_by_cols.append(expr)
-                    if not isinstance(expr, str) or expr != column.dimension_name:
+                    if column.function is not None:
                         # In this case we are replacing any existing query name with an expression
                         # or alias, and so the old name must be removed.
+                        # Test the function rather than the expression: _make_group_by_column_expr
+                        # quotes bare column names, so comparing expr to dimension_name would be
+                        # true for every plain column and would drop metadata that an earlier
+                        # aggregation recorded under an alias (final_metadata spans the whole
+                        # aggregation loop in process_stacked_aggregations).
```
elainethale: nit: This comment is documenting why the current code is better than the now-deprecated code. Future readers won't care.

dsgrid/ibis/session.py
```diff
@@ -276,8 +276,27 @@ def _create_spark_session(name="dsgrid", check_env=True, spark_conf=None) -> Any
...
+    # Callers needing another render TZ set it explicitly via ``spark_conf=``,
+    # ``SPARK_CONF_DIR``, or :func:`dsgrid.ibis.tz.custom_time_zone`.
+    #
+    # Apply the default only when this call creates the session. getOrCreate returns an
+    # active session but still pushes conf's spark.sql.* values onto it, so an
+    # unconditional default would reset a TZ another component set on the live session
+    # (e.g. inside custom_time_zone) whenever any code path calls init_runtime_session
+    # mid-flow. ``conf`` cannot answer "did the caller ask for a TZ?" — a fresh
+    # SparkConf() inherits the JVM defaults, which already carry the UTC dsgrid set when
+    # it created the earlier session. Only ``spark_conf`` reflects an explicit request.
```
elainethale, line 292: nit: This is quite in the weeds and hard to follow.

dsgrid/data_models.py
```diff
@@ -124,6 +124,10 @@ def __init__(self, value, description, **kwargs):
 class DSGEnum(Enum):
     """dsgrid Enum class"""
 
+    # Declared, never assigned: an assignment here would make this an enum member.
+    # Each member's value is set in __new__.
+    description: str | None
+
```
elainethale, line 130: Is this a new field, or is it being reinstated post-initial-ibis refactor?

dsgrid/dsgrid_rc.py
```diff
@@ -54,9 +54,15 @@ def warn_legacy_fields(cls, data: dict[str, Any]) -> dict[str, Any]:
...
+                    f"ignored. Please remove it from {RC_FILENAME}. This will cause an error "
                     "in a future release."
                 )
+                # Log as well as warn. This validator runs at ``import dsgrid`` time, and
+                # Python's default filters hide a DeprecationWarning raised from library
+                # code, so the warning alone never reaches a user running the CLI -- which
+                # is exactly the "silently ignored" behavior the deprecation exists to
+                # avoid. The warning is kept for programmatic callers and tests.
+                logger.warning(msg)
                 warn(msg, DeprecationWarning, stacklevel=2)
```
elainethale: Nit: Should we be using `logging.captureWarnings(True)` somewhere more central, rather than calling both `logger.warning` and `warn`? https://stackoverflow.com/a/37979724

elainethale (approved):
LGTM. A few minor comments/questions.

dsgrid/data_models.py
daniel-thom: It was previously a dynamically-added field, and ty was complaining. This fixes the issue with no real behavioral change.

dsgrid/dsgrid_rc.py
daniel-thom: Cool, I didn't know about this.

PR #426, opened 2026-08-20
Port eh/minor_updates fixes onto the Ibis backend

elainethale:
﻿## Description

Re-implements the fixes from #415 (`eh/minor_updates`) on top of the Ibis backend refactor (#414). Each original change is evaluated against the refactored codebase and ported as its own reviewed commit; the port plan with per-change rationale is maintained in `PORT_FIXES_PLAN.md` (local working document; can be committed to the branch if useful).

Progress (implemented step by step, each step reviewed before the next begins):

- [x] **Step A - error messages** (port of e6ba7169): wrap pydantic `ValidationError` with file/record context in `DSGBaseModel.from_file`, `MappingTableModel.from_pre_registered_model`, and `convert_record_dicts_to_classes`; replace the bare `assert` on a missing `mapping_reference` with an informative `DSGInvalidDimensionMapping`; delete dead `create_registration()`. Review fixes: the query validate CLI's dead `except ValidationError`, and a built-but-never-raised error in `check_dataset_mapping_plan`.
- [x] **Step B - lowercase `column_format.dtype` + validator non-dict guards** (port of 3a428b93): normalize the dtype discriminator tag (with rename warning) before union resolution; non-dict guards on all six handle_legacy_fields validators in dimensions.py; new tests/test_dimensions.py.
- [x] **Step F - apply sort/pivot only at the true final save** (port of f834bae9): _postprocess hoisted to QuerySubmitterBase and gated by a final flag on the save methods, so _convert_time_zone's re-save no longer destroys sort order or receives a pivoted table; regression tests assert global sortedness across the tz-conversion variants and for the pivoted format. Review fix: pivot before sorting, since the pivot is a GROUP BY aggregation that does not preserve row order.
- [x] **Step E - chronify schema-staleness shim** (port of 319862b2, e9aad950): _adjust_time_config_for_post_localization in the shared _get_src_schema/_get_dst_schema helpers rebuilds the chronify DatetimeRange as TIMESTAMP_TZ when the table's time column is already localized; regression test verified to fail without the shim. Root-cause GitHub issue drafted, pending approval. Awaiting review.
- [x] **Step C - fail dataset submission when time_zone info is needed but missing** (port of cbc5b438): dataset-config validation no longer skipped by use_project_geography_time_zone; new submission-time check in _submit_dataset; _convert_time_dimension considers the target time dimension and falls back gracefully in standalone dataset queries; all-null time_zone detection at config and runtime layers; docs updated; new tests/test_timezone_validation.py. Awaiting review.
- [x] **Step D - `time_zone_format` for index time** (port of 8ab30e49): IndexTimeDimensionModel gains the time_zone_format union (legacy configs default to aligned_in_std_clock_time with a warning); to_chronify() branches to IndexTimeRange for the single-tz format; project geography tz check now conditional on the time dimensions; index-time docs updated. Deviation: get_time_zone() is format-based, not count-based. Awaiting review.

Dropped as obsolete: the `localize_to_time_zone` legacy-field removal (field no longer exists anywhere).

## Checklist

- [x] All tests pass


dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -880,7 +881,13 @@ def _remap_dimension_columns(
                     dim_mapping.name,
                 )
                 continue
-            assert dim_mapping.mapping_reference is not None
+            if dim_mapping.mapping_reference is None:
+                msg = (
+                    f"Mapping operation '{dim_mapping.name}' has no mapping reference. "
+                    "This can occur when the mapping plan specifies a dimension name that "
```
elainethale: "dimension name" or "dimension mapping name"?

dsgrid/utils/utilities.py
```diff
@@ -175,16 +176,20 @@ def convert_record_dicts_to_classes(iterable, cls, check_duplicates: None | list
...
         elif len(row) != length:
-            msg = f"Rows have inconsistent length: first_row_length={length} {row=}"
+            msg = f"Record {i} has {len(row)} columns; expected {length}: {row}"
             raise ValueError(msg)
-        record = cls(**row)
+        try:
+            record = cls(**row)
+        except ValidationError as e:
+            msg = f"Validation error in record {i}: {row}\n{e}"
+            raise ValueError(msg) from e
```
elainethale, line 194: Should this be a `ValueError` or a different exception class?

tests/test_mapping_tables.py
elainethale, line 1: Is this the right place for these tests?

Are these tests complete and concise for the functionality (better error messages) that was added?

dsgrid/dataset/dataset_schema_handler_base.py
elainethale: Good catch - it was muddled. The plan's `name` is resolved as a *dimension* name (`project_config.get_dimension(mapping.name)` in `check_dataset_mapping_plan`), and the reference stays `None` when no registered dataset-to-project dimension *mapping* matches that dimension. Reworded in 2ea9cc19 to say exactly that.

dsgrid/utils/utilities.py
elainethale: ValueError is deliberate: this function's production callers are pydantic field validators (add_records in mapping_tables.py and dimensions.py), and pydantic converts ValueError raised inside a validator into a ValidationError with proper field context - a DSG exception would escape the validator and surface as a raw traceback instead. It also matches the function's two pre-existing raises. Added a code comment explaining the constraint in 2ea9cc19.

tests/test_mapping_tables.py
elainethale: Partly - reorganized in 2ea9cc19. The convert_record_dicts_to_classes tests moved to tests/test_utilities.py (that module's existing test home) and gained coverage for the ragged-record and None-key messages. tests/test_error_messages.py now holds only the two file-context wraps (from_file, from_pre_registered_model), since no per-module test files exist for data_models or mapping_tables; both now also assert the ValidationError cause chain, and the mapping test asserts the record-level context survives the nesting. Not unit-tested: the mapping_reference-is-None raise and the supplemental-dimension raise (both need a populated registry + handler; exercised indirectly by tests/test_queries.py) - can add integration-level coverage if you want it.

elainethale, line 1: Given that the remaining tests are for mapping table configs, should this filename be test_config_mapping_tables.py or test_mapping_table_configs.py? (What name would be consistent with the rest of the test suite?)

elainethale: The suite's convention is naming test modules after the source module they test (test_dimension_config.py, test_registration_models.py, test_query_models.py), so in ac883ff9 I split it along those lines instead of a combined name: the from_pre_registered_model test is now tests/test_mapping_tables.py (mirroring config/mapping_tables.py) and the from_file test is tests/test_data_models.py (it targets DSGBaseModel.from_file in data_models.py; MappingTableRecordModel is just the vehicle). test_error_messages.py is gone.

dsgrid/query/query_submitter.py
```diff
@@ -122,6 +122,21 @@ def table_filename(path: Path):
...
+
+        Only invoked at the actual final save site; intermediate saves skip this so
+        downstream stages (e.g. time zone conversion in ``_convert_time_zone``)
+        operate on the stacked, value-column-bearing schema they expect.
+        """
+        if context.model.result.sort_columns:
+            df = df.order_by(*context.model.result.sort_columns)
+
+        if isinstance(context.model.result.table_format, PivotedTableFormatModel):
+            df = _pivot_table(df, context)
```
elainethale, line 138: Does _pivot_table reliably preserve sorting?

tests/test_queries.py
elainethale, line 1: Is this a sufficient test? (Were we already asking for a query that both sets a different time zone and sorts, we just weren't testing that sorting actually happened/was preserved and visible for the end user?)

dsgrid/query/query_submitter.py
elainethale: No, it doesn't - good catch. pivot() is pivot_wider with values_agg='sum', i.e. a GROUP BY aggregation, and SQL gives no guarantee that pre-aggregation row order survives. This ordering (sort then pivot) was inherited from the pre-existing DatasetQuerySubmitter._postprocess and the original branch, and nothing ever tested the combination. Fixed in 1132367c: _postprocess now pivots first and sorts last. A side effect worth knowing: sort_columns must name columns that exist in the final (pivoted) table - the group columns survive, the pivoted-out dimension and value column do not - which now fails loudly with a column-not-found error rather than silently sorting a table that the pivot then scrambles.

tests/test_queries.py
elainethale: Half right. The ElectricityValues query family already set time_zone (parametrized over None / America/Los_Angeles / 'geography'), but it did NOT previously ask for sort_columns - no test in the suite combined time_zone with sort_columns or with the pivoted format, which is exactly why the bug was invisible. The change adds sort_columns to that query so all three tz variants now assert global sortedness of the end-user table read back from the output parquet (sort without conversion, sort + named-tz conversion, sort + geography conversion). With 1132367c the pivoted-format query (QueryTestMapAnnualTime) also gains sort_columns plus the same assertion, covering the pivot+sort combination your other comment flagged. Still uncovered: pivot combined with time_zone conversion (no current query requests both) - say the word if you want that variant added too.

dsgrid/utils/dataset.py
```diff
@@ -761,6 +762,60 @@ def _to_chronify_time_based_data_adjustment(
...
+    """Reflect post-localization shape in the chronify time_config.
+
+    ``DateTimeDimensionConfig.to_chronify()`` describes the pre-localization shape
+    (NTZ dtype + naive ``start``) when the dim has a ``localize_to_single_tz`` plan,
+    because some pipelines (e.g. ``localize_time_zone_with_chronify_*``) invoke
+    chronify *before* localization. The query path, however, hands chronify a
+    table whose time column has already been localized (the registered parquet on
+    disk is ``TIMESTAMP WITH TIME ZONE``) while the persisted time-dimension record
+    still says ``timestamp_ntz``. When the actual time column is tz-aware, rebuild
+    the time_config so its dtype and start match the data.
```
elainethale: Please review for accuracy and conciseness. This is quite dense.

dsgrid/utils/dataset.py
```diff
@@ -761,6 +762,60 @@ def _to_chronify_time_based_data_adjustment(
...
+    """
+    if not isinstance(time_dim, DateTimeDimensionConfig):
+        return time_config
+    if time_dim.get_localization_plan() != "localize_to_single_tz":
+        return time_config
+    # ``localize_to_single_tz`` only ever pairs with the ALIGNED_IN_ABSOLUTE_TIME
+    # branch of ``to_chronify()``, which returns ``DatetimeRange``.
+    if not isinstance(time_config, chronify.DatetimeRange):
+        return time_config
```
elainethale: Based on the comment, should this be asserting chronify.DatetimeRange instead?

dsgrid/utils/dataset.py
elainethale: Rewritten in 0e4ff6b0 - about half the length, and it now cites issue #427 for the root-cause story instead of re-explaining it inline. Accuracy re-checked against the code: the claim that localization itself runs through chronify (hence the pre-localization shape from to_chronify) matches localize_time_zone_with_chronify_* usage, and the persisted-record staleness is as documented in the issue.

dsgrid/utils/dataset.py
elainethale: Yes - changed to an assert in 0e4ff6b0. Verified the invariant is airtight today: get_localization_plan() only returns 'localize_to_single_tz' when format_type is ALIGNED_IN_ABSOLUTE_TIME, and that to_chronify() branch unconditionally returns DatetimeRange. A silent pass-through would have buried any future violation (and resurfaced as the original confusing cast error); the assert fails fast at the invariant.

dsgrid/utils/dataset.py
```diff
@@ -761,6 +762,57 @@ def _to_chronify_time_based_data_adjustment(
...
+    df: ibis.Table,
+) -> chronify.TimeBaseModel:
+    """Return ``time_config`` adjusted to match a time column that was already localized.
+
+    With a ``localize_to_single_tz`` plan, ``to_chronify()`` reports the
+    pre-localization shape (NTZ dtype, naive start) because localization itself
+    runs through chronify. After registration the stored data is tz-aware, but
+    the persisted dimension config still says ``timestamp_ntz`` (see #427). If
+    ``df``'s time column is tz-aware, rebuild the config as ``TIMESTAMP_TZ``
+    with a localized start; otherwise return it unchanged.
```
elainethale: I don't think the description is accurate. How about:

```suggestion
    Currently ``to_chronify()`` always reports the original datetime shape, 
    which doesn't match the actual data with a ``localize_to_single_tz`` plan.
    In that case, the pre-localization shape has a NTZ dtype and a naive start,  
    but after registration the stored data is tz-aware (see #427). If 
    ``df``'s time column is tz-aware, rebuild the config as ``TIMESTAMP_TZ``
    with a localized start; otherwise return it unchanged.
```

elainethale: Applied verbatim in the latest commit (trailing whitespace trimmed to satisfy the pre-commit hook). Thanks - your version reads better.

docs/source/user_guide/how_tos/how_to_time_dimension.md
elainethale, line 1: The original implementation also made edits under `**Requirements**:`, `### Common use cases`, etc. Is there a reason those changes weren't ported?

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -986,6 +989,47 @@ def _convert_time_dimension(
...
+                            "Time mapping requires time zone information from a "
+                            "geography dimension, and use_project_geography_time_zone "
+                            "is True, but no target geography dimension is available "
+                            "and the dataset's geography dimension does not include a "
+                            "time_zone column. Either set "
+                            "use_project_geography_time_zone=False and add a time_zone "
+                            "column to the dataset's geography dimension, or run the "
+                            "query through a project."
+                        )
+                        raise DSGInvalidDataset(msg)
```
elainethale: In the previous implementation, this was preceded by `if time_dim.model.is_time_zone_required_in_geography():`. Why is that no longer needed here? Was that branch erroneous to begin with? (It looks like it led to error messages that only differed by "The dataset's time dimension requires ..." compared to "The target time dimension requires ...".)

tests/test_timezone_validation.py
```diff
@@ -0,0 +1,353 @@
...
+
+def test_check_timezone_in_geography_valid_passes(tmp_path):
+    geo_dim = _make_geography_dimension(tmp_path, with_time_zone=True)
+    check_timezone_in_geography(geo_dim)
+
+
+def test_check_timezone_in_geography_custom_err_msg(tmp_path):
+    geo_dim = _make_geography_dimension(tmp_path, with_time_zone=False)
+    with pytest.raises(DSGInvalidDimension, match="Custom error"):
+        check_timezone_in_geography(geo_dim, err_msg="Custom error about missing time zones")
```
elainethale: The "Custom error" expected text confuses me. Also, is this a real test, and if so, how is it different from `test_check_timezone_in_geography_all_null_raises`?

tests/test_timezone_validation.py
```diff
@@ -0,0 +1,353 @@
...
+                pd.Timestamp("2018-01-01 01:00:00"),
+            ],
+            "geography": ["g1", "g1"],
+            VALUE_COLUMN: [1.0, 2.0],
+        }
+    )
+    schema = StructType(
+        [
+            StructField(TIME_COLUMN, TimestampNTZType(), True),
+            StructField("geography", StringType(), True),
```
elainethale: Why "geography" when `TIME_COLUMN` and `VALUE_COLUMN`?

docs/source/user_guide/how_tos/how_to_time_dimension.md
elainethale: Those edits come from the other original commit: cbc5b438 (this step) touched exactly one line of how_to_time_dimension.md, which is what's ported here. The Requirements / Common use cases / index-time-section edits are from 8ab30e49 (time_zone_format for index time) and are planned for Step D, whose doc item covers exactly that section - verified with git show --stat on both commits.

dsgrid/dataset/dataset_schema_handler_base.py
elainethale: Not erroneous, just message selection: in the original, that branch chose between two error strings that differed only in their first clause ('The dataset's time dimension requires...' vs 'The target time dimension requires...') - the guidance and remedies were identical. The port collapses them into one message that doesn't name which side triggered the requirement ('Time mapping requires time zone information from a geography dimension...'), which is why the branch disappeared. This was a planned deviation (noted in PORT_FIXES_PLAN.md and the commit message). If you'd rather keep the which-side detail for debuggability, I can parameterize one message with the triggering dimension instead of duplicating the text - say the word.

tests/test_timezone_validation.py
elainethale: It is a real behavior test, but poorly expressed - fixed in 8dd3f872. check_timezone_in_geography(dimension, err_msg=...) takes an optional caller-supplied message; callers like DatasetConfigModel.check_time_zone pass context-specific text ('This dataset uses aligned_in_std_clock_time...'). The all_null test pins the DEFAULT message on the all-null path; this one pins that a caller-supplied err_msg REPLACES that default. Renamed to test_check_timezone_in_geography_err_msg_override with a docstring, a realistic message string, and an assertion that the default text is absent.

tests/test_timezone_validation.py
elainethale: No good reason - dsgrid.common has no GEOGRAPHY_COLUMN constant (the column is named by DimensionType.GEOGRAPHY.value, which is what add_time_zone joins on), and the original test hardcoded the literal. Fixed in 8dd3f872 to use DimensionType.GEOGRAPHY.value.

dsgrid/config/project_config.py
```diff
@@ -273,12 +273,19 @@ def check_names(cls, values: list) -> list:
...
-        for dimension in values:
-            if dimension.dimension_type == DimensionType.GEOGRAPHY:
-                check_timezone_in_geography(
-                    dimension,
-                    err_msg="Project geography dimension records must include a time_zone column",
-                )
+        time_dims = [d for d in values if d.dimension_type == DimensionType.TIME]
+        # Conservative: when the time dimension is supplied via references (not
+        # inline), keep the historical unconditional geography check.
+        requires_tz = (not time_dims) or any(
```
elainethale: Requiring a time zone when there are no time_dims seems unnecessary / like a gotcha if at some point we relax dimensionality requirements?

tests/test_create_time_dimensions.py
```diff
@@ -368,3 +377,12 @@ def test_time_dimension_model_lead_day_adjustment(time_dimension_model0):
...
     )
+
+
+def test_index_time_single_tz_model(index_time_dimension_model_single_tz):
+    model = index_time_dimension_model_single_tz
+    assert not model.is_time_zone_required_in_geography()
+    config = IndexTimeDimensionConfig(model)
+    assert config.get_time_zone() == "Etc/GMT+5"
+    assert config.get_lengths() == [8784]
+    assert config.get_start_times()[0].tzinfo is not None
```
elainethale, line 394: These tests seem trivial / like they're just regurgitating values set by us on a Pydantic model. Is that correct?

dsgrid/config/project_config.py
elainethale: Agreed - dropped in e8c1df6d, which also brings this back in line with the original implementation (the not-time_dims clause was my deviation). requires_tz is now simply any(is_time_zone_required_in_geography) over the inline time dims. One caveat for the record: a project whose time dimension arrives via base_dimension_references now skips this check entirely (the references carry only IDs, so the format is unknowable at config-validation time) - the submission-time check from Step C still covers datasets mapping into such projects.

tests/test_create_time_dimensions.py
elainethale: Partially correct - cleaned up in e8c1df6d. The get_time_zone assertion was indeed an echo and is gone. What remains is the non-trivial part: the new json5 entry #[8] parsing end-to-end through DimensionsConfigModel (the fixture), is_time_zone_required_in_geography() being format-dependent (new branch logic), and derived values computed from the config (get_lengths from the index range, tz-aware start times from the single-tz format). Accessor behavior on directly-constructed models lives in tests/test_timezone_validation.py; the test docstring now states that split.

dsgrid/dataset/dataset_schema_handler_base.py
```diff
@@ -986,6 +989,47 @@ def _convert_time_dimension(
...
+                    logger.info("Add time zone from project geography dimension.")
+                    geography_dim = to_geo_dim
+                else:
+                    # Reachable from standalone dataset queries, which have no
+                    # dataset-to-project geography mapping.
+                    geography_dim = self._config.get_dimension_with_records(
+                        DimensionType.GEOGRAPHY
+                    )
+                    if geography_dim is not None and _has_time_zone_column(geography_dim):
+                        logger.info(
```
daniel-thom: This case is hard to follow. How do we get here? Why would the user set the use_project_geography_time_zone AND provide time zones in the dataset geography dimension? That would imply different final timestamps in project mapping vs standalone dataset mapping.

dsgrid/query/query_submitter.py
```diff
@@ -545,12 +567,6 @@ def _process_aggregations(
         if context.model.result.replace_ids_with_names:
             df = handler.replace_ids_with_names(df)
 
-        if context.model.result.sort_columns:
```
daniel-thom, line 548: So, we were previously sorting and pivoting the returned dataframe but not the one on disk. Now, we are doing the opposite. We need to re-read the dataframe from disk after post-processing.

elainethale: Confirmed, and there was a second half to it. Fixed in b91e8a53.

You're right that the returned dataframe stopped being post-processed: `_save_query_results` applied `_postprocess` to the copy it wrote and returned only the filename, so the table `submit()` hands back (and the one `_convert_time_zone` consumes) was unsorted and unpivoted.

But re-reading from disk on its own would still hand back an unsorted dataframe on Spark, because the file can't carry the order. A Spark parquet write is a directory of part files; `ORDER BY` range-partitions so each file is internally sorted, but reading the directory back packs splits by descending file size, so the concatenation is rotated. Measured on a 281k-row result:

```
part-00000 size=292379 t=[0 .. 71550]        sorted=True
part-00001 size=274980 t=[71551 .. 138946]   sorted=True
part-00002 size=289008 t=[138947 .. 209846]  sorted=True
part-00003 size=290386 t=[209847 .. 281087]  sorted=True
read-back globally sorted: False   # returned as 00000, 00003, 00002, 00001
```

With one part file it happens to work, which is why the pivoted annual-time query passed and `test_electricity_values` didn't.

So: `_save_result` now returns the table re-read from disk, and the ordering is re-applied to that (`_apply_sort`), which does hold on both backends. `sort_columns` therefore promises an ordered *dataframe*; a new opt-in `single_output_file` coalesces to one part file for callers who need the file itself ordered.

Two details worth flagging:

- Intermediate saves deliberately keep the caller's table instead of the re-read one. Binding it to `output_dir/table.parquet` meant `_convert_time_zone` could build lazily on a path it then deletes before rewriting.
- CSV output is not re-read. `write_csv` already collects to a single ordered file, so it matches the post-processed table, and a round trip would re-infer dtypes from text.

Tests now assert sortedness on the returned dataframe (via `assert_globally_sorted`, which projects to the sort columns before collecting) rather than on the file.

dsgrid/utils/dataset.py
```diff
@@ -761,6 +762,57 @@ def _to_chronify_time_based_data_adjustment(
...
+    """
+    if not isinstance(time_dim, DateTimeDimensionConfig):
+        return time_config
+    if time_dim.get_localization_plan() != "localize_to_single_tz":
+        return time_config
+    # localize_to_single_tz requires ALIGNED_IN_ABSOLUTE_TIME, whose to_chronify()
+    # branch always returns DatetimeRange.
+    assert isinstance(time_config, chronify.DatetimeRange), time_config
```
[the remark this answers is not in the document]
elainethale: I disagree that this warrants raising an Exception. We are checking for something that should always be true at this point in the code, not something that should be impacted by input data.

dsgrid/config/dimensions.py
```diff
@@ -892,6 +926,12 @@ class IndexTimeDimensionModel(TimeDimensionBaseModel):
             "options": MeasurementType.format_for_docs(),
         },
     )
+    time_zone_format: Union[AlignedTimeSingleTimeZone, LocalTimeMultipleTimeZones] = Field(
+        title="time_zone_format",
+        discriminator="format_type",
+        description="Specifies whether all geographies share a single time zone or each "
+        "geography has its own.",
+    )
```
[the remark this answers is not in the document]
elainethale: https://github.com/dsgrid/dsgrid/pull/426/commits/f59b2acc894d4b444cab76be55e757e148d5df4e

dsgrid/config/index_time_dimension_config.py
```diff
@@ -1,14 +1,17 @@
 import logging
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, tzinfo
 from typing import Type, Union
+from zoneinfo import ZoneInfo
```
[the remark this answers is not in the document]
elainethale: https://github.com/dsgrid/dsgrid/pull/426/commits/f59b2acc894d4b444cab76be55e757e148d5df4e

daniel-thom (approved):

dsgrid/dataset/dataset_schema_handler_base.py
elainethale: Good question, and chasing it down turned up that the flag is a vestige. I've removed it.

**How we got there.** `d3867e3f` (2022) added `use_project_geography_time_zone` to select an *ordering*, not a source:

```python
def _convert_time_before_other_dimensions(self):
    return (
        time_dim.model.does_geography_require_time_zone()
        and not self._config.model.use_project_geography_time_zone
    )
```

`False` meant convert time **before** mapping the other dimensions, so the dataset's own geography ids were still in the column and its own records joined correctly. `True` meant convert afterward against the project's geography. Both were coherent.

`b074d0c2` ("Simplify time dimension interfaces", Jun 2024) deleted `_convert_time_before_project_mapping` and every call site, making time mapping unconditionally last. The flag survived. Since then the `False` branch joins the **dataset's** geography records against **post-mapping** ids, an inner join that drops every row with no error. StandardScenarios ships `state_to_county` disaggregation mappings for the EFS and EIA 861 datasets with the flag at its default of `False`; that only fails to bite because those datasets use annual or absolute time.

**Why the branch you flagged existed.** `add_time_zone` joins on the `geography` column, so the only records that can match are those owning the ids in that column at that moment. `to_geo_dim is not None` already encodes exactly that, which makes the flag redundant when it agrees and wrong when it disagrees. And to your second question: the combination is one dsgrid itself mandates. The config validator required a `time_zone` column on the dataset's geography whenever the time dimension is locally aligned with naive timestamps, *even with the flag true*, because registration localizes with the dataset's geography. Derived datasets also get `use_project_geography_time_zone: true` hardcoded.

**What's here now.** Time mapping takes time zones from the geography dimension that owns the ids in the column. A project query always supplies the project's base geography, mapped or not, since an unmapped dataset geography is already in the project's id space. Only a standalone dataset query can leave it unset, and then the dataset's own records are the right ones. No flag, no fallback branch.

Four related fixes came out of the trace, each with tests:

- The selected geography is checked for a usable `time_zone` column, so a missing one names the dimension instead of failing deep in chronify. The pydantic field always exists, so a geography CSV without the column still joined cleanly and produced an all-null column.
- The geography join is skipped when the table already carries `time_zone`. Registration-time localization leaves the column behind, and a dataset may supply it directly; joining again collided on the column name.
- Time zones absent from the time dimension's `time_zones` list are now rejected. chronify silently discarded those rows during localization: four rows in, two out, no error.
- `_check_time_zone_for_mapping` validates the project's base geography, closing a gap where a project on absolute time accepting a locally-aligned dataset passed every check and failed inside chronify.

The two-table format now also accepts a `time_zone` column, matching what one-table already allowed.

Two things I found but did not fix, filed separately: two-table registration writes the load_data/lookup join to the load_data path when localization runs, and chronify has no mapper for datetime with per-row time zones.

